PR #115, opened 2015-01-12
doc: mention ByteTensor for querying/filling elems

deltheil:
The documentation related to binary masks for setting or getting tensor values was missing.


soumith:
Excellent! Thanks

PR #116, opened 2015-01-14
Add batch matrix-matrix multiplication.

dominikgrewe:
Add torch.bmm, which expects two 3D tensors as inputs. A 'b x m x n'
tensor is interpreted as a batch of b matrices of size 'm x n'.
The result is a batch of matrices computed by multiplying the
corresponding pair of matrices in the two input batches.

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -312,6 +312,13 @@ for _,Tensor in ipairs({"ByteTensor", "CharTensor",
          {name=Tensor, dim=2}}
      )
 
+   wrap("bmm",
+        cname("bmm"),
+        {{name=Tensor, default=true, returned=true},
+         {name=Tensor, dim=3},
+         {name=Tensor, dim=3}}
+     )
+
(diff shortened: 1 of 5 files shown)


dominikgrewe:
It may seem a bit unnecessary to add this to torch, because you can easily write this loop yourself in Lua. But for cutorch that wouldn't be very efficient unless the matrices are very large. cuBLAS provides a function that takes two batches of matrices and performs the matrix-matrix multiplications in parallel, which will be much faster. So I'm planning to add a THCudaTensor_bmm method to cutorch, taking advantage of that.

nicholas-leonard:
@dominikgrewe Yeah I agree that this is a must-have for torch. I have used the cuda batched_gemm before in cunnx and I was hoping to see it wrapped in torch. Thanks dominik!

clementfarabet:
This is great, also useful cpu side for very small matrix muls!

PR #119, opened 2015-01-20
Added support for hashing tensors.

jjh42:
(note: this incorporates public domain
code from https://github.com/andikleen/spooky-c
to implement spooky hash).


lib/TH/generic/THTensorMath.c
```diff
@@ -216,6 +218,35 @@ accreal THTensor_(sumall)(THTensor *tensor)
...
+
+long THTensor_(chash)(THTensor *tensor)
+{
+  if(tensor->nDimension == 0) {
+    return 0;
+  }
+
+  if(THTensor_(isContiguous)(tensor)) {
+    return spooky_hash64(THTensor_(data)(tensor),
+                         THTensor_(nElement)(tensor)*sizeof(real), 42);
```
timharley: Is this 42 the same as on line 224 above?

lib/TH/generic/THTensorMath.c
```diff
@@ -216,6 +218,35 @@ accreal THTensor_(sumall)(THTensor *tensor)
   return sum;
 }
 
+static long THTensor_(noncontiguousHash)(THTensor *tensor)
+{
+  struct spooky_state hash_state;
+  unsigned long hash1 = 42, hash2 = 42;
```
timharley: Perhaps add a comment explaining why these need to be initialized to 42?

clementfarabet:
This is great. One quick request: could you add a check in CMake to only include/build this code if the proper platform is detected? It probably doesn't work correctly on 32bit machines right?

lib/TH/generic/THTensorMath.c
jjh42: its just an arbitrary magic number. I will seperate it out into a constant.

jjh42:
@clementfarabet I think it should also work on a 32-bit machine but the hash for the same content might change. I'll try and test.

andresy:
Hi, do you really want to add that in the core? I mean, there are plenty of different hashes out there, I am not sure I like the idea of having one specific one in the core. For e.g., in tds, https://github.com/torch/tds, I chose also a version of Bob Jenkins' hash, but it is another implementation.

soumith:
it is not building a hash container, but just returning the hash of a tensor
i feel like this has lots of good use cases, and maybe we can have one implementation in the core, as a reference and out of the box...

soumith:
what i dont like is that we are adding a "thirdparty" into the core.

jjh42:
@soumith what do you prefer for using a hashing library?

soumith:
@jjh42 i dont not like it too much, it's just the fact that we are now shipping the core with a third-party lib embedded in there, but again, we have precedent there with the sorting functions that @jucor wrote in the past (if I remember)

I will let the others comment and review this.

jucor:
Thanks for the shout-out. However, I do not remember relying on extra third-party lib when @d11 and I tweaked the sorting functions: see commits 7a41d6da21da8ba0a45550b8c82f1dd0ed480db7 and 1d7b661ae17060ad56898b0685a6f17a9f0437a6. Could you please help me remember which functions you were thinking of?

soumith:
yes, i meant that the precedent here is that we incorporated public domain sorting functions in-line.

soumith:
thanks for pullling the commits @jucor

jucor:
No problem, happy to help :) The import of the public domain sorting code was actually done by @koraykv loooong ago, if I remember my conversations with him back when @d11 and I tweaked them a bit. Unfortunately, I can't find the commit of his original import to double-check, since it is masked by the massive "initial revamp" commit https://github.com/torch/torch7/commit/053065ba239e3421d1b2cbec55e3f177a21d6829. Worth double-checking with him, though. That makes it an even-stronger precedent :)

jucor:
Ah, a bit of archeology: the public code for sort was actually imported by @andresy  in revision 655 of Torch5 (!), according to `svn blame` on file `packages/lab/manip.c` of the old Torch5 repo obtained via `svn checkout svn://svn.code.sf.net/p/torch5/code/trunk torch5-code`.

timharley:
Just wanted to add that the Mersenne Twister implementation for the random number generator is third party too:
https://github.com/torch/torch7/blob/master/lib/TH/THRandom.c#L41

clementfarabet:
Agreed with @timharley – I don't really see the issue of having third party code integrated, if it's clean. The question then is which is better: third party code or no code at all :-) 

Also, the hashing function of Lua itself is pretty arbitrary. It works ok for common cases, poorly for certain cases (not even mentioning LuaJIT's hashing function...).

koraykv:
I am a bit divided too, this functionality could easily be provided as an additional lib, but having a hash of tensors is quite basic requirement like sort. The hash code in this case is quite clean, very commonly used and trusted and public domain. So I would vote to include in this case.

soumith:
ok this is great, but one really important question that has none of us raised. What do we do with torch.hash for CUDA? How trivial or non-trivial will it be, and in that case, should we rely on a hash implementation that maybe has both C and CUDA, or is not too hard to implement in CUDA...

jjh42:
@dominikgrewe do you know anything about hashing on CUDA?

jjh42:
@soumith spooky hash is in pure C so I think it should be relatively straightforward to port to CUDA, but wouldn't be making us of the GPU at all (its not parallel).

Do you think that its important that tensors with the same content hash to the same value regardless of type? (i.e. so a cuda and float tensor with identical contents have identical hashes).

timharley:
Another discussion point: Should Tensors with the same content really hash to the same value regardless of *shape*?

jjh42:
@timharley I think so. The reason is I could think of cases where you might want this (i.e. you reshape something for matrix operations) but I couldn't think of cases where you wouldn't.

jucor:
Tricky question. If we want to diminish the surprise: what does Python do
for nd-arrays? Should we leave the choice to the user (e.g. via flag) to be
either shape-and-type-dependent or shape-and-type-independent? Thus with
possibility for the shape-and-type-dependent to use a different and faster
algo on CUDATensors.

Julien Cornebise | Senior Research Scientist | Google DeepMind |
jucor@google.com

jjh42:
I don't think numpy has direct support for hashing arrays. You can indirectly hash the array.data (which ignores the shape)
https://stackoverflow.com/questions/16589791/most-efficient-property-to-hash-for-numpy-array

soumith:
tic toc, what have we decided on this. @dominikgrewe how trivial (or non-trivial) would this be on the CUDA side?

andresy:
i am afraid it is still undecided. some people told me about different hashes they might use in different use cases. so i am not sure. adding this the shape issue mentioned above, i am even less sure.

andresy:
http://on-demand.gputechconf.com/gtc/2014/presentations/S4507-evaluation-of-parallel-hashing-techniques.pdf

dominikgrewe:
I don't have any experience with hashing on GPUs. Most of the things you find online seem to be about computing lots of hash values in parallel.

If we choose a "tree hash" though, it seems possible to get some decent performance on GPUs. Tree hashes break your data into blocks and compute individual hashes. These hashes are then reduced to a single hash value. See https://www.andrew.cmu.edu/user/gbarboza/418/

soumith:
at this point it looks like this PR is a no-go for the core? I personally like some form of it, but the others disagree. closing it. Open it if you have an issue.

soumith:
even if there's hashing for a storage, i think that's useful. That's exactly what numpy does.

PR #120, opened 2015-01-20
Element-wise pow function

GeorgOstrovski:
Fully analogous to cmul/cdiv.


soumith:
thanks a lot, could you squash your commits.

PR #122, opened 2015-01-21
Added alternative version of pow function

GeorgOstrovski:
Added a version of pow which takes a scalar (n) as basis and a tensor (x) as exponent instead of the other way around.


soumith:
thanks again!

PR #123, opened 2015-01-21
torch.totable for Storages and for Tensors.

timharley:
* Currently, you can convert a Storage to a lua table with `storage:totable()`. This patch adds the same functionality to Tensors: `tensor:totable()`. Multidimensional tensors give nested tables of equal shape to the source tensor.

* Add general function for storages and tensors: `torch.totable(...)`.

* Added `torch.isStorage` function.

### DIFF
--- a/Tensor.lua
+++ b/Tensor.lua
@@ -277,7 +277,7 @@ end
 function Tensor.expand(result,tensor,...)
    -- get sizes
    local sizes = {...}
-   
+
    local t = torch.type(tensor)
    if (t == 'number' or t == 'torch.LongStorage') then
       table.insert(sizes,1,tensor)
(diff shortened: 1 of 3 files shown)


soumith:
great! thanks.

clementfarabet:
:+1: nice

andresy:
thanks!
what is the idea behind isStorage though?

timharley:
I'll admit that it doesn't add much, but seeing as we already have `torch.isTensor`, I thought it would make things more consistent. Also makes for a nicer error from `torch.totable(...)`.

andresy:
ok, thanks :+1:

nicholas-leonard:
awesome.

PR #128, opened 2015-01-22
Renamed baddmm -> addbmm; added baddbmm.

clementfarabet:
- addbmm is a batch MM + reduce add
- baddbmm is a batch MM + batch add

### DIFF
--- a/lib/TH/generic/THTensorMath.c
+++ b/lib/TH/generic/THTensorMath.c
@@ -658,7 +658,7 @@ void THTensor_(bmm)(THTensor *result, THTensor *batch1, THTensor *batch2)
 
   THTensor *matrix1 = THTensor_(new)();
   THTensor *matrix2 = THTensor_(new)();
-  THTensor *result_matrix = THTensor_(new());
+  THTensor *result_matrix = THTensor_(new)();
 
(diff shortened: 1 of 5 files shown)


clementfarabet:
Anyone has any objections to the naming? The idea is to have the "b" prepended to the part of the operator that's batched.

lib/TH/generic/THTensorMath.c
```diff
@@ -709,6 +709,46 @@ void THTensor_(baddmm)(THTensor *result, real beta, THTensor *t, real alpha, THT
   THTensor_(free)(matrix2);
 }
 
+void THTensor_(baddbmm)(THTensor *result, real beta, THTensor *t, real alpha, THTensor *batch1, THTensor *batch2)
```
dominikgrewe: We could implement torch.bmm in terms of baddbmm, couldn't we? We just need to make sure that torch.bmm resizes the result tensor before passing it to badbmm.

clementfarabet:
@dominikgrewe that's right – we could simplify a bit like that.

soumith:
naming looks good to me

clementfarabet:
Cool, merging.

PR #132, opened 2015-01-23
Substitute ~ for user home dir in File.load and File.save

szagoruyko:
Really miss this in torch. Let's say you use th and type ```torch.load '~/Docume..'``` and it autocompletes the path for you, but can't save or load after!

### DIFF
--- a/File.lua
+++ b/File.lua
@@ -258,6 +258,7 @@ end
 -- simple helpers to save/load arbitrary objects/tables
 function torch.save(filename, object, mode)
    mode = mode or 'binary'
+   local filename = filename:gsub('~',os.getenv('HOME'))
    local file = torch.DiskFile(filename, 'w')
    file[mode](file)
    file:writeObject(object)
@@ -266,6 +267,7 @@ end
 
 function torch.load(filename, mode)
    mode = mode or 'binary'
+   local filename = filename:gsub('~',os.getenv('HOME'))
    local file = torch.DiskFile(filename, 'r')
    file[mode](file)
    local object = file:readObject()


soumith:
this is not a standard thing. ~ expansion is not for just HOME
~user will give you someone else. 

All in all i think it is not generalizable yet, with this PR

szagoruyko:
ah, I see. Probably the right way is to use POSIX wordexp in DiskFile.

szagoruyko:
it's +9 lines https://github.com/szagoruyko/torch7/compare/torch:master...master, but won't work in Windows and crossplatform will be massive.

leonbottou:
On Friday, January 23, 2015 05:27:30 AM Sergey Zagoruyko wrote:
> it's +9 lines
> https://github.com/szagoruyko/torch7/compare/torch:master...master, but
> won't work in Windows and crossplatform will be massive.

What about filenames with spaces? with reserved shell characters? with quotes? 
with dollar signs?  Function wordexp() is too strong for this because it also 
splits the input string into many words, performs backquote expansion (calling 
arbitrary programs, btw), performs wildcard expansion, etc.

- L.

szagoruyko:
ok, it's not trivial to implement, thanks Leon.

leonbottou:
On Friday, January 23, 2015 07:43:20 AM Sergey Zagoruyko wrote:
> ok, it's not trivial to implement, thanks Leon.

What you can do is the following

if filename[0] == '~' 
   username = longest-string-of-alphanumeric-chars-following-tilde
   rest = remaining-characters-in-filename
   if username == '' then
	   home = getenv(HOME)
      if home != '' then 
	     return getenv(HOME) .. rest
      end
   elif username != '' then
      pwent = getpwnam(username)   --- unix only. 
      if pwent != null then
        return pwent->pw_dir .. rest
      end
   end
return filename

Note that ~ and ~yourname might return different results if HOME has been 
redefined. Same as in the shell in fact.

- L

szagoruyko:
The problem is getpwnam is not available in lua and torch and crossplatform c code to use it would be huge and messy. I'd suggest to only handle the case '~/...': https://github.com/szagoruyko/torch7/compare/torch:master...master

PR #134, opened 2015-01-26
Permute convenience function 

GeorgOstrovski:
Added Tensor.permute convenience function to replace sequences of transpose() calls by a more succinct and less error-prone notation x:permute(a1, a2, ...), where a full permutation (a1, a2, ... , aN) is specified for the N dimensions of a tensor.

### DIFF
--- a/Tensor.lua
+++ b/Tensor.lua
@@ -525,6 +525,25 @@ function Tensor.totable(tensor)
   return result
 end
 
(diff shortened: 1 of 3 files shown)


nicholas-leonard:
This looks good.

GeorgOstrovski:
I just reopened this as a different PR (sorry for the confusion).

PR #135, opened 2015-01-26
Permute

GeorgOstrovski:
Added Tensor.permute convenience function to replace sequences of transpose() calls by a more succinct and less error-prone notation x:permute(a1, a2, ...), where a full permutation (a1, a2, ... , aN) is specified for the N dimensions of a tensor.

### DIFF
--- a/Tensor.lua
+++ b/Tensor.lua
@@ -525,6 +525,25 @@ function Tensor.totable(tensor)
   return result
 end
 
+function Tensor.permute(tensor, ...)
+  local perm = {...}
+  local nDims = tensor:dim()
+  assert(#perm == nDims, 'Invalid permutation')
+  local j
+  for i, p in ipairs(perm) do
+    if p ~= i and p ~= 0 then
+      j = i
+      repeat
+        assert(0 < perm[j] and perm[j] <= nDims, 'Invalid permutation')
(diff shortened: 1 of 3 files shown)


Tensor.lua
```diff
@@ -525,6 +525,25 @@ function Tensor.totable(tensor)
   return result
 end
 
+function Tensor.permute(tensor, ...)
+  local perm = {...}
+  local nDims = tensor:dim()
+  assert(#perm == nDims, 'Invalid permutation')
```
nicholas-leonard: Could you make this constraint `#perm <= nDims`, and update doc and unit tests to reflect this?

GeorgOstrovski: How would an under-specified permutation be interpreted then? E.g., 
x = torch.Tensor(1,2,3,4)
y = x:permute(4,3) 
-- not really well-defined and possibly mistaken for transpose(4,3)

For this to work, it would require that the user specified a full permutation on the first n <= N entries (e.g. in the above example, permute(3,1,2) would actually work).
This seemed to me a somewhat questionable usecase and not worth encouraging?

nicholas-leonard: I see what you mean. Disregard my last comment.

soumith:
looks good to me. falls nicely in like with matlab's permute. thanks!

PR #136, opened 2015-01-27
torch.gels crashes in underdetermined case

jucor:
Dear all,

`torch.gels()` crashes in underdetermined case when m < n, although [the doc says](https://github.com/torch/torch7/blob/master/doc/maths.md#torchgelsresb-resa-ba) it should work -- since Lapack supports it.

As we dug deeper with the help of @yozw, @yuvaltassa and @theoweber, we found it is because torch.gels actually [passes the wrong dimensions](https://github.com/torch/torch7/blob/9b519070d8a854e760a05b4ac9273f70450ad045/lib/TH/generic/THTensorLapack.c#L155) to lapack.

I do not know how to fix this (I am not fluent enough with extra allocations, and remember @koraykv saying that lapack wrapping in torch is black magic better left to certified wizards ;-) ), so this pull-request unfortunately does not contain a fix.

However, what it contains is:
* full unit test to reproduce the crash and check that the other cases do not crash
* full unit test to ensure  sure that we get the proper result  both in the over-determined, exactly-determined, and under-determined cases, comparing to an implementation of least squares using regularized pseudo-inverse.

Would someone with better handling of torch's Lapack wrapping be willing/able to take it from there please, or to explain how to proceed, please?

Thanks!

### DIFF
--- a/test/testGELS.lua
+++ b/test/testGELS.lua
@@ -1,0 +1,169 @@
+-- Test linear regression with m observations, and dimension n
+-- Note: in statistics and linear regression, the number of obs would be 'n'
+-- and the dimension 'p', but I stick to torch and lapack's conventions
+
+require 'torch'
+require 'totem'
+
+local tester = totem.Tester()
+local myTests = {}
+
+
+local verbose = false
+local verbose, verbose
+if verbose then
+  verbose = print
+else
+  verbose = function() end
+end
+
+
+-- Regression using torch.gels
+local function regGels(X, Y)
+  local theta = torch.gels(Y, X)
+  return theta
+end
+
+
+-- Regression reimplemented with pseudo inverse and regularization
+local function regPseudoInv(X, Y)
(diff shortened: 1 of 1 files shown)


soumith:
ping @koraykv the magician.

jucor:
For the record, @koraykv can speak to @yozw who investigated the problem in depth :)

koraykv:
I will look into this this with @jucor .

jucor:
With pleasure @koraykv, let's catch up tomorrow.

Julien Cornebise | Senior Research Scientist | Google DeepMind |
jucor@google.com

On Sun, Feb 15, 2015 at 7:52 PM, koray kavukcuoglu <notifications@github.com

yozw:
This pull request can be closed, I think. See https://github.com/torch/torch7/pull/158.

jucor:
Sorry, forgot about it. Closing, thanks!

PR #137, opened 2015-01-27
torch.histc (histogram): doc + test

deltheil:
A piece of doc + test. Also: `hist.lua` sounds legacy (it is not present within `CMakeLists.txt`) so I removed it in addition.


soumith:
looks good. as a policy (for torch,nn,cutorch,cunn), we squash the commits that are in a Pull Request to 1 single commit. This helps a lot in looking at commit history, and reduces the time in triaging a bug.

So, just squash your commits, with:
git rebase -i
then you can push it back to the same PR with
git push --force

let me know when you do that and will merge. thanks.

deltheil:
@soumith done!

Regarding the policy: good to know, thanks :)

soumith:
thank you!

PR #139, opened 2015-01-28
Fixed getRNGState in uninitialized case

GeorgOstrovski:
Previously, getRNGState would return an invalid state if called before any random function call. This would mean that

> require 'torch'
> s = torch.getRNGState()
> =torch.uniform()
0.9989492399618
> torch.setRNGState(s)
> = torch.uniform()
0.046927026240155 -- different result from before, though the state is supposedly the same!!


soumith:
thanks!

PR #141, opened 2015-01-30
Make sure RNG is always seeded

GeorgOstrovski:
The current way the torch RNG is initialized  (THGenerator_new) creates an invalid (unseeded) first RNG state, which is then lazily seeded in a number of functions depending on a valid state. This is error prone (as was seen previously, where getRNGState would potentially return an invalid state). 

I suggest a minor refactor (thanks to @timharley for the suggestion): THGenerator_newUnseeded is provided as a static function in THRandom.c, so that THRandom_manualSeed which does not need a seeded generator can still obtain an unseeded one. On the other hand, THGenerator_new is changed to always return a seeded generator, so that unseeded generators never leak outside of THRandom.c. The lazy seeding is removed from all places where this was previously needed.

Further, THRandom_getState and _setState are removed (their use from within getRNGState and setRNGState has been removed a while ago and they seem deprecated). 

@koraykv

### DIFF
--- a/doc/random.md
+++ b/doc/random.md
@@ -8,15 +8,15 @@ random number generator.
 <a name=":torch.seed.dok"/>
 ## Seed Handling ##
 
-If no seed is provided to the random generator (using
-[seed()](#torch.seed) or [manualSeed()](#torch.manualSeed)), a
-random seed will be set according to [seed()](#torch.seed) the first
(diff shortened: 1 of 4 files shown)


soumith:
thansk Georg, looks good (I think)

PR #144, opened 2015-02-04
Make bmm use baddbmm.

dominikgrewe:
Just like mm uses addmm, we can write bmm in terms of baddbmm. This way the
two behave the same in terms of when the output is resized.

Also gets rid of unnecessary zeroing out of output of mm and mv.


soumith:
looks good i think!

PR #145, opened 2015-02-05
Fix parent class validity check

GeorgOstrovski:
Previously, the existence of the parent class wasn't actually verified, since the return value of luaT_pushmetatable wasn't checked. 
This for instance allowed the following:

my = {}
MyC, p = torch.class('my.class', 'nonexistent.class')
MyC:__init() print('constructor') end

Weird effect:
x = my.class() -- constructor not actually called here! (broken metatable?)


soumith:
ping @andresy

soumith:
thanks Georg

PR #147, opened 2015-02-09
Qsort from reference article

jucor:
As disscussed with @soumith in https://github.com/torch/DEPRECEATED-torch7-distro/pull/198, I have rebased my forgotten unmerged PR onto this new repo :)

I have checked: all the tests pass, and so do the timeSort.lua script, which can be used to try the speed-up (or -down) on your own machine. The graphs in the original submission were on my laptop. I include here the graphs on a much more powerful desktop machine: the improvement is bigger but only kicks in later. 

You might want to check it on your own machine to see what you get and if it's worth it :)

On my desktop machine:
benchmarkratio

---- Original submission below, with graphs on my laptop ---
Hi guys

As discussed with Koray, this pull request replaces the wikipedia-based-then-improved handmade implementation by an implementation from a reference academic article by Sedgewick (who did his thesis with Knuth on quicksort):
 
Sedgewick's 1978 "Implementing Quicksort Programs"
http://www.csie.ntu.edu.tw/~b93076/p847-sedgewick.pdf                                                                                                                                         

It is the state of the art existing implementation of regular single-pivot quicksort. The macros are here to make as close a match as possible to the pseudocode of Program 2 p.851.

Here are the benchmarks against the existing implementation: 

benchmarkratio
benchmarktime

As you can see, although we loose a little on very small arrays (which are so fast to sort that it is not noticeable), we gain up to 50% on constant arrays, and 10% on random arrays. We can even see the difference when some caching stops being done for large array sizes.

The advantage, beyond this reasonable improvement, is also that we now have a proper reference, with details about each carefuly made implementation decision by Sedgewick. 

Note that I have only done the ascending sort, leaving the descending sort as it was for comparison purposes. If you decide to merge this branch, I will copy/paste the code and change the inequalities to get the descending sort.
Cheers.

### DIFF
--- a/lib/TH/generic/THTensorMath.c
+++ b/lib/TH/generic/THTensorMath.c
@@ -1,4 +1,3 @@
-#include <assert.h>
 #ifndef TH_GENERIC_FILE
 #define TH_GENERIC_FILE "generic/THTensorMath.c"
 #else
--- a/lib/TH/generic/THTensorMath.c
+++ b/lib/TH/generic/THTensorMath.c
@@ -1080,14 +1080,10 @@ static void THTensor_(quicksortascend)(real *arr, long *idx, long elements, long
       P=(L+R)>>1; /* Choose pivot as middle element of the current block */
 
       piv=arr[P*stride];
-      rswap=arr[L*stride];
-      arr[L*stride]=piv;
-      arr[P*stride]=rswap;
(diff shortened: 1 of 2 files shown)


soumith:
@jucor could you rebase the commits to a single block (according to new policy :) ). reviewing now.

jucor:
Done, rebased in pull-request #148, closing and continuing the discussion there.

PR #151, opened 2015-02-12
Added torch.[de]serialize{To/From}Storage to File.lua.

malcolmreynolds:
These match the previous [de]serialize functions, but return
a torch.CharStorage instead of a Lua string. This can be used to
prevent hitting the LuaJIT 2GB limit for very large objects, because the
CharStorage is allocated on the C side (not counting towards this limit)
but the Lua string will count towards this limit.

This modification fixed a repeatable OOM crash I found for trying to
serialize certain large models.

### DIFF
--- a/File.lua
+++ b/File.lua
@@ -275,20 +275,25 @@ end
 
 -- simple helpers to serialize/deserialize arbitrary objects/tables
 function torch.serialize(object, mode)
+   local storage = torch.serializeToStorage(object, mode)
+   return storage:string()
+end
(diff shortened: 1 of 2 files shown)


dominikgrewe:
It's probably also worth noting that this eliminates an extra copy (from storage to string), and should therefore be faster.

soumith:
cool! thanks. This limit is being removed soon (mike pall already has the non-JIT version working in his mainline).

PR #152, opened 2015-02-12
maskedCopy,maskedFill,maskedSelect

nicholas-leonard:
This PR exposes the above C functions as Tensor methods. Useful when you want to avoid reallocating memory, which isn't possible with the current `Tensor[mask]` schemes (`maskedSelect` in particular). The PR includes unit tests and doc.


soumith:
thanks nick!

PR #154, opened 2015-02-13
Checking for valid RNG state

GeorgOstrovski:
After my last PR, the RNG state returned by getRNGState is guaranteed to be valid. RNG states stored previously (with older version of torch) however could be invalid. For this and other reasons, it seems reasonable to check for the edge case of an invalid RNG state in setRNGState and throw an error when invalid. (this seems safer then silently re-seeding such state, leading to non-deterministic behaviour)

I renamed `initf` to `seeded` (as suggested by @timharley) and extracted the validity check into a new function, THRandom_isValidState.


soumith:
Thanks georg

PR #158, opened 2015-02-19
Fixing torch.gels in the underdetermined case.

yozw:
The torch.gels function threw a lapack error when calling it with an underdetermined system, i.e., the case m < n. As discussed with @koraykv, I have fixed this and added tests.

### DIFF
--- a/lib/TH/generic/THTensorLapack.c
+++ b/lib/TH/generic/THTensorLapack.c
@@ -2,11 +2,17 @@
 #define TH_GENERIC_FILE "generic/THTensorLapack.c"
 #else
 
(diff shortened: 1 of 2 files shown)


soumith:
Thank you. Can you please squash your commits to a single commit and I will merge it in after a test.

yozw:
Done. With thanks to @jucor.

yozw:
BTW, maybe let's give @koraykv the opportunity to comment before merging this.

jucor:
Always happy to help, @yozw :) Thanks for your work fixing this tricky bug!

koraykv:
Cool, this looks good to me. Thanks a lot!

PR #164, opened 2015-03-02
Remove wrong pushudata

GeorgOstrovski:
This pushudata call seems wrong in the case where the result `tensor` is one of the arguments to maskedSelect.

### DIFF
--- a/generic/Tensor.c
+++ b/generic/Tensor.c
@@ -474,7 +474,6 @@ static int torch_Tensor_(maskedSelect)(lua_State *L)
     src = luaT_checkudata(L, 2, torch_Tensor);
     mask = luaT_checkudata(L, 3, "torch.ByteTensor");
     tensor = luaT_checkudata(L,1,torch_Tensor);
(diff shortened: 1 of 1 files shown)


soumith:
the function is "return 1", which means it expects an argument on stack. 
Also, I believe the idea is to also return the result tensor, even if it was an argument, like: t = t:add(1)

GeorgOstrovski:
I was assuming the returned tensor in this case is the one passed in, already at position 1 on the stack? As in the below function maskedCopy, where only this case is happening and no tensor needs to be put on the stack.

koraykv:
I think the change is proper, @soumith luaT_checkudata is supposed to leave the result on the stack, right? So you would not need to re-push it.

soumith:
Yes seems fine. I am mistaken with my earlier comment.

soumith:
Thanks Georg.

koraykv:
Thanks Soumith.

yozw:
Thanks Koray.

PR #177, opened 2015-03-18
Improve consistency of documentation typography

eliemichel:
All these changes are superficial typograhic changes, but they
try and unify a bit its consistency.

I had to make some choices among the different usages I found:

 * Display all tensor sizes with the syntax `m x n`
   (found mxn, m-by-n)

 * Add `` arround isolated variables and everything that is code

 * Specify that code blacks are in Lua using the GitHub ```lua syntax

 * Put commas inside brackets when arguments are optional
   e.g. commas for [resval, resind,] x [,d] [,flag]
   or also [res,] x, k [, 'F' or 'V']

 * Use '> ' as prompt and display it only if the returned value is then
   shown (and so don't use = or print)
   (found '>', '=', '= ', '> = ', 't7> ', etc. )
   Still work to do on it.

 * Add spaces around * or = but not after commas (this is the guideline
   used in the majority of cases), except in inline code.

+ Correct some typos
+ Move `torch.histc` at its correct place in alphabetical order
+ Review doc for torch.multinomial, esepcially default flag value
+ Review doc for LAPACK related functions

TODO:

 * torch.gels doc not clear
 * Improve signature consistency, especially in tensor.md
 * Continue on other files…

### DIFF
--- a/doc/maths.md
+++ b/doc/maths.md
@@ -53,6 +53,7 @@ The advantage of second case is, same `res2` tensor can be used successively in 
 for i=1,100 do
     torch.conv2(res2,x,k)
 end
+
 =res2:dist(res1)
 0
 ```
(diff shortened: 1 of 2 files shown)


soumith:
wow, it must have taken you a long time to do this! thank you very much.

nicholas-leonard:
Looks good.

PR #179, opened 2015-03-19
Added kthvalue and median operations, tests, doc

GeorgOstrovski:
Added kthvalue function, to select k-th smallest value. Interface as close as possible to min/max, i.e. returns values and indices. Based on Quickselect algorithm (i.e. linear time) adapted from 'Numerical Recipes'. 

Also added median function based on same algorithm. Median returns 'middle' element in case of odd-many elements, otherwise one-before-middle element (could also do the other convention to take mean of the two around-the-middle elements, but that would be _twice_ more expensive, so I decided for this one).

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -478,7 +478,21 @@ for _,Tensor in ipairs({"ByteTensor", "CharTensor",
          {name=Tensor},
          {name="index", default=lastdim(3)},
          {name="boolean", default=0}})
-   
+
+   wrap("kthvalue",
+        cname("kthvalue"),
+        {{name=Tensor, default=true, returned=true},
+         {name="IndexTensor", default=true, returned=true, noreadadd=true},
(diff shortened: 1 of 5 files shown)


nicholas-leonard:
This seems really useful. Thanks.

andresy:
I do not remember the license on Numerical Recipes code, but I thought it was not free...

GeorgOstrovski:
You are right I should have checked. Will revise. Sorry.

GeorgOstrovski:
Would a Public Domain re-implementation as [this one](http://ndevilla.free.fr/median/median/src/quickselect.c) work? Alternatively [this](http://ndevilla.free.fr/median/median/src/wirth.c).

soumith:
@GeorgOstrovski Public domain implementations which aren't copyrighted are fine. thanks.

PR #180, opened 2015-03-21
Change display of null tensors

eliemichel:
They where displayed with narrower columns than others.

For instance, before this commit:

```lua
> torch.Tensor(2,2):zero()
0 0
0 0
[torch.DoubleTensor of dimension 2x2]

> torch.Tensor(2,2):fill(1)
 1  1
 1  1
[torch.DoubleTensor of dimension 2x2]
```

This is because `log10()` is used to determine number width in base 10
but the case where it is 0 is handled differently.


soumith:
thanks Elie!

PR #181, opened 2015-03-21
Homogenize style of code blocks

eliemichel:
As it seems to be ok, I continue what I started in #177 while reading the doc.
Here is an update of `tensor.md` esepcially focused on code blocks.

 * Use '> ' as prompt when the returned value is displayed
 * Use no prompt when there is no returned value
 * Do not use print or return to display a value (it introduces extra parenthesis and noisy characters)
 * Skip a line after displayed value, but not before
 * (In tensor.md only)


soumith:
more good stuff! thanks again.

PR #182, opened 2015-03-22
Added kthvalue and median operations, tests, doc

GeorgOstrovski:
Added kthvalue function, to select k-th smallest value. Interface as close as possible to min/max, i.e. returns values and indices. Based on Quickselect algorithm (i.e. linear time) adapted from [this](http://ndevilla.free.fr/median/median/src/quickselect.c) Public Domain implementation, somewhat changed with tricks from the Quicksort implementation used in Torch.

Also added median function based on same algorithm. Median returns 'middle' element in case of odd-many elements, otherwise one-before-middle element (could also do the other convention to take mean of the two around-the-middle elements, but that would be twice more expensive, so I decided for this one).

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -478,7 +478,22 @@ for _,Tensor in ipairs({"ByteTensor", "CharTensor",
          {name=Tensor},
          {name="index", default=lastdim(3)},
          {name="boolean", default=0}})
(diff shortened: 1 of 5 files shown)


andresy:
good stuff, thanks

PR #184, opened 2015-03-23
Minor refactoring of RNG code

GeorgOstrovski:
Cleaned up some bits of RNG code:
* removed obsolete declarations of no longer existent functions  THRandom_getState/setState
* replaced THRandom_isValidState by THGenerator_isValid
* changed \_\_uniform\_\_ to use THRandom_random (code was practically identical) , which removes a brittleness in how the nextState has to be generated (tricky, error-prone 3 lines of code which were reused in two places previously)

### DIFF
--- a/doc/random.md
+++ b/doc/random.md
@@ -8,10 +8,9 @@ random number generator.
 <a name=":torch.seed.dok"/>
 ## Seed Handling ##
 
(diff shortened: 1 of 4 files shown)


andresy:
thanks

PR #185, opened 2015-03-26
Fix FindSSE SSE4* check crash on MSVC.

wishstudio:
This commit adds initial values to test programs in FindSSE.cmake.
Without these, the vc2013 debug runtime will complain "The variable 'b' is being used without being initialized." when the test program is run, and then mark SSE4* as unsupported.


soumith:
thanks @wishstudio

PR #187, opened 2015-03-31
fixed counter bug in maskedCopy

soumith:
maskedCopy was incorrectly not incrementing the counter on mask value == 0

### DIFF
--- a/lib/TH/generic/THTensorMath.c
+++ b/lib/TH/generic/THTensorMath.c
@@ -41,6 +41,10 @@ void THTensor_(maskedCopy)(THTensor *tensor, THByteTensor *mask, THTensor* src )
 		     cntr++;
 		     if (cntr > nelem)
 		       THError("Number of elements of src != mask");
+		   } else if (*mask_data == 0)
+		   {
+		     src_data++;
+		     cntr++;
 		   });
   if (cntr != nelem)
     THError("Number of elements of src != mask");


soumith:
actually, i'm not sure which one is wrong, the documentation or the implementation:
https://github.com/torch/torch7/blob/master/doc/tensor.md#tensor-maskedcopymask-tensor

The documentation copies elements according to a mask, the implement sets all elements of a masked tensor to src tensor.

koraykv:
I think the documentation is wrong. The only requirement seems to be: the
number of elements in the "tensor"  has to be at least the number of '1's
in the mask. So you can do something like.


t7> from_mask = torch.ByteTensor{1, 0, 1, 0, 1, 0, 1, 0, 1, 0}
t7> to_mask = torch.ByteTensor{1, 1, 1, 1, 1, 0, 0, 0, 0, 0}
t7> x=torch.range(1,10)
t7> y=torch.zeros(10)
t7> y:maskedCopy(to_mask, x:maskedSelect(from_mask))
t7> =x
  1
  2
  3
  4
  5
  6
  7
  8
  9
 10
[torch.DoubleTensor of dimension 10]

t7> =y
 1
 3
 5
 7
 9
 0
 0
 0
 0
 0
[torch.DoubleTensor of dimension 10]

soumith:
thanks for the clarification Koray, will fix the docs then.

PR #193, opened 2015-04-03
Revert "Add file and line number to THError."

soumith:
Reverts torch/torch7#192


soumith:
@adamlerer quite a few errors on OSX:
https://gist.github.com/soumith/397387e5d56e6578f7be

PR #197, opened 2015-04-04
Add C file/line to THError et al. Add THAssert. Improve error messages.

adamlerer:
* Add C file/line to THError and THArgCheck. Add tensor sizeStr() method.
* Add THAssert THAssertMsg functions.
* Add sizeDesc() and desc() functions, which return tensor descriptor
strings for use in error messages. To avoid memory leaks, we have to pass these
strings on the stack within structs, so it's a bit unwieldly.
* Improved a bunch of error messages.

e.g.

```
th> torch.mm(torch.Tensor(4,4), torch.Tensor(5,5))
size mismatch, m1: [4 x 4], m2: [5 x 5] at /home/alerer/git/torch7-2/lib/TH/generic/THTensorMath.c:511
stack traceback:
...
th> torch.Tensor(1):size(4)
[...]:1: bad argument #1 to 'size' (dimension 4 out of range of 1D tensor at /home/alerer/git/torch7-2/generic/Tensor.c:16)
stack traceback:
...
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,8 @@ SET(CMAKE_MODULE_PATH
   "${CMAKE_CURRENT_SOURCE_DIR}/cmake"
   "${CMAKE_MODULE_PATH}")
 
+SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror=implicit-function-declaration -Werror=format")
+
 # Flags
 # When using MSVC
 IF(MSVC)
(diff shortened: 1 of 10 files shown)


adamlerer:
This commit adds some more improvements to error reporting:

* THAssert and THAssertMsg, which I think would be nice as a replacement for THArgCheck pretty much everywhere (opinions?)
* Start using THTensor_sizeStr in every error message about mismatched sizes, invalid tensor dim, etc.
* Make it so that printing size-0 storages (and thus, size-0 tensor sizes) doesn't cause an error

adamlerer:
My latest set of changes addresses some of @andresy's concerns.

* Avoid sizeDesc() memory leak by passing string buffer on the stack
* Added desc() function that returns something like "torch.DoubleTensor of size 5x5"
* Improved a bunch of the error messages, some using sizeDesc().

adamlerer:
@soumith @andresy I think your concerns are addressed and this is ready to merge.

andresy:
looks good to me... thanks a lot!

PR #198, opened 2015-04-05
fixing travis to properly build and test

soumith:
Travis was constantly noisy and with false-positives. Also, it was completely invalid for PRs, and did not actually build the PR branch (but built master).

This PR fixes all of that, and also now travis correctly errors out on tests failing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,23 @@
 language: c
 compiler:
   - gcc
+  - clang
+cache: apt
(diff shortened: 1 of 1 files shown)


soumith:
@clementfarabet if you're interested, you can now edit the travis.yml to include the pure LUA51 build as an additional config in the build matrix. That way we can guarantee that this doesn't break for non LuaJIT users like you.

PR #200, opened 2015-04-08
Atomic

andresy:
@clementfarabet @koraykv @soumith could you look at that and tell me if you like it or not?

some atomic functions to have more fun with threads :game_die:

### DIFF
--- a/lib/TH/THAtomic.c
+++ b/lib/TH/THAtomic.c
@@ -1,0 +1,103 @@
+#include "THAtomic.h"
+
+/*
+  Note: I thank Leon Bottou for his useful comments.
+  Ronan.
+*/
+
+#if defined(USE_C11_ATOMICS)
+#include <stdatomic.h>
+#endif
+
+#if defined(USE_MSC_ATOMICS)
+#include <intrin.h>
+#endif
+
+#if !defined(USE_MSC_ATOMICS) && !defined(USE_GCC_ATOMICS) && defined(USE_PTHREAD_ATOMICS)
+#include <pthread.h>
+static pthread_mutex_t ptm = PTHREAD_MUTEX_INITIALIZER;
+#endif
+
+void THAtomicSet(int volatile *a, int newvalue)
+{
+#if defined(USE_C11_ATOMICS)
+  atomic_store(a, newvalue);
+#elif defined(USE_MSC_ATOMICS)
+  _InterlockedExchange((long*)a, newvalue);
+#elif defined(USE_GCC_ATOMICS)
+  __sync_lock_test_and_set(a, newvalue);
+#else
+  int oldvalue;
+  do {
+    oldvalue = *a;
+  } while (!THAtomicCompareAndSwap(a, oldvalue, newvalue));
+#endif
+}
+
+int THAtomicGet(int volatile *a)
+{
+#if defined(USE_C11_ATOMICS)
+  return atomic_load(a);
+#else
+  int value;
+  do {
+    value = *a;
+  } while (!THAtomicCompareAndSwap(a, value, value));
+  return value;
+#endif
+}
+
+int THAtomicAdd(int volatile *a, int value)
+{
+#if defined(USE_C11_ATOMICS)
+  return atomic_fetch_add(a, value);
+#elif defined(USE_MSC_ATOMICS)
+  return _InterlockedExchangeAdd((long*)a, value);
+#elif defined(USE_GCC_ATOMICS)
+  return __sync_fetch_and_add(a, value);
+#else
+  int oldvalue;
+  do {
+    oldvalue = *a;
+  } while (!THAtomicCompareAndSwap(a, oldvalue, (oldvalue + value)));
+  return oldvalue;
+#endif
+}
+
+void THAtomicIncrementRef(int volatile *a)
+{
+  THAtomicAdd(a, 1);
+}
+
+int THAtomicDecrementRef(int volatile *a)
+{
+  return (THAtomicAdd(a, -1) == 1);
+}
+
+int THAtomicCompareAndSwap(int volatile *a, int oldvalue, int newvalue)
+{
+#if defined(USE_C11_ATOMICS)
+  return atomic_compare_exchange_strong(a, &oldvalue, newvalue);
+#elif defined(USE_MSC_ATOMICS)
+  return (_InterlockedCompareExchange((long*)a, (long)newvalue, (long)oldvalue) == (long)oldvalue);
+#elif defined(USE_GCC_ATOMICS)
+  return __sync_bool_compare_and_swap(a, oldvalue, newvalue);
+#elif defined(USE_PTHREAD_ATOMICS)
+  int ret = 0;
+  pthread_mutex_lock(&ptm);
+  if(*a == oldvalue) {
+    *a = newvalue;
+    ret = 1;
+  }
+  pthread_mutex_unlock(&ptm);
+  return ret;
+#else
+#warning THAtomic is not thread safe
+  if(*a == oldvalue) {
+    *a = newvalue;
+    return 1;
+  }
+  else
+    return 0;
+#endif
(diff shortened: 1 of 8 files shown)


soumith:
i like it. finally we're safe! ;)

dominikgrewe:
This is great!

Just two little comments:
I don't think it's immediately obvious what the THAtomic functions return. At the very least we should add some comments documenting this.

I'm also not sure about the return values of THAtomicIncrement and THAtomicDecrement; given that THAtomicAdd and THAtomicSet return the old value it seems strange that they behave differently. I know that it's convenient for ref counting, but if those functions are only meant to be used for ref-counting, how about we rename them to something like THAtomicIncrementRef and THAtomicDecrementRef?

koraykv:
I like this! Thanks Ronan.

My only comment is similar to Dominik's second point. Can we make THAtomic(Increment/Decrement) also return old value and do the check in storage? That way it seems more consistent.

andresy:
I totally agree (and i messed up the commit, i thought i added some doc).
THAtomicIncrementRef and THAtomicDecrementRef seems good to me (I hesitated naming them like this).

andresy:
(added few comments, and explicit "Ref" for increment/decrement)

lib/TH/generic/THStorage.c
```diff
@@ -107,9 +107,9 @@ void THStorage_(free)(THStorage *storage)
   if(!storage)
     return;
 
-  if((storage->flag & TH_STORAGE_REFCOUNTED) && (storage->refcount > 0))
+  if((storage->flag & TH_STORAGE_REFCOUNTED) && (THAtomicGet(&storage->refcount) > 0))
   {
```
timharley: Isn't this AtomicGet unnecessary? The ref counting for Tensors below just jumps straight to the DecrementRef.

andresy: i hesitated to access atomic variables only with atomic functions, as an habit. (in fact, my first version had atomic as a struct, to refrain the user touching directly the number in there). true, here does not do much... i put it only for consistency reasons. you think i should remove it?

leonbottou: On Friday, April 10, 2015 10:44:05 AM Ronan Collobert wrote:
> i hesitated to access atomic variables only with atomic functions, as an
> habit. (in fact, my first version had atomic as a struct, to refrain the
> user touching directly the number in there). true, here does not do much...
> i put it only for consistency reasons. you think i should remove it?

When I started using atomics in djvu, I decided not to use atomic gets and 
sets on aligned elementary types (1,2,4,8 bytes longs things) because I do not 
believe that reasonable hardware would break such aligned accesses into 
multiple parts. I only use them for basic read-modify-write operations such as 
increment, decrement, compare-and-swap, test-and-set, etc.

It is possible to be much more refined in using atomics. One can go faster by 
adequately using acquire/release semantics to ensure correct memory ordering. 
But I never felt comfortable with these things and never got a good platform 
to test them (the intel platform is very forgiving of these). So I stay with a 
few basic read-modify-write operations...

- L.

wickedfoo: >When I started using atomics in djvu, I decided not to use atomic gets and
sets on aligned elementary types (1,2,4,8 bytes longs things) because I do not
believe that reasonable hardware would break such aligned accesses into
multiple parts. I only use them for basic read-modify-write operations such as
increment, decrement, compare-and-swap, test-and-set, etc.

FYI, now that C11/C++11 has an explicit memory model (e.g., http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf, section 5.1.2.4), this behavior is very dangerous.

The issue is not just what the hardware does, but what the *compiler* is also allowed to do.

There's no guarantee that without a memory fence, a write that one thread makes must be visible at all to any other thread, or that a set of writes will be seen in order by other threads. The compiler is completely free to do whatever it wants to with this write, as long as the current thread only sees its writes in order. In fact, the standard basically indicates that it could take an arbitrary long period of time for writes by one thread to be made visible to another thread without fences or atomics, if at all. A compiler has to make some assumption across function calls and has to issue its writes out to memory before making them (unless it is optimizing them together and has visibility into what is being called), at which point you depend upon the hardware level memory visibility rules.

Such behavior makes a lot of old patterns now dangerous (http://plv.mpi-sws.org/c11comp/popl15.pdf), though the behavior was somewhat undefined before. Java, for instance, always had a rule that writes to 8 byte values were potentially non-atomic and were prone to tearing, since the underlying hardware could provide no such guarantees.

Getting back to the issue:

>Isn't this AtomicGet unnecessary? The ref counting for Tensors below just jumps straight to the DecrementRef.
>i hesitated to access atomic variables only with atomic functions, as an habit. (in fact, my first version had atomic as a struct, to refrain the user touching directly the number in there). true, here does not do much... i put it only for consistency reasons. you think i should remove it?

You need an atomic read here, since you are writing to it atomically, and there's no memory fence for any normal read implied (an atomic write provides no general fence guarantees for all writes, just for that since value, if it is implemented as a hardware-level CAS).

Rather I'd raise the question of why you are reading the value beforehand at all (checking that storage->refcount is > 0)? Why not just atomic decrement, and if zero, free?

andresy: it is because certain storages used to have negative refcount, and untouched somehow (before i added the flag about refcount). i suppose i could remove that, not sure for backward reasons.

timharley: > Rather I'd raise the question of why you are reading the value beforehand at all (checking that storage->refcount is > 0)? Why not just atomic decrement, and if zero, free?

This was what I was trying to say, though I phrased it poorly.

What I was trying to say: the expression on the right hand side of the && is unnecessary (the read/AtomicGet). 
What I wasn't trying to say: the expression on the right hand side doesn't need to be atomic.

lib/TH/THAtomic.c
```diff
@@ -0,0 +1,90 @@
...
+  if(*a == oldvalue) {
+    *a = newvalue;
+    ret = 1;
+  }
+  pthread_mutex_unlock(&ptm);
+  return ret;
+#else
+#warning THAtomic is not thread safe
+  if(*a == oldvalue) {
+    *a = newvalue;
```
wickedfoo: And if you cared about this case at all (probably not, since it's not thread safe :) ), `volatile*` likely doesn't do what you think it does. `volatile` is only a compiler barrier, not a memory barrier. There's no guarantee that any other thread would ever see `a` being `newvalue` until you eventually hit a memory barrier somewhere else.

When a compiler sees a memory barrier (e.g., in C11, atomic_thread_fence) it will do the right thing and behave as a compiler barrier as well. A mutex contains a memory barrier, and the compiler barrier is usually the function call. Code is either correct or not correct regardless of the `volatile`.

http://lwn.net/Articles/233482/?hc_location=ufi

Also, I don't know if you want Torch7 to require only the C99 standard or the C11 standard, but if you were ok with C11, you could use stdatomic.h, which should be more portable and possibly better performance (see http://en.cppreference.com/w/c/atomic/atomic_fetch_add etc)

http://en.cppreference.com/w/c/atomic

andresy: Thanks Jeff -- I put volatile only for the compiler not to make stupid optimizations.
I saw the C11 atomics, but my understanding was it is not well spread yet. I can however add an additional CMake test to check if it there or not.

lib/TH/generic/THStorage.c
andresy: Euh, I wrote this in a hurry. Basically:
- Before we did not have TH_STORAGE_REFCOUNTED, and the way to not refcount a tensor/storage was to put the count at a negative value. The second part is thus to support this old behavior.
- The first part of the line is to support the new behavior.

I could remove the second part, as it is unlikely people are using this way now.

andresy:
@wickedfoo I added support for C11 atomic functions.

PR #203, opened 2015-04-10
Fixed inconsistency in Tensor.__tostring__

atamahjoubfar:
The string that `Tensor.__tostring__` function returns includes size of different dimensions of the tensor (`tensor:size()`), but not the dimension of the tensor (`tensor:dim()`). For example, the dimension of the `torch.rand(7)`, which is obtained by `torch.rand(7):dim()` is 1, whereas `Tensor.__tostring__` returns
`[torch.DoubleTensor of dimension 7]`,
which is incorrect. This is a tensor of dimension 1, size 7. The same problem exists for Tensors of dimension 2 and beyond. For instance, for `torch.rand(8,5)` the `Tensor.__tostring__` function returns
`[torch.DoubleTensor of dimension 8x5]`
instead of returning `[torch.DoubleTensor of dimension 2]` as expected by `torch.rand(8,5):dim()`.

To fix this issue and make the returned description consistent with the terminology of the Torch package (i.e. `tensor:dim()` and `tensor:size()`), the string `' of dimension '` has been changed to `' of size '` in all three usages.

Here is another example in a screenshot of Trepl:

small

### DIFF
--- a/Tensor.lua
+++ b/Tensor.lua
@@ -208,13 +208,13 @@ function Tensor.__tostring__(self)
                table.insert(strt, string.format(format, tensor[i]) .. '\n')
             end
          end
-         table.insert(strt, '[' .. torch.typename(self) .. ' of dimension ' .. tensor:size(1) .. ']\n')
+         table.insert(strt, '[' .. torch.typename(self) .. ' of size ' .. tensor:size(1) .. ']\n')
       elseif tensor:nDimension() == 2 then
          table.insert(strt, Tensor__printMatrix(tensor))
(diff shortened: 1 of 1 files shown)


atamahjoubfar:
If you agree with this change, I can update the doc files accordingly too.

soumith:
i think the wording "dimension" here just means that it's of that dimensionality, it's probably slightly unclear or confusing. thanks for the chance.

atamahjoubfar:
Thanks @soumith. I am new to the Torch, and the wording was quite confusing at the beginning.

PR #205, opened 2015-04-10
Add hasSize method for Tensors.

timharley:


timharley:
Changing name to `sizeIs`.

PR #206, opened 2015-04-10
Add isSize method for Tensors.

timharley:


timharley:
What are your thoughts @soumith @korayk?

koraykv:
I am good, maybe the name hasSize is better?

soumith:
sizeIs is good, isSize is also good (goes with isSameSizeAs), hasSize is also good, sizeEquals is good as well.

andresy:
I am not the English native speaker there, I let you guys choose. I can also think about isOfSize()

timharley:
+1 for `:isSize(size)`, but they all work really. Does anybody have a strong preference?

soumith:
@timharley looks like not, change it to one of the options above if you'd like and we can wrap it up.

timharley:
I've amended the commit to add `isSize`, should be good to merge.

PR #209, opened 2015-04-16
Pass optional allocator to storage.

dominikgrewe:
Allows passing in an allocator as the first parameter when creating a storage.

### DIFF
--- a/generic/Storage.c
+++ b/generic/Storage.c
@@ -4,22 +4,32 @@
 
 static int torch_Storage_(new)(lua_State *L)
 {
+  int index = 1;
   THStorage *storage;
-  if(lua_type(L, 1) == LUA_TSTRING)
+  THAllocator *allocator = luaT_toudata(L, index, "torch.Allocator");
+  if (allocator) index++;
+
+  if(lua_type(L, index) == LUA_TSTRING)
   {
-    const char *fileName = luaL_checkstring(L, 1);
-    int isShared = luaT_optboolean(L, 2, 0);
-    long size = luaL_optlong(L, 3, 0);
+    if (allocator)
+      THError("Passing allocator not supported when using file mapping");
+
+    const char *fileName = luaL_checkstring(L, index);
+    int isShared = luaT_optboolean(L, index + 1, 0);
+    long size = luaL_optlong(L, index + 2, 0);
     storage = THStorage_(newWithMapping)(fileName, size, isShared);
   }
-  else if(lua_type(L, 1) == LUA_TTABLE)
+  else if(lua_type(L, index) == LUA_TTABLE)
   {
-    long size = lua_objlen(L, 1);
+    long size = lua_objlen(L, index);
     long i;
-    storage = THStorage_(newWithSize)(size);
+    if (allocator)
+      storage = THStorage_(newWithAllocator)(size, allocator, NULL);
+    else
+      storage = THStorage_(newWithSize)(size);
     for(i = 1; i <= size; i++)
     {
-      lua_rawgeti(L, 1, i);
+      lua_rawgeti(L, index, i);
       if(!lua_isnumber(L, -1))
       {
         THStorage_(free)(storage);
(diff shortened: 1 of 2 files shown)


dominikgrewe:
Usage example: https://github.com/dominikgrewe/cutorch/tree/cuda_host_alloc

dominikgrewe:
Any comments or concerns?

soumith:
What is the actual use for cudaHostAllocator?

Should probably wait for ronan's comments. Touches the core of storages.

is a check of lightuserdata enough to determine that it's an allocator?

andresy:
While i am ok having the optional allocator as argument, i wonder if it makes sense to have it passed as a lightuserdata. In that respect, it would be reasonable to have a storage constructor which takes a real* as argument, which would be then a lightuserdata. Why not making the allocator a plain userdata?

dominikgrewe:
@soumith cudaHostAllocator allocates pinned memory. Transferring data to/from the GPU using pinned memory is significantly faster than using pageable memory. It should be used sparingly though...

@andresy I guess both lightuserdata and userdata makes sense, doesn't it? In the end, we just want a pointer to a THAllocator. I don't understand your point about real*; I'm not passing in memory for the storage, but an allocator to create memory.

dominikgrewe:
I'm now using `lua_isuserdata` to accept both light and full userdata.

andresy:
In C we have Storage constructor which takes a real* as argument. I was mentioning it could make sense that if you pass a lightuserdata, one could expect the real* constructor to be called, and not the one related to the allocator. I have seen few rare cases where this could be useful (without FFI).

One cannot typecheck lightuserdatas. So I was mentioning seeing the allocator as a userdata would be probably a better solution. I think that providing a lua interface to different allocator also makes sense.

dominikgrewe:
The case of passing in a real* is currently covered by passing in the size and the real* (as a number). I can see how it might get confusing though if we add another pointer argument.

I was hoping to avoid providing different interfaces for each allocator. Can we maybe have something like `torch.StorageWithAllocator(allocator, ...)`?

andresy:
I am not sure what you mean my "providing different interfaces". Note that at the simplest:
- you could register an empty metatable for Allocator
- given an allocator pointer, you can simply push this pointer as a userdata using luaT_pushudata or torch.pushudata from lua (if using ffi).

dominikgrewe:
I meant having separate Lua functions for each allocator. So the allocator would be completely hidden from the user and you'd handle the allocator in C only. But that doesn't seem very scalable.

Deciding whether something is a real* or a THAllocator* based on whether it's light or full userdata seems a bit unsafe. For the real* we need a size any way, so we could rely on the current method of passing in the size and the pointer (either as a number or userdata).

Or we have a completely new interface for creating a storage with an allocator as I described above.

Or we allow the user to change the default allocator (something like `torch.getAllocator` & `torch.setAllocator`)? Unfortunately the default allocator is a global, so it wouldn't be thread-safe.

What do you think?

dominikgrewe:
I've implemented one of the alternatives. `torch.*Storage` behaves as before, but there's now a `torch.*StorageWithAllocator` that takes an allocator as the first argument, followed by the usual arguments for storages.
@andresy What do you think?

dominikgrewe:
Any more thoughts on this? With the latest change the existing interface is unchanged, so it shouldn't interfere with any use cases.

dominikgrewe:
@andresy I think I misunderstood your earlier comment:
Do you mean that we could register "Allocator" as a Torch type and then we're able to check for that specific type instead of relying on a rather unsafe check for userdata?

andresy:
yes, exactly :) you can register a dummy type (with nothing in the metatable), and play with torch.pushudata if needed (or luaT_pushudata)

dominikgrewe:
I'm now registering `torch.Allocator` as a new type and check for that. The type doesn't have a constructor or destructor, because I imagine them being managed entirely in C (for example: https://github.com/torch/cutorch/compare/master...dominikgrewe:cuda_host_alloc)

@andresy Please have a look.

dominikgrewe:
@andresy Any thoughts? Are you happy with the current version?

soumith:
i think it's finally spot on to what @andresy was suggesting. merging.

soumith:
a couple of small unit tests would be good, to make sure for interface breakage in the future.

dominikgrewe:
Torch doesn't currently expose an allocator. I could create a dummy one for testing but, as far as I can tell, I'd have to fall back to C to actually assign it malloc and free methods.

mshiryaev:
Hi,
Is it possible to set custom allocator for Torch package from user script?

If so, how to pass my own allocator to be read here:
THAllocator *allocator = luaT_toudata(L, index, "torch.Allocator");

PR #210, opened 2015-04-17
Add scatter and gather operations.

dominikgrewe:
Functions to read to / write from tensor, selecting one element from
each slice along the specified dimension.

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -189,6 +189,35 @@ for _,Tensor in ipairs({"ByteTensor", "CharTensor",
          {name=Tensor},
          {name="LongArg"}})
 
+   wrap("gather",
+        cname("gather"),
+        {{name=Tensor, default=true, returned=true,
+          init=function(arg)
+                  return table.concat(
+                     {
+                        arg.__metatable.init(arg),
+                        string.format("THLongStorage* %s_size = THLongTensor_newSizeOf(%s);", arg:carg(), arg.args[4]:carg()),
+                        string.format("TH%s_resize(%s, %s_size, NULL);", Tensor, arg:carg(), arg:carg()),
+                        string.format("THLongStorage_free(%s_size);", arg:carg())
+                     }, '\n')
+               end
+         },
+         {name=Tensor},
+         {name="index"},
+         {name="IndexTensor", noreadadd=true}})
+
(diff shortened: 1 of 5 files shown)


dominikgrewe:
These functions are different from indexCopy, indexFill etc, because they select individual elements from each slice, rather than entire slices. Does anyone have better ideas for the name? Maybe `sparseIndexRead` or something along those lines?

Currently exactly one element is selected from each slice. We could generalize that to multiple elements if needed.

Let me know what you think.

dominikgrewe:
Talking to @koraykv and @GeorgOstrovski we thought `reduceSelect`, `reduceCopy` and `reduceFill` would be more descriptive and in line with functions such as `indexSelect`. I'm planning to update this PR accordingly and also support indexing more than one element in each slice.

soumith:
How about broadcastRead, broadcastWrite, broadcastFill, because you are applying the same operation to each row (or slice), and that is called broadcasting in other places, like numpy.

dominikgrewe:
Isn't broadcasting more about applying a single value to each value within a row, similar to what you'd do with an expand in Torch? These operations feel more like a scatter/gather (see http://en.wikipedia.org/wiki/Gather-scatter_%28vector_addressing%29).

nicholas-leonard:
I like reduce.

dominikgrewe:
I'm actually not sure about reduce any more. It makes sense for the select case, where each row is reduced to a fixed number of elements. But it doesn't really work for copy and fill. I prefer scatter/gather.

dominikgrewe:
I've updated the PR:
The functions are now called scatter and gather (see previous comment) and support reading/writing more than one element per row. Any thoughts?

soumith:
fine with me. ping @andresy

soumith:
it looks good overall. I think the tests need to be randomized over the dimension as well. Right now the unit tests do not hit the corner cases, dim1 and dimN. It is good that they do hit the middle dimensions.

dominikgrewe:
@soumith Thanks. I've updated the tests to pick a random dimension. Let me know if you're happy with that and I'll squash the commits.

soumith:
looks  good.

dominikgrewe:
Thanks. Squashed and rebased.

PR #211, opened 2015-04-17
Fix argument checks

GeorgOstrovski:
Some of the argument numbers passed to THArgCheck seemed to be off by one, creating confusing error messages. Also, dimension argument wasn't checked to be >= 0, e.g. in THTensor_(squeeze1d) a negative dimension could actually be used leading to segfault.

[ my editor automatically removed trailing whitespace - sorry for the noise ]


soumith:
Thansk @GeorgOstrovski

PR #212, opened 2015-04-19
expose storage and tensor retain() and free() methods into Lua

andresy:


soumith:
Looks good. Not much to discuss. Merged.

PR #213, opened 2015-04-22
Add elementSize method to *Storage types.

timharley:


timharley:
What are your thoughts @soumith @koraykv ?

soumith:
I like it. I had quite a few people ask for this.

timharley:
It might be worth adding the same method to tensors too?

soumith:
I think for tensors makes sense as well.

PR #214, opened 2015-04-22
fixed maskedCopy to accept src with size greater than the num of ones in mask

soumith:
also added small test.


soumith:
This is exactly as described in the docs. (finally)

PR #216, opened 2015-04-24
Fix bug in split while reusing the result table.

fmassa:
There was a small bug in `split` when the `result` table was given as input to be reused, as explained in the documentation. Instead of putting the output in `result`, it was only emptying it. It passed the tests because the size of `results` was 0, so it didn't check the assertions.
Added a few more test cases to compare both outputs.
Additionally, I set the `precision` variable to be local in `test.lua`

### DIFF
--- a/Tensor.lua
+++ b/Tensor.lua
@@ -492,15 +492,14 @@ function Tensor.split(result, tensor, splitSize, dim)
       end
    end
    dim = dim or 1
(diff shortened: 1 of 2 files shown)


soumith:
Thanks, should have caught the bug in the first code review itself. My bad.

nicholas-leonard:
Thanks.

PR #221, opened 2015-04-27
Add :elementSize() for Tensors.

timharley:
* Add `:elementSize()` method to Tensors.
* Slight refactor to use new methods for handling Tensor sizes where possible.
* Squelch some trailing whitespace.


soumith:
thanks Tim

PR #223, opened 2015-04-28
Support luaffi.

colesbury:
Add support for LuaFFI:

1) Try requiring "ffi" when not running in luajit
2) Lua doesn't have a built-in cdata type, so add a function to check if a lua value is cdata
3) The "ctype" in luaffi do not have a unique tonumber, unlike in luajit. We use tostring to create a unique representation for torch.metatype.

### DIFF
--- a/FFI.lua
+++ b/FFI.lua
@@ -1,7 +1,5 @@
-if jit then
-
-   local ffi = require 'ffi'
-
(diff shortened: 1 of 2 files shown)


colesbury:
Subsumed by #225

PR #225, opened 2015-04-29
Add support for Lua 5.2

colesbury:


andresy:
i see this overlaps with #223, right?
note that there are some other issues when going to 5.2: e.g. math.log10 does not exist anymore

soumith:
@andresy looks like @colesbury fixed a bunch of issues. Relook!?

andresy:
yes, i am closely looking at what is being done here :)
still some tests to do

andresy:
So one has to understand that this PR includes changing the way function upvalues are serialized. It should not be a problem in general (because it is rare that we write function on disk), but it is a change which might affect people who have models saved since a long time, and containing functions with upvalues. In fact, it also fixes the serialization of those upvalues which was bogus.

@koraykv @clementfarabet is that ok for you?

soumith:
ping @koraykv @clementfarabet

colesbury:
@koraykv @clementfarabet, how do you feel about changing the serialization format for upvalues? If it's a problem, I can probably rework that part of this change.

clementfarabet:
Mildly annoying because we'll have to update a couple of serialized objects, but support for 5.2 is more important.

colesbury:
@koraykv, is this okay with you?

koraykv:
checking now

koraykv:
So sorry for the late reply. I have looked through it now. We actually serialize a lot of functions with upvalues and this change would break all those.

Can't we make this change backward compatible by putting a version marker in the upvalue table and checking for that?

@dominikgrewe  can you take a look if there is anything else that would break.

andresy:
Koray: note that the current way of saving upvalues is buggy when you have nil upvalues.

koraykv:
Yes, I see that. But I think we should fix that and add 5.2 compat without breaking all the old files.

koraykv:
To be clear, I suggest to write all the new files with the new version, but be able to read the old version too. This way, a read/write will fix old files and we will be writing without the bug from now on.

dominikgrewe:
Looks good to me in principle. I agree with Koray though. Reading an old file with this code will fail (regardless of whether there were any nil upvalues), because upvalues are now stored as tables (name + value) rather than just values.

colesbury:
@koraykv, please take another look

I've updated the PR to define type "8" for the new serialization format for closures. Lua 5.1 can read the old format and write the new format.

NOTE: Functions serialized serialized in Lua 5.1 cannot be loaded in Lua 5.2, although that's not because the function body is incompatible.

Here's a simple test to read and write closures:
https://gist.github.com/colesbury/63af043f190e93c04b84
https://gist.github.com/colesbury/19dd14813a5dc11d86a2

koraykv:
I think this looks fine now. Thanks a lot. One last comment. I don't think you want to do the change at line 214 (https://github.com/torch/torch7/pull/225/files#diff-3207f2f64277bb1178d658f13456dc3aL214). TYPE_FUNCTION is not used anymore, it was replaced by TYPE_RECUR_FUNCTION and was kept just for backward compatibility, similar to what you did. You can see that File:isWritableObject does not return a TYPE_FUNCTION.

colesbury:
Removed change to TYPE_FUNCTION deserialization

colesbury:
@andresy, is this good with you? With this change we can now read the old format and read and write the new, compatible with Lua 5.2 format for serializing functions.

andresy:
yes, awesome, thanks a ton for this :+1:

PR #230, opened 2015-05-12
Add LAPACK QR decomposition to Torch.

d11:
Three functions are added: torch.geqrf() and torch.orgqr(), which
call the corresponding LAPACK functions directly, and torch.qr(),
which is just a convenience function that calls them both correctly
and puts the results in a more intuitive form.

Tests and documentation are included.


test/test_qr.lua
```diff
@@ -0,0 +1,254 @@
...
+      {-12.8840987267251261, -14.5916298832790581, -17.0753115655393231},
+      {  0,                  -1.0413152017509357,  -1.770235842976589 },
+      {  0,                   0,                    0.5477225575051664}
+  }
+  checkQRWithExpected(testOpts, a, expected_q, expected_r)
+end, {doubleTensorOnly=true})
+
+-- Decomposing a sequence of medium-sized random matrices.
+addTestVariations(tests, 'randomMediumQR', function(testOpts)
+  torch.manualSeed(1)
```
soumith: if you want to control the seed for this test, can you restore the seed after? we might lose randomized tests if this test executes first.

test/test_qr.lua
```diff
@@ -0,0 +1,254 @@
...
+      local n = math.pow(2, y)
+      local x = torch.rand(m, n)
+      checkQR(testOpts, x:typeAs(testOpts.tensorFunc()))
+    end
+  end
+end)
+
+-- Decomposing a sequence of small random matrices.
+addTestVariations(tests, 'randomSmallQR', function(testOpts)
+  torch.manualSeed(1)
```
soumith: same here

test/test_qr.lua
```diff
@@ -0,0 +1,254 @@
...
+  for m = 1, 40 do
+    for n = 1, 40 do
+      checkQR(testOpts, torch.rand(m, n):typeAs(testOpts.tensorFunc()))
+    end
+  end
+end)
+
+-- Decomposing a sequence of small matrices that are not contiguous in memory.
+addTestVariations(tests, 'randomNonContiguous', function(testOpts)
+  torch.manualSeed(1)
```
soumith: and here.

soumith:
looks great, except for a couple of small comments.

d11:
Good point! I've updated it to restore the RNG state after each test now. Cheers!

soumith:
thanks Dan!

PR #238, opened 2015-05-22
Use assignment in indexSelect for 1-dimensional contiguous input.

dominikgrewe:
Gives another 2-3x over memcpy for large index vectors.


soumith:
nice

PR #239, opened 2015-05-22
Random number generator for a custom discrete distribution.

zakattacktwitter:


timharley:
Sorry if I'm mistaken, but doesn't torch.multinomial already do this? 
Some simple benchmarks also show torch.multinomial to be considerably faster on my machine.
(Admittedly, torch.multinomial has the wrong name).

clementfarabet:
So `torch.distribution` is implemented with the alias method. There's a cost to create the lookup data structure, but then sampling is pretty much free. It's particularly useful for distributions with tons of outcomes (>1e6). Example:

```lua

-- Probs:
local p = torch.Tensor(10e6):random(1,100)
p:div(p:sum())

-- Build sampler:
local d = torch.distribution(p)

-- Sample:
sys.tic()
local s = torch.multinomial(p,100,true)
print('sampling with multinomial:', sys.toc(), 'sec')

sys.tic()
local s = d(100)
print('sampling with distribution:', sys.toc(), 'sec')

-- will print:
sampling with multinomial:      0.17442488670349        sec
sampling with distribution:     0.00027298927307129     sec
```

So sampling is essentially free in this case (large outcome distribution, need to sample few instances in a batch).

But, it's true that the sampler itself would benefit being written in C, because the loop (line 67) is obviously inefficient...

clementfarabet:
So ideally, the sampler itself (lines 55-85) would need to be in C so that sampling into a very large tensor remains cheap (to avoid the lua for loop). But the data struct prep before is a 1 time cost, so fine if that's still lua. 

Training large-scale word models and other systems where the updates are extremely cheap (< 1ms) requires the sampler itself to be really cheap too (< 0.1ms if possible). That's why we use this type of sampler. 

Not sure how to reconcile both approaches though. Maybe we should extend multinomial() and have an optional flag for this method?

timharley:
Would precomputing the samples work?
```lua
local probs = torch.rand(1e7)
local gen = torch.distribution(probs) -- Takes ~30 seconds (precomputed)
local samples1 = gen(1e6) -- Takes ~1.2s
local samples2 = gen(1e2) -- Takes ~0.14ms (average over 1e5 runs)

local samples3 = torch.multinomial(probs, 1e6, true) -- Takes ~0.9s (but could be precomputed)
local samples4 = torch.multinomial(probs, 1e2, true) -- Takes ~0.2s

local samples5 = samples3:narrow(1, 1, 1e2) -- Basically free.
```

clementfarabet:
That works indeed. Ok closing this one.

clementfarabet:
Oh one last thing: i actually didn't know we had a multinomial function: should we maybe move its doc to the random section instead of math?

PR #240, opened 2015-05-22
Adding support for Storage views and exposing Storage type.

zakattacktwitter:
This PR allows storage instances to be a view of another storage. It is similar to the way tensors can share the same underlying storage.

All the THStorage->type related code has been removed.

### DIFF
--- a/generic/Storage.c
+++ b/generic/Storage.c
@@ -39,6 +39,27 @@ static int torch_Storage_(new)(lua_State *L)
       lua_pop(L, 1);
     }
   }
+  else if(lua_type(L, index) == LUA_TUSERDATA)
+  {
+    THStorage *src = luaT_checkudata(L, index, torch_Storage);
+    if (src->flag & TH_STORAGE_VIEW) {
+      // don't allow chained views
+      src = src->view;
+    }
+    real *ptr = src->data;
+    long offset = luaL_optlong(L, index + 1, 1) - 1;
+    if (offset < 0 || offset >= src->size) {
+      luaL_error(L, "offset out of bounds");
+    }
+    long size = luaL_optlong(L, index + 2, src->size - offset);
+    if (size < 1 || size > (src->size - offset)) {
+      luaL_error(L, "size out of bounds");
+    }
+    storage = THStorage_(newWithData)(ptr + offset, size);
+    storage->flag = TH_STORAGE_REFCOUNTED | TH_STORAGE_VIEW;
+    storage->view = src;
+    THStorage_(retain)(storage->view);
+  }
   else if(lua_type(L, index + 1) == LUA_TNUMBER)
   {
     long size = luaL_optlong(L, index, 0);
@@ -47,7 +68,7 @@ static int torch_Storage_(new)(lua_State *L)
       storage = THStorage_(newWithDataAndAllocator)(ptr, size, allocator, NULL);
     else
       storage = THStorage_(newWithData)(ptr, size);
-    storage->flag = 0;
+    storage->flag = TH_STORAGE_REFCOUNTED;
   }
   else
   {
--- a/generic/Storage.c
+++ b/generic/Storage.c
@@ -45,10 +45,6 @@ static int torch_Storage_(new)(lua_State *L)
       THError("Passing allocator not supported when using storage views");
 
     THStorage *src = luaT_checkudata(L, index, torch_Storage);
-    if (src->flag & TH_STORAGE_VIEW) {
-      // don't allow chained views
-      src = src->view;
-    }
     real *ptr = src->data;
     long offset = luaL_optlong(L, index + 1, 1) - 1;
     if (offset < 0 || offset >= src->size) {
--- a/generic/Storage.c
+++ b/generic/Storage.c
@@ -41,6 +41,9 @@ static int torch_Storage_(new)(lua_State *L)
   }
   else if(lua_type(L, index) == LUA_TUSERDATA)
   {
+    if (allocator)
+      THError("Passing allocator not supported when using storage views");
+
     THStorage *src = luaT_checkudata(L, index, torch_Storage);
     if (src->flag & TH_STORAGE_VIEW) {
       // don't allow chained views
@@ -231,7 +234,7 @@ static int torch_Storage_(write)(lua_State *L)
 {
   THStorage *storage = luaT_checkudata(L, 1, torch_Storage);
   THFile *file = luaT_checkudata(L, 2, "torch.File");
- 
+
   THFile_writeLongScalar(file, storage->size);
   THFile_writeRealRaw(file, storage->data, storage->size);
 
--- a/lib/TH/generic/THStorage.h
+++ b/lib/TH/generic/THStorage.h
@@ -19,6 +19,7 @@
 #define TH_STORAGE_REFCOUNTED 1
 #define TH_STORAGE_RESIZABLE  2
 #define TH_STORAGE_FREEMEM    4
+#define TH_STORAGE_VIEW       8
 
 typedef struct THStorage
 {
@@ -28,6 +29,8 @@ typedef struct THStorage
     char flag;
     THAllocator *allocator;
     void *allocatorContext;
+    int type;
+    struct THStorage *view;
 } THStorage;
 
 TH_API real* THStorage_(data)(const THStorage*);
--- a/lib/TH/generic/THStorage.h
+++ b/lib/TH/generic/THStorage.h
@@ -8,10 +8,10 @@
 
    Pb: THMapStorage is kind of a class
    THLab_()... comment je m'en sors?
-   
+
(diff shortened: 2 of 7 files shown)


soumith:
Could you elaborate or document on what this change brings in.

clementfarabet:
The goal is to enable views for storages, without having to use tensors. It's particularly useful for networking/serialization, where you have a buffer (ByteStorage) that holds multiple binary blobs (jpegs, ...), and you want to just view a subset of it. Obviously Tensors provide that functionality, but it's handy to have storages be able to do this as well. 

It enables:
```lua
s = torch.ByteStorage(10)
s:fill(1)
s2 = torch.ByteStorage(s, 3,4)
s2:fill(10)
print(s)
```

clementfarabet:
Adding doc for this.

generic/Storage.c
```diff
@@ -47,7 +68,7 @@ static int torch_Storage_(new)(lua_State *L)
       storage = THStorage_(newWithDataAndAllocator)(ptr, size, allocator, NULL);
     else
       storage = THStorage_(newWithData)(ptr, size);
-    storage->flag = 0;
+    storage->flag = TH_STORAGE_REFCOUNTED;
```
zakattacktwitter: @clementfarabet What did you mean by this? Its not changed like this in cutorch.

clementfarabet: The original code here was wrong: even if you create a storage from an existing data buffer (externally managed), you still want to be refcounted so that the storage struct can get freed. You just don't want to have TH_STORAGE_FREEMEM.

lib/TH/generic/THStorage.h
```diff
@@ -28,6 +29,8 @@ typedef struct THStorage
     char flag;
     THAllocator *allocator;
     void *allocatorContext;
+    int type;
```
dominikgrewe: What is this new `type` variable used for?

generic/Storage.c
```diff
@@ -39,6 +39,27 @@ static int torch_Storage_(new)(lua_State *L)
       lua_pop(L, 1);
     }
   }
+  else if(lua_type(L, index) == LUA_TUSERDATA)
+  {
+    THStorage *src = luaT_checkudata(L, index, torch_Storage);
+    if (src->flag & TH_STORAGE_VIEW) {
+      // don't allow chained views
+      src = src->view;
```
dominikgrewe: This doesn't seem very intuitive. If you don't want to support chained views, why no throw an error? What is the problem with chained views in the first place?

generic/Storage.c
```diff
@@ -39,6 +39,27 @@ static int torch_Storage_(new)(lua_State *L)
       lua_pop(L, 1);
     }
   }
+  else if(lua_type(L, index) == LUA_TUSERDATA)
+  {
+    THStorage *src = luaT_checkudata(L, index, torch_Storage);
```
dominikgrewe: Maybe check that allocator is NULL. It doesn't make sense to pass an allocator when creating a view.

lib/TH/generic/THStorage.h
dominikgrewe: Sorry, just saw the comment in the description. What I don't like about it is the `TH_Cuda` type. That should really be in cutorch, shouldn't it?

dominikgrewe: Actually, I still don't understand why this is needed. If you have C code that can handle arbitrary storages, you should know that type of storage you're dealing with, don't you? Maybe you can elaborate a bit more on this?

generic/Storage.c
clementfarabet: This code works, no need to throw an error: if you have 3 storages, A, B and C, with A=>B, and then create C=>A, it just replaces it by C=>B. No need to chain because it makes the recounting more tricky, and doesn't add any functionality. It's strictly equivalent to have all storages point to the same end storage.

lib/TH/generic/THStorage.h
zakattacktwitter: When we pass this storage data to an underlying transport mechanism (MPI or Libfabric) in a multi-machine environment it needs to know the type of data being transferred. For example if we pass a storage of floats from a little endian to a big endian machine the transport would re-order the bytes of each float for us.

We could maybe do it without storage->type but it would require us knowing machine endian-ness at some other level of the stack, not that easy to do.

generic/Storage.c
dominikgrewe: But that doesn't seem very intuitive to me. Creating a view on A is different from creating a view on B, because A may have an offset into B, which is ignored in this case, isn't it?

lib/TH/generic/THStorage.h
clementfarabet: So the main reason for this is the (not so great) templating system in C. Our torch env is wrapped in different types of environments (Java/Scala/JNI), and the way these wrappers are written is without templating. They discover the type of the storage at runtime, and then act appropriately. 

It's pretty useful for us.

About the Cuda type, it was convenient to have them grouped in a same header, but fair point.

timharley: The type of an element is already encoded in the type of the C struct: `THStorage` is part of a macro that gets expanded to `THFloatStorage`, `THByteStorage`, etc. See [here](https://github.com/torch/torch7/blob/4fd73c14285dc4a4d787276dd78dd1ef63542628/lib/TH/THStorage.h#L7).

generic/Storage.c
zakattacktwitter: Done!

lib/TH/generic/THStorage.h
zakattacktwitter: So given a pointer to a THStorage, how would you know it is float or double, etc?

  THStorage *storage = luaT_checkudata(L, 1, torch_Storage);
  // what type of data is in storage?

generic/Storage.c
clementfarabet: Oh I see. Yep, good point, not super intuitive. So we can still keep that code, but set the ptr to the original src. This way we still don't have to do chained references, but the pointer will use the chained offset. It would look like:

```
real *ptr = src->data;
if (src->flag & TH_STORAGE_VIEW) {
  src = src->view;
}
```

Essentially just moving the ptr set above. Would that be ok?

generic/Storage.c
```diff
@@ -39,6 +39,30 @@ static int torch_Storage_(new)(lua_State *L)
...
+  {
+    if (allocator)
+      THError("Passing allocator not supported when using storage views");
+
+    THStorage *src = luaT_checkudata(L, index, torch_Storage);
+    if (src->flag & TH_STORAGE_VIEW) {
+      // don't allow chained views
+      src = src->view;
+    }
+    real *ptr = src->data;
```
clementfarabet: actually not that trivial to enable cascading. Need to think about it a bit.

lib/TH/generic/THStorage.h
timharley: Sorry, I deleted my comment as Clement wrote a comment whilst I was writing mine and it didn't seem valid any more.

In a generic file (i.e. one that uses THStorage macros), the macro "real" expands to the element type of the Storage, e.g. float, double, int, etc. See [here](https://github.com/torch/torch7/blob/master/lib/TH/THGenerateAllTypes.h#L5) for more details. 
i.e.
```C
THStorage *storage = luaT_checkudata(L, 1, torch_Storage);
real item; // This is a float if storage is a THFloatStorage, an int if it's a THIntStorage, etc.
```

Alternatively `#ifdef TH_REAL_IS_BYTE` or similar (defined in the same header) could probably help.

Just to be clear, (sorry if you know all this) THStorage is not an actual type, but part of the macro templating system.

zakattacktwitter:
Hi Folks,

I removed the type variable from THStorage. This PR is now just storage views, it should be much less controversial. Lemme know if it needs anything else.

Thanks!!

generic/Storage.c
```diff
@@ -39,6 +39,26 @@ static int torch_Storage_(new)(lua_State *L)
       lua_pop(L, 1);
     }
   }
+  else if(lua_type(L, index) == LUA_TUSERDATA)
+  {
+    if (allocator)
+      THError("Passing allocator not supported when using storage views");
+
+    THStorage *src = luaT_checkudata(L, index, torch_Storage);
+    real *ptr = src->data;
```
clementfarabet: @dominikgrewe we removed the check here, it turns out that cascading views actually works fine, and has the behavior you would expect (cascaded offsets).

dominikgrewe:
Thanks guys. Looks good to me now. Can you maybe just add some tests?

clementfarabet:
Test added. Merging. Thanks for the reviews.

dominikgrewe:
Thanks. Can you squash the commits before merging next time?

clementfarabet:
Oops, didn't see it had multiple commits. Will do.

PR #241, opened 2015-05-22
SSE optimizations for 5x5 convolution.

zakattacktwitter:
This is a pretty specific optimization to Twitter's usage of 5x5 but it could be extended to support more sizes in the future.

### DIFF
--- a/lib/TH/CMakeLists.txt
+++ b/lib/TH/CMakeLists.txt
@@ -89,9 +89,20 @@ ENDIF(C_SSE3_FOUND)
 IF(C_SSE4_1_FOUND)
   SET(CMAKE_C_FLAGS "${C_SSE4_1_FLAGS} -DUSE_SSE4_1 ${CMAKE_C_FLAGS}")
 ENDIF(C_SSE4_1_FOUND)
+IF(C_AVX_FOUND OR C_SSE4_2_FOUND)
+  SET(simd generic/simd/convolve.c)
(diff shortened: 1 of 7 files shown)


zakattacktwitter:
Does anyone have any opinions on this? We also have a fast-path for 3x3 convolutions.

soumith:
no opinion. it looks good!
Would love to have 3x3 as well.
Thanks for putting it in the generic/simd folder. That makes it clear (unlike THVector.h).

clementfarabet:
Great, let's add the 3x3 kernels in then.

andresy:
More or less same comment than Soumith': looks great, would love to have more like those. The simd directory is indeed a good idea. The only question is do we want this in the core, or as a package (like simd ;))?

dominikgrewe:
I think it's great too! Some tests would be good. And can we hook this up to Lua somehow?

I don't understand why the code is in the "generic" folder though if it's only for float.

soumith:
@dominikgrewe it looks like both Float and Double.

soumith:
or extendable to be so. because the sse instructions are macro-templated as well.

clementfarabet:
The macros are Float only actually, that's right. I guess we put them in generic because they're loaded by THTensorConv, and we prob want to have a double version at some point. 

Also we only did 3x3 and 5x5 because it's the only two kernels we use for everything.

soumith:
I'll give the final merge call on this to @andresy , this will establish our directory and file structure for pushing SIMD optimizations going forward, so needs a bit of thought.

soumith:
these aren't hooked into the lua side yet seems like it... coming in a later PR?

PR #244, opened 2015-05-26
Allowing usage of posix_memalign to be disabled through a new flag, DISABLE_POSIX_MEMALIGN

yozw:
The usage of posix_memalign causes us problems with tcmalloc. I am adding a flag to disable its usage.

### DIFF
--- a/lib/TH/THGeneral.c
+++ b/lib/TH/THGeneral.c
@@ -103,7 +103,7 @@ void* THAlloc(long size)
 
   if (size > 5120)
   {
-#if defined(__unix) || defined(__APPLE__)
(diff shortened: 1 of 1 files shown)


soumith:
thanks. Anything we can do to make it work for you as well? Not sure what the issue with tcmalloc is!

yozw:
Yes, maybe it is good to document this. The problem is that tcmalloc's realloc function fails when called with a pointer that was allocated by posix_memalign. The error message looks like:

RAW: realloc/memalign mismatch at <address>: non-NULL pointers passed to realloc must be obtained from malloc, calloc, or realloc.

We're fine with the current solution, but I might spend some time later on to see if we can solve this properly.

yozw:
Thanks for the quick reply, btw.

PR #245, opened 2015-05-27
Adding support for a static build of torch.

zakattacktwitter:
Sometimes you need a static build of torch, this allows us to build one.


soumith:
that's nice!

clementfarabet:
note: it also means that every package built with this torch cmake macro will also benefit from this double build and produce a .a on top of the .so/.dylib.

soumith:
but does the addition to cmake/TorchPackage.cmake work universally? I dont think so. It works for this package, but the macro is used by many other packages...

clementfarabet:
so it will build a .a for every other package, but you typically need to add "-DCMAKE_C_FLAGS=-fPIC -DCMAKE_CXX_FLAGS=-fPIC" to your cmake command to have a position-independent archive. So it doesn't solve the entire thing.

clementfarabet:
I guess we could set these flags in the TorchPackage.cmake file to force it for every package, but this might have side effects.

zakattacktwitter:
We do set them, it is part of the PR.

soumith:
@zakattacktwitter they are only set for this package. not for other packages. Let me quickly try to build nn after this merge. I think it might fail.

clementfarabet:
well they're set for this package (torch), but for other packages (like
sys) that use this macro to build themselves, they also have to set it in
their rock spec to be position independent. So it's not a complete solution
for 3rd party packages, but works for torch.

On Wed, May 27, 2015 at 11:58 AM, zakattacktwitter <notifications@github.com

soumith:
how about if we add the -fPIC flags in TorchPackage.cmake itself...

what side effects do you see then?

clementfarabet:
It will work but the newly generated libnn_static.a will be unusable without the addition of "-DCMAKE_C_FLAGS=-fPIC -DCMAKE_CXX_FLAGS=-fPIC" to the cmake command.

soumith:
yea, nn compiles, and the static lib does generate, but the -fPIC flags are needed to make it usable. 
Can we just add them to TorchPackage.cmake itself so that we dont need to change like 300 rockspecs...

clementfarabet:
no it's the right thing to do, I don't think there's any side effect. It's actually two separate targets, so we just have to add the -fPIC to that _static target.

soumith:
ok, so right now in this PR, the -fPIC flag is added to both targets, not just the static target.

clementfarabet:
making another PR.

clementfarabet:
this should do it: https://github.com/torch/torch7/pull/246

Atcold:
What does "static build of torch" mean? :frowning:

ajtulloch:
http://en.wikipedia.org/wiki/Static_build

Atcold:
:+1:

Atcold:
Wikipedia was not that helpful, but shortly put, is it about optimising the calls to the libraries?
Basically, @zakattacktwitter when exactly "you need a static build of torch"? Which case scenario?

soumith:
@Atcold let's take this offline. this question is super-basic.

PR #247, opened 2015-05-27
Adding elementType to complement elementSize.

zakattacktwitter:
This is a very simple version of RTTI for Tensor/Storage, it is very useful when interfacing with other systems, in particular where we need to marshal storage data in a type safe way.

### DIFF
--- a/doc/storage.md
+++ b/doc/storage.md
@@ -287,3 +287,21 @@ Increment the reference counter of the storage.
 
 Decrement the reference counter of the storage. Free the storage if the
 counter is at 0.
+
+<a name="torch.Storage.elementSize"/>
+### [number] elementSize() ###
+
+Returns the size in bytes of an element in the storage. This is platform
+dependent and is not guaranteed to be the same everywhere.
+
+<a name="torch.Storage.elementType"/>
+### [number] elementType() ###
+
+Returns a string describing the type of element in the storage. This
+is the closest thing to RTTI for storage instantiations.
+
+```lua
+> x = torch.CharStorage()
+> print(x:elementType())
+Char
+```
--- a/doc/tensor.md
+++ b/doc/tensor.md
@@ -2179,3 +2179,21 @@ Increment the reference counter of the tensor.
 
 Decrement the reference counter of the tensor. Free the tensor if the
 counter is at 0.
+
+<a name="torch.Storage.elementSize"/>
+### [number] elementSize() ###
+
+Returns the size in bytes of an element in the storage. This is platform
+dependent and is not guaranteed to be the same everywhere.
+
+<a name="torch.Storage.elementType"/>
+### [number] elementType() ###
+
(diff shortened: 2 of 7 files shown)


timharley:
Was `torch.typename()` not suitable for your purposes?

Alternatively: `ffi.typeof(tensor:data())` gives ctype information too.

soumith:
@timharley they dont use LuaJIT, so ffi.typeof is off the cards

clementfarabet:
On the lua side, we could use typename(), but we really need this on the C side, to ease our lives when interfacing from external libraries. e.g. pass a (void *) and discover the type automatically, without having to use templates.

andresy:
does this need to be in C (it could be easily added in pure lua)

clementfarabet:
I guess our comments crossed, yes it needs to be in C, in fact, it's mostly the C part that we care about. The basic thing we're trying to solve with this is: when embedding Lua/torch in an other framework (Java/Scala in our case), we want to avoid using the C generic macros, and just have a way to discover/validate a storage/tensor's element type on the fly. Right now we don't have this ability. Does it make sense?

soumith:
I think that makes a lot of sense, now that this is in the context of pure TH.

andresy:
yes, makes sense -- not sure for the Char vs char, etc, ..., especially if we add new C types in there (and then what about Cuda? And if we have Float and Doubles later in Cuda?)

soumith:
we are definitely adding new cuda types as soon as possible. (even if we dont define all math operations on every one of them). So a little thought into the future wont hurt here.

zakattacktwitter:
Certainly, types like "Half" should just work with this.

timharley:
@clementfarabet I'm not sure this PR helps if you only have a `void *`, you still need to know the underlying type to be able to call the appropriate implementation of `elementType()`, e.g. `THFloatStorage_elementType()`.

Perhaps one could implement something like a bit like a fat pointer or a box:
```C
struct AbstractStorage {
  void * storage;
  int type; /* like the type member you had in the StorageView PR */
  /* Maybe a vtable too? */
};
```
This wouldn't require any additions to TH, it could just sit on top of it.

andresy:
What i meant is that i would expect elementType to return the C type. In which case "float" or "fp16" might be clearer than "Float" or "Half". In this case, you would be missing the difference between a tensor on CPU, or a tensor on GPU, which might require another method.

If you stick to Float, Double, Cuda, etc... then you can differentiate a Cuda tensor from a CPU one, but the returned type is not a C name anymore. Not sure which one makes more sense.

zakattacktwitter:
I guess we can do it ourselves by wrapping our arbitrary lua code with a simple function. We know what return type we expect, its just a matter of verifying it. So simply adding little lua function that wraps the inner function and verifies its return type would be sufficient.

zakattacktwitter:
Thanks guys, going to discard this.

PR #252, opened 2015-06-03
Tweak torch.range to be more numerically robust.

d11:
Otherwise it can produce mildly surprising behaviour, in the floating-point setting, due to rounding errors.

For example, we get:

    > torch.FloatTensor():range(0.6, 0.9, 0.1)
    0.6000
    0.7000
    0.8000
    [torch.FloatTensor of size 3]

When we'd expect:

     > torch.range(0.6, 0.9, 0.1)
     0.6000
     0.7000
     0.8000
     0.9000
     [torch.FloatTensor of size 4]

This happens because the size for the result tensor is computed as 

    size = (long)((xmax-xmin)/step+1);

and in single-precision this gives 3. We can avoid this by
* changing the input types to accreal, so that the size is computed at double precision even for FloatTensors.
* dividing by step before subtracting xmin from xmax, which seems to be more robust to rounding error.

Both appear to be necessary in my tests...

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -483,9 +483,9 @@ for _,Tensor in ipairs({"ByteTensor", "CharTensor",
    wrap("range",
         cname("range"),
         {{name=Tensor, default=true, returned=true, method={default='nil'}},
-         {name=real},
-         {name=real},
-         {name=real, default=1}})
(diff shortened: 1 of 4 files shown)


soumith:
thanks!

PR #254, opened 2015-06-08
FindBLAS.cmake: fix missing stdlib.h include

deltheil:
This is to make sure BLAS F2C auto detection does not fail because of compiler warnings:

    error: implicit declaration of function 'exit'

I discovered this problem while re-installing Torch on Mac OS X (10.10.3) with [GCC 5](https://github.com/Homebrew/homebrew-versions/blob/316dddf599f0c34a224374ef1ee6267cea6d1364/gcc5.rb). The dot product between two `float` tensors returned 0.

Looking at cmake outputs I got:

    -- Performing Test BLAS_F2C_DOUBLE_WORKS
    -- Performing Test BLAS_F2C_DOUBLE_WORKS - Failed
    -- Performing Test BLAS_F2C_FLOAT_WORKS
    -- Performing Test BLAS_F2C_FLOAT_WORKS - Failed

And CMakeError.log indicated that these auto detection programs failed because of `some warnings being treated as errors`.

I extended the unit test to make sure there is coverage for `sdot`. This extended test properly fails for my setup (OS X 10.10.3 / GCC 5) prior to this fix.

### DIFF
--- a/lib/TH/cmake/FindBLAS.cmake
+++ b/lib/TH/cmake/FindBLAS.cmake
@@ -243,6 +243,7 @@ endif()
 IF (BLAS_LIBRARIES)
   SET(CMAKE_REQUIRED_LIBRARIES ${BLAS_LIBRARIES})
   CHECK_C_SOURCE_RUNS("
+#include <stdlib.h>
 #include <stdio.h>
 float x[4] = { 1, 2, 3, 4 };
 float y[4] = { .1, .01, .001, .0001 };
@@ -255,6 +256,7 @@ int main() {
   exit((float)r != (float).1234);
 }" BLAS_F2C_DOUBLE_WORKS )
   CHECK_C_SOURCE_RUNS("
+#include <stdlib.h>
 #include <stdio.h>
 float x[4] = { 1, 2, 3, 4 };
 float y[4] = { .1, .01, .001, .0001 };
--- a/test/test.lua
+++ b/test/test.lua
@@ -16,19 +16,25 @@ local function maxdiff(x,y)
 end
 
 function torchtest.dot()
-   local v1 = torch.randn(100)
-   local v2 = torch.randn(100)
+   local types = {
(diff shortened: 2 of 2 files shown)


soumith:
> The dot product between two float tensors returned 0.

Hmmm, even our naive BLAS fallback should compute the dot-product correctly regardless of whether BLAS was present or not...

deltheil:
> Hmmm, even our naive BLAS fallback should compute the dot-product correctly regardless of whether BLAS was present or not...

In such a case only the F2C auto detection is wrong. On the other hand the Accelerate framework is properly detected by cmake as a BLAS library. So `USE_BLAS` is defined and thus the fallback is *not* called, i.e. `THGeneral.h` looks like this after cmake:

```C
#define USE_BLAS
#define USE_LAPACK
#define BLAS_IS_ACCELERATE
/* #undef BLAS_F2C */
```

This means [`ffloat` is defined as `float`](https://github.com/torch/torch7/blob/8d3154a028ed660d4ebfab9e7a64e0ac31a876f9/lib/TH/generic/THBlas.c#L8) so [`sdot_`](https://github.com/torch/torch7/blob/8d3154a028ed660d4ebfab9e7a64e0ac31a876f9/lib/TH/generic/THBlas.c#L20) has the wrong return type (the way I understand it). Using this signature returns a null dot product.

I observe the same thing by adapting the auto detection code:

```C
/* f2c.c */
#include <stdlib.h>
#include <stdio.h>
float x[4] = { 1, 2, 3, 4 };
float y[4] = { .1, .01, .001, .0001 };
int four = 4;
int one = 1;
extern FFLOAT sdot_();
int main() {
  int i;
  double r = sdot_(&four, x, &one, y, &one);
  printf("r=%f\n", r);
}
```

This gives:

```
$ clang -o f2c -DFFLOAT=float f2c.c -framework Accelerate; ./f2c
r=-0.000000
$ clang -o f2c -DFFLOAT=double f2c.c -framework Accelerate; ./f2c
r=0.123400
```

PR #260, opened 2015-06-15
Fix problem with NaNs in max and min

sergomezcol:
Now min and max always return NaN when the input contains a NaN.

They only returned NaN before when the first element of the input was NaN and the min or max value excluding NaNs otherwise.

### DIFF
--- a/lib/TH/generic/THTensorMath.c
+++ b/lib/TH/generic/THTensorMath.c
@@ -284,19 +284,39 @@ accreal THTensor_(dot)(THTensor *tensor, THTensor *src)
 real THTensor_(minall)(THTensor *tensor)
 {
   real theMin;
+  real value;
+
   THArgCheck(tensor->nDimension > 0, 1, "tensor must have one dimension");
   theMin = THTensor_(data)(tensor)[0];
-  TH_TENSOR_APPLY(real, tensor, if(*tensor_data < theMin) theMin = *tensor_data;);
-  return theMin; 
+  TH_TENSOR_APPLY(real, tensor,
+                  value = *tensor_data;
+                  /* This is not the same as value<theMin in the case of NaNs */
+                  if(!(value >= theMin))
+                  {
+                    theMin = value;
+                    if (isnan(value))
+                      break;
+                  });
+  return theMin;
 }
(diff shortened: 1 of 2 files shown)


sergomezcol:
I've just added a few clarifying comments and some additional test cases.

soumith:
I guess this is the C/C++ math semantics, i.e. for min/max do not propagate the NaN forward. Looks good to me, Anyone have objections for merge?

hughperkins:
Sounds good to me :-)

andresy:
How slow is isnan()?

hughperkins:
Oh, good point, I hadn't checked/noticed that isnan comes from standard-libraries.  According to http://stackoverflow.com/questions/2249110/how-do-i-make-a-portable-isnan-isinf-function , the answer by Johann, it can be implemented portably as `int isnan(double x) { return x != x; }`, as long as the fast math flag is not set.  Not sure if we set that or not?

soumith:
we dont set fast math, but isnan is probably compiled down to a single assembly instruction.

hughperkins:
Per the stackoverflow thread, std::isnan is not available on intel compiler, so we'd need to #ifdef it. If we used the `value ~= value` implementation, it looks like this wouldnt detect nans anywhere where someone has deliberately chosen not to use ieee-compliant maths.   But I'm not terribly sympathetic about that actually.  It will build just fine everywhere, and will be more correct than our current implementation.

sergomezcol:
I used `isnan()` because I saw [lib/TH/THLogAdd.c](https://github.com/torch/torch7/blob/e271689b897780b4e4042641a20f4885931fa185/lib/TH/THLogAdd.c#L26) is using it. But my code also relies on `nan==nan` being false for the `if(!(value >= theMin))` and `if(!(value <= theMax))` conditions. Should I change something?

hughperkins:
I'm ok with merging either way, and in favor of merging in one of these ways.  If we're already using `std::isnan`, then presumably it builds on all the platforms that people are currently using.

soumith:
looks like it si good to go, isnan is not really a performance overhead as far as I see.

PR #268, opened 2015-06-26
Make FFI tensor map more robust.

dominikgrewe:
The 'default' tensor map (in generic/Tensor.c) makes sure that the
input tensors are of the same type. The FFI implementation (for the
contiguous case) doesn't do that. It should do so for consistency and also
to avoid segfaults when passing in a CudaTensor, for example.


dominikgrewe:
The `checkArgument*` functions mimic luaT's error messages. I didn't see anything I could reuse from Lua.

PR #269, opened 2015-06-26
Fix loop logic error in maskedCopy.

dominikgrewe:
Prevents segfault when the src tensor is empty.


soumith:
wait, with this PR, it actually tries to access src tensor, even though it is empty, triggering a segfault.

dominikgrewe:
No, that was the case before. This PR is fixing it.

soumith:
pshh, i got confused between red and green colors. my bad.

dominikgrewe:
Never mind. Happened to me just the other day :)

PR #274, opened 2015-06-30
Documentation tweaks

JonyEpsilon:
I've made small changes to the documentation in places where I got tripped up, as someone completely new to the project, on a first read-through.

Happy to squash/cherry-pick the commits as you wish.


JonyEpsilon:
I would also like to clarify the docs for `scatter` and `gather` as I found them quite confusing, but I'm not yet sure that I understand them well enough to do that. In particular, I think I understand what the do for 2D tensors, but was unable to come up with examples of them working on higher dimensional tensors. Perhaps it would be useful if someone could help me clarify their usage, and then I could try and update the docs from the perspective of a newly educated newbie :-)

soumith:
hi! thanks a lot, all the changes look good. If you squash your commits then it's good to go for merge.

scatter/gather, cc: @dominikgrewe who added them.

dominikgrewe:
Thanks for updating the docs. I struggled a bit when explaining scatter/gather, so if you could help clarify the documentation that would be great. Is there anything in particular that's unclear? I always find the analogy with min/max useful:

```lua
> t = torch.Tensor(5,5,5):uniform()
> v, i = torch.max(t, 2)
> w = torch.gather(t, 2, i)  -- w is equal to v
```
Does that help?

JonyEpsilon:
Thanks @dominikgrewe, that does help explain it. I think I understand what it does now, but it is very hard to put into words clearly! I feel like there's a verb that I'm missing ...
screen shot 2015-07-01 at 21 08 18

JonyEpsilon:
OK, I've had a crack at clarifying the `gather` docs:

https://github.com/JonyEpsilon/torch7/blob/doc-tweaks/doc/tensor.md#tensor-torchgathersrc-dim-index

I'm not sure how you feel about having a diagram. I do think it makes things a bit clearer though :-) Let me know what you think, and I'll squash if you're happy.

BTW, I don't think that the constraint on selecting an element only once is being enforced.

dominikgrewe:
A diagram is a good idea. I found the colouring a bit confusing though, because I always think about it the other way around :) E.g. in the 2D case, I think of it as selecting two elements from each column, i.e. reducing each column to two elements. Whereas your diagram suggests (at least to me) that you're creating two rows, with elements picked from each column. The two are equivalent, but maybe that's what's causing confusion...

> BTW, I don't think that the constraint on selecting an element only once is being enforced.

Yes, that's true. It's more a case of if the constraint isn't met, behaviour is undefined. That could be clarified.

doc/tensor.md
```diff
@@ -1327,36 +1331,51 @@ x:indexFill(2,torch.LongTensor{4,2},-10)
...
 
 <a name="torch.Tensor.gather"/>
-### [Tensor] gather(src, dim, index) ###
+### [Tensor] torch.gather(src, dim, index) ###
 
-Reads values from `src` at the specified indices into the output tensor, selecting a fixed number of values in each row of `src` along the specified dimension.
-The values of `index` must be between 1 and `src->size(dim)` and all values in a row along the specified dimension must be unique.
-All tensors have the same dimensionality and both `index` and the output tensor are of the same size at the specified dimension.
+Creates a new `Tensor` from the `src` tensor by gathering a number of values from each row or column of `src`. 
+The parameter `dim` specifies whether to gather from rows (`dim = 2`) or columns (`dim = 1`) \[or, more generally, the `dim`th dimesion's rows
```
dominikgrewe: I think I'd rather talk about the general case first, and then use 2D as an example. Wdyt?
Is there a general term for "rows" in the case of more than 2 dimensions? We're essentially talking about all vectors along the specified dimension, aren't we?

JonyEpsilon:
> Whereas your diagram suggests (at least to me) that you're creating two rows, with elements picked from each column.

Is it that all of the elements in the first row of index are green, and all in the second are pink - suggesting that somehow the rows "belong" together - that is off? I can see that the colouring gives a "row-centric" view, whereas the operation is not row-oriented. An alternative would be to give each cell its own colour, but then it might look like clown vomit. I can give it a try ...

Regarding the use of a 2D example first: I think there's an advantage in presenting something that can be understood intuitively before presenting the general case. I thought it might be useful to write down what the operation is mathematically as well, perhaps that could come first, then a 2D example, and then comments on N-D. Let me draft it and see what you think.

JonyEpsilon:
OK, I've started by defining the operation mathematically, and I think that helps a lot (hopefully I've also got it correct). I've simplified the picture, as I think it's more clear now with the definition at the top. I've made the picture row-column agnostic.

I think it reads well now.

doc/tensor.md
```diff
@@ -1327,36 +1331,64 @@ x:indexFill(2,torch.LongTensor{4,2},-10)
...
+
+-- dim = 2
+result[i][j][k]... = src[i][index[i][j][k]...][k]...
+
+-- etc.
+```
+
+The same number of values are selected from each row or column, and the same value cannot be selected from a row or column
+more than once. The values in the `index` tensor must be smaller than the length of the row or column, that is between
+1 and `src:size(dim)`. It can be somewhat confusing to ensure that the `index` tensor has the correct shape. Viewed
```
dominikgrewe: In the first paragraph you only talk about rows, but here you say "row or column". Did you mean to do that or did you just forget to change that from your earlier version?
Otherwise it looks good, thanks!

JonyEpsilon:
@dominikgrewe Yes, well spotted. I think everything is consistent now.

@soumith I've squashed the commits, so hopefully this is good to go :-)

doc/tensor.md
```diff
@@ -1327,36 +1331,64 @@ x:indexFill(2,torch.LongTensor{4,2},-10)
...
+result[i][j][k]... = src[index[i][j][k]...][j][k]...
+
+-- dim = 2
+result[i][j][k]... = src[i][index[i][j][k]...][k]...
+
+-- etc.
+```
+
+The same number of values are selected from each row, and the same value cannot be selected from a row more than
+once. The values in the `index` tensor must be smaller than the length of the row, that is between
```
dominikgrewe: "must be smaller" -> "must not be larger" ?

JonyEpsilon: Quite right! Stuck in my zero-based-indexing ways :-)

doc/tensor.md
```diff
@@ -1327,36 +1331,64 @@ x:indexFill(2,torch.LongTensor{4,2},-10)
 ```
 
 <a name="torch.Tensor.gather"/>
-### [Tensor] gather(src, dim, index) ###
+### [Tensor] torch.gather(src, dim, index) ###
```
dominikgrewe: Sorry, just one more thing :)
Can you remove the "torch."? It's inconsistent with the rest of the docs.

JonyEpsilon: No worries, keep 'em coming :-)

I put the `torch.` in there to try and make it _more_ consistent, but it must have backfired! My reasoning was that the other functions are mostly documented for the calling pattern `src:func(param1, param2)` whereas this one is documented with the `src` tensor as an argument which is more like the `torch.func(src, param1, param2)` calling pattern (or the `src:func(src, param1, param2)` pattern, which doesn't make much sense for this function). My first instinct was to remove the `src` parameter from the heading, but then there's no name to refer to the `src` tensor :-(

The reason I changed it is because I found the existing docs confusing - before I understood the function I thought `src` was a *third* tensor involved somehow (like `x:gather(src, 1, i)). I defer to your judgment as to the best way to present it. Let me know what you think and I'll make the change.

dominikgrewe: Yes, that's a bit confusing. You can actually do all of the following:
```lua
torch.gather(src, dim, idx)  -- Writes the result to a new tensor.
src:gather(dim, idx)         -- Same as above.
dst:gather(src, dim, idx)    -- Writes the result to dst.
```
This isn't the case for some other functions, but I'm not sure whether the `torch.` helps here.

JonyEpsilon: I've changed it to `[Tensor] gather(dim, index)` which makes it consistent with the rest of the docs, and re-worded slightly to introduce `src` when it is needed.

doc/tensor.md
```diff
@@ -1327,36 +1331,66 @@ x:indexFill(2,torch.LongTensor{4,2},-10)
 ```
 
 <a name="torch.Tensor.gather"/>
-### [Tensor] gather(src, dim, index) ###
+### [Tensor] gather(dim, index) ###
```
fmassa: There is still a problem in here, `src` is missing.

JonyEpsilon: I think `src` *should* be missing, to be consistent with the rest of the docs which present the functions, on the whole, as method calls. (As it is the docs aren't completely consistent in their style, but not so much as to be confusing. But I did find the particular way that `gather`'s signature was presented was confusing hence the suggestion to change it.)

fmassa: True. My bad !

JonyEpsilon: :-)

dominikgrewe: I'm still not convinced that this is better, because you can use it both as a method on the input tensor `src` or on the output tensor (in which case `src` is a parameter). But if you think this is clearer than I'm happy to leave it. I just want to make sure that you see both use cases.

JonyEpsilon: If I understand correctly, it's true for many of the functions that you can call them either as `src:func()` or `torch.func(src)`. Is the point here that this function can additionally be called as `dst:func(src)` which doesn't work for all functions? If that's the case then it's perhaps best to present the `src:func()` case in the heading, which would be consistent with the other functions, and familiar to the reader, and then make an explicit note in the description that it can be called the `dst:func(src)` way? Perhaps the way it is done at the bottom of the description of `index` "It is possible to store the result into an existing Tensor with `result:index(source, ...)`"

How does that sound?

dominikgrewe: It's actually quite common to be able to call a function both as `src:func()` and as `dst:func(src)`. However, often the first case means essentially `src:func(src)`, i.e. it's an in-place operation. This is where `gather` differs.

I'd say the most common uses of `gather` will be `dst:gather(src, dim, idx)`, because you want to avoid memory allocations where possible. So maybe that's an argument for leaving `src` in the header :)

dominikgrewe:
I just noticed you took out the comment about indices having to be unique. You're right about `gather`, where it doesn't matter, but `scatter` is different, because you're writing to these locations and we want to ensure that no location is written two more than once (will be useful for CUDA implementation). Or do you think that'd be too restrictive?

doc/tensor.md
timharley: > essentially src:func(src), i.e. it's an in-place operation

Sorry to nitpick, but I think it is much more common for `src:func(args)`
to return a freshly allocated tensor that holds the result, rather than a
destination for the result being explicitly provided like in:
`dst:func(src, args)`.

JonyEpsilon:
@dominikgrewe I think there's a comment to that effect:

"the same value cannot be selected from a row more than once"

but maybe that is too oblique. Generally explicit is good, so maybe it's better to say exactly what you said above in the scatter `docs`.

Regarding the form of the call to use (and more generally!) I think I might be out of my depth here - the changes in the PR are from my first read through of the docs from the perspective of a new user. I think the best thing is perhaps to comment on why I changed it (i.e. what confused me) and invite someone more knowledgeable to write something that might make it easier for newbies, but without losing the subtleties as my changes might. So:

What confused me was that it wasn't obvious whether `src` was a third tensor as in `x:gather(src, dim, index) or not. Now I see that the answer is maybe :-) I note that at the top of (I think) the Math docs there is a note about calling functions as either `torch.func(t)` or `t:func`, but I hadn't read that bit yet, so when I read the tensor docs I wasn't really fully aware of that. From the perspective of a new user, I'd say it's useful to be as explicit as possible, at least at the start.

Can I ask someone more experienced to make the call on the best way to present the gather function, ideally that clarifies it for newbies - maybe just list all three alternatives? Overall it might be better to add a note somewhere near the top on calling conventions, but that's probably for another PR. It would be good to keep this one clearly scoped if possible.

JonyEpsilon:
OK, I've had a stab at incorporating the comments so far, although I still must confess to being a bit out of my depth.

- I've added a paragraph towards the top of the docs highlighting explicitly that the functions can be called in more than one way.
- I've stuck with `gather(dim, index)` for the heading, which is consistent with the presentation of most of the other functions, but added a note that `result:gather(src, dim, index)` is possible.
- Reinstated the explicit note about index uniqueness in `scatter`.

How does that look?

hughperkins:
Hi JonyEpsilon, I can give you an example of a 'real-world' usage for gather/scatter.  Not sure if that helps?  eg, in a neural network, the last layer in a classification network will typically have the following inputs:

inputs to ClassNLLCriterion:
- tensor 'input', containing the value of each output neuron, maybe there are 4 of them let's say
- 'ground truth' label for this sample, named 'target'.  It will be a number, from 1 to number of classes.  In this case there are 10 classes, so its a number from 1 to 4, eg 3

Here are some example inputs:
- input: {0.1, 0.2, 0.6, 0.1}
- target: 3

The output from this ClassNLLCriterion:updateOutput class should then be very simply the value of the 'input' tensor at the position 'target'.  In this case, it will be input[target], ie input[3], ie '0.6'

So, straight away, you can start to see this is kind of gather-ish, but we only have one dimension.

In reality, the input tensor is not 1-dimensional, certianly not when we're using the GPU, but is a mini-batch, with eg 128 training examples.  For each training example we have here 4 values, so it's here a 128x4 tensor, ie a torch.Tensor(128,4).  Similarly, target is no longer a single value, but a tensor of 128 values, ie, for simplicity, it's a torch.Tensor(128,1)

And now, very simply our ClassNLLCriterion:updateOutput method simply has to call gather :-)

```
self.output:gather( input, 2, target )
```

Now, in reality, we might normalize it a bit, and target might not be a Tensor(128,1), but a Tensor(128), so we might need an unfold, and a div.  But the important operation is the gather, and here it is:

https://github.com/hughperkins/clnn/blob/master/ClassNLLCriterion.lua#L30
```
function nn.ClassNLLCriterion:updateOutput(input, target)
  self.buffer:gather(input, 2, target:unfold(1,1,1))
  self.output = - self.buffer:sum() / num_samples
  return self.output
end
```

Similarly, scatter is just the reverse.  During backpropagation, we have to send the loss gradient from upstream down to where the inputs came from downstream.  That's a scatter:

https://github.com/hughperkins/clnn/blob/master/ClassNLLCriterion.lua#L47
```
function nn.ClassNLLCriterion:updateGradInput(input, target)
  z = z / target:size(1)
  self.gradInput:scatter(2, target:unfold(1,1,1), z)
  return self.gradInput
end
```

hughperkins:
@dominikgrewe Presumably, for scatter fill, it doesnt matter if two threads write the same value to the same location?  Only matters in the case of the three tensor scatter version, with a `src` tensor?

dominikgrewe:
@JonyEpsilon I think it looks good now. Thanks for the clarification on how to call the functions.
@soumith Are you happy with the new "A note on function calls" section?

dominikgrewe:
@hughperkins You're right about that. It doesn't matter for scatter fill. I'm inclined to leave it as-is though. Changing the behaviour depending on whether a tensor or scalar is passed might be confusing. And I don't think it's too restrictive, do you?

JonyEpsilon:
@hughperkins Thanks :-) It is useful to see a real world use, as while I think I understood mechanically what it does, I wasn't sure how it would be used!

I've squashed the commits into one.

soumith:
Thanks Jony!

PR #277, opened 2015-07-01
fixing torch.random(a,b) for a,b >= 2^31

clementfarabet:
Current implementation casts random number to int :-(. @andresy was this intended?


soumith:
ouch! bad.

clementfarabet:
yep :-(

Also if (b-a) > 2^32, it really only ever returns things in [a,a+2^32-1]. Should we assert instead of silently doing the wrong thing?

PR #280, opened 2015-07-01
THAlloc/THRealloc attempt lua GC on failure

adamlerer:
The Lua garbage collector isn't aware of torch-allocated memory, so it doesn't know when to run the GC and the user ends up having to call collectgarbage() manually from their script, or running out of memory.

This PR lets THAlloc and THRealloc attempt lua GC if malloc/realloc fails.

This idea is thanks to @colesbury, I just had some free time to implement it.

To demonstrate, you can run the following

```lua
require 'torch'
for i=1,1000 do
  print(i)
  a = torch.FloatTensor(1000000000):zero()
end
```

This loop will run out of memory prior to this PR, but now the GC will be run.

I'll add this to cutorch as will if this gets acepted.

@soumith @andresy

### DIFF
--- a/lib/TH/THGeneral.c
+++ b/lib/TH/THGeneral.c
@@ -13,6 +13,9 @@ static void defaultTorchErrorHandlerFunction(const char *msg, void *data)
 static __thread void (*torchErrorHandlerFunction)(const char *msg, void *data) = defaultTorchErrorHandlerFunction;
 static __thread void *torchErrorHandlerData;
 
+static __thread void (*torchGCFunction)(void *data) = NULL;
+static __thread void *torchGCData;
+
 void _THError(const char *file, const int line, const char *fmt, ...)
 {
   char msg[2048];
@@ -91,15 +94,15 @@ void THSetArgErrorHandler( void (*torchArgErrorHandlerFunction_)(int argNumber, 
   torchArgErrorHandlerData = data;
 }
 
-void* THAlloc(long size)
+void THSetGCHandler( void (*torchGCFunction_)(void *data), void *data )
+{
+  torchGCFunction = torchGCFunction_;
+  torchGCData = data;
+}
+
+void* THAllocInternal(long size)
 {
   void *ptr;
-
-  if(size < 0)
-    THError("$ Torch: invalid memory size -- maybe an overflow?");
-
-  if(size == 0)
-    return NULL;
 
   if (size > 5120)
   {
@@ -115,6 +118,25 @@ void* THAlloc(long size)
   else
   {
     ptr = malloc(size);
+  }
+  return ptr;
+}
+
+void* THAlloc(long size)
+{
+  void *ptr;
+
+  if(size < 0)
+    THError("$ Torch: invalid memory size -- maybe an overflow?");
+
+  if(size == 0)
+    return NULL;
+
+  ptr = THAllocInternal(size);
+
+  if(!ptr && torchGCFunction) {
+    torchGCFunction(torchGCData);
+    ptr = THAllocInternal(size);
   }
 
   if(!ptr)
@@ -137,10 +159,17 @@ void* THRealloc(void *ptr, long size)
   if(size < 0)
     THError("$ Torch: invalid memory size -- maybe an overflow?");
 
-  ptr = realloc(ptr, size);
-  if(!ptr)
+  void *newptr = realloc(ptr, size);
+
+  if(!newptr && torchGCFunction) {
+    torchGCFunction(torchGCData);
+    newptr = realloc(ptr, size);
+  }
+
+  if(!newptr)
     THError("$ Torch: not enough memory: you tried to reallocate %dGB. Buy new RAM!", size/1073741824);
-  return ptr;
+
+  return newptr;
 }
 
 void THFree(void *ptr)
--- a/lib/TH/THGeneral.c
+++ b/lib/TH/THGeneral.c
@@ -3,6 +3,14 @@
 #ifndef TH_HAVE_THREAD
 #define __thread
 #endif
+
+#if defined(TH_DISABLE_HEAP_TRACKING)
+#elif (defined(__unix) || defined(_WIN32))
+#include <malloc.h>
+#elif defined(__APPLE__)
+#include <malloc/malloc.h>
+#endif
+
 /* Torch Error Handling */
 static void defaultTorchErrorHandlerFunction(const char *msg, void *data)
 {
@@ -12,9 +20,6 @@ static void defaultTorchErrorHandlerFunction(const char *msg, void *data)
 
 static __thread void (*torchErrorHandlerFunction)(const char *msg, void *data) = defaultTorchErrorHandlerFunction;
 static __thread void *torchErrorHandlerData;
-
-static __thread void (*torchGCFunction)(void *data) = NULL;
-static __thread void *torchGCData;
 
 void _THError(const char *file, const int line, const char *fmt, ...)
 {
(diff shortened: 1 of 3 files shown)


lib/TH/THGeneral.c
```diff
@@ -137,10 +159,17 @@ void* THRealloc(void *ptr, long size)
...
-  if(!ptr)
+  void *newptr = realloc(ptr, size);
+
+  if(!newptr && torchGCFunction) {
+    torchGCFunction(torchGCData);
+    newptr = realloc(ptr, size);
+  }
+
+  if(!newptr)
     THError("$ Torch: not enough memory: you tried to reallocate %dGB. Buy new RAM!", size/1073741824);
```
tkoeppe: Doesn't this leak `ptr`?

andresy:
Thanks Adam, that is interesting!

Question: did you benchmark the overhead of calling lua_gc every time one alloc something? In particular, if  one had some C loop which was allocating/deallocating data, then now it will also call a gc step each time... so one might start to see some gc overhead in C code where one should not expect this to happen.

Remark: TH is agnostic of Lua. Maybe it would be better to allow customization of memory allocation functions, rather than having a customizable GC function which does not mean much in C code.

lib/TH/THGeneral.c
adamlerer: Well, at least in the standard (Lua) implementation of THError, it calls lua_error which is a longjmp so this function will never return.

I suppose if this function did return after calling THError, it would leak.

adamlerer:
@andresy 

No I did not benchmark because in this PR you're only running the GC when malloc returns NULL (i.e. you totally run out of memory). So it's only going to run very rarely. However, I have an upcoming PR that will run the GC more regularly, because (a) you don't want to use up the whole RAM before running GC, and (b) on some machines malloc will never return NULL, you'll just start thrashing virtual memory or get killed by the kernel. In that version I have to be more careful not to run GC more often. I'll send that tomorrow.

Regarding your remark, I did keep lua_State out of TH, but you're right that I'm adding an (optional) GC function. I can move this out of TH if you prefer and move the whole allocator into init.c or something.

lib/TH/THGeneral.c
dominikgrewe: Calling `lua_error` doesn't mean your process will exit (e.g. if this code was executed within a `pcall`). So you have to consider freeing memory before calling THError to avoid leaks.

tkoeppe: Ah, does this function iftself have the same semantics as `realloc`? That is, if it returns null, then the original allocation has not been touched? That means that the caller has to deal with that and free their original memory.

mraleph:
@adamlerer have you considered doing `lua_gc(L, LUA_GCSTEP, ...)` every so often when you allocate a tensor (with a size of the step determined based on the size of the tensor, total size of tensors allocated and the total amount of Lua memory)? 

Seems like ensuring the constant progress of the incremental collector is the best way to keep the memory consumption at bay and avoid OOM situation.

lib/TH/THGeneral.c
adamlerer: @dominikgrewe I didn't say the process would exit, just that the function would never return. For example, if the caller writes `self->data = THRealloc(self->data, size)`, and you hit the error, `self->data` will not be modified and will be freed when `self` is cleaned up.
Notice that I haven't changed the behavior of THRealloc w.r.t. OOM - it always just called THError.

@tkoeppe no, `THRealloc` does not have the same semantics as `realloc`. The semantics are "if `THRealloc` ever returns, then it returns a pointer to a reallocated block of the requested size".

tkoeppe: Interesting. So what I'd like to understand is this. Is the following code correct?

    void * p = THAlloc(n);
    void * q = THRealloc(p, 2 * n);
    THFree(q);

Or should I have to write:

    void * p = THAlloc(n);
    void * q = THRealloc(p, 2 * n);
    THFree(q ? q : p);

Or is neither code correct because `THRealloc` never returns and I just can't free `p`?

adamlerer:
@mraleph yes, see my earlier comment. Calling the lua GC only on malloc failure doesn't always work well, so I am working on keeping track of TH memory allocations and using a basic scheme for periodically running the Lua GC.

To clarify though, the reason I have to trigger GC from TH is because the TH memory is not managed by Lua, which is why Lua is not triggering GC on its own. If I allocate some huge TH objects, Lua still only sees a few bytes of Lua memory allocated, so it will not necessarily run GC, and it will not necessarily clean up these objects on an incremental GC. Therefore I think it's better to periodically run full GC. AFAIK torch is used for batch jobs that are insensitive to GC pauses; if that's not the case, I can add an off switch :).

lib/TH/THGeneral.c
adamlerer: tl;dr Both versions will leak if you catch the OOM within Lua, but lots of TH code isn't safe for this scenario.

@tkoeppe it depends on the context. Both `THAlloc` and `THRealloc` can call `THError` which in Lua calls `lua_error` and returns control to the Lua interpreter, which can catch the error in a `pcall`. So, if you want to avoid leaking memory, you need to make sure you remain in a consistent state if alloc never returns.

Note that this is not unique to `THRealloc`. Here's another example:
```C
void *p = THAlloc(n);
void *q = THAlloc(n); // OOM leaks p!
free(p);
free(q);
```

Correct code (for Lua) must assign all allocations to Lua-managed objects that will be cleaned up properly by the Lua GC. Here's a correct example from the code:

```C
tensor->size = THRealloc(self->size, sizeof(long)*nDimension);
```
`tensor` is a `THTensor*` sitting on the Lua stack, so `tensor->size` will be freed when `tensor` gets GC'd. If `THRealloc` never returns that's fine, the `tensor->size` still has the old pointer and will free it when it's GC'd.

Don't take this to mean that it's safe to catch OOM errors in a `pcall`. TH does *not* leave objects in a consistent state upon OOM. For example, consider generic/THTensor.c:667:

```C
self->size = THRealloc(self->size, sizeof(long)*nDimension);
self->stride = THRealloc(self->stride, sizeof(long)*nDimension);
self->nDimension = nDimension;
```
If `THRealloc` OOMs in one of the intermediate calls, `self` is broken. Calling `size()` will probably segfault. So this whole discussion is quite moot IMO.

tkoeppe: OK, that explanation makes sense. Thanks!

adamlerer:
@andresy 

I added another commit to this PR that triggers GC based on a dynamic heap size soft max, instead of waiting for `malloc` to fail. I've tested the example code above on Linux and OSX and it keeps memory usage down and prevents OOM.

I profiled the overhead of this code for repeated allocation/deallocation of small (size 1) and large (size 1e9) tensors. In both cases, the GC hooks add an overhead of 8-9% to allocation time. I can elaborate more if you're interested.

I thought about your earlier remark about being lua-agnostic, and I think it's better to leave the GC hooks in THGeneral.c. I'm simply adding an optional hook in TH where GC'd front-ends can get their garbage collectors triggered. If you're not using a GC'd front-end, you just don't register that hook. I added comments in THGeneral.c to clarify that.

tkoeppe:
(Maybe rebase, squash and force-push?)

adamlerer:
@tkoeppe I'll squash once it's accepted; I just want to keep a record of GC-on-failure vs. GC-on-heap-size.

tkoeppe:
Of course!

andresy:
@adamlerer Thanks Adam, I think it is a good idea to control the GC with the heap size. Given it is used everywhere, yes, I would be happy if you have ideas to elaborate on this :smiley: 

Considering the GC hook: in the end you added this hook it to hack THMalloc/THFree. I thought it would be cleaner to allow more general hooks for allocations methods (if anyone want to change it for some other allocators) rather than adding this extra specific hook. E.g., if Lua GC had a way to specify the actual size of userdatas (which might be the case in other scripting languages), I am not sure one would need this hook or would schedule calls to this hook in the same way.

So in the end: do you think this is general enough?

adamlerer:
@andresy what do you mean by "elaborate on this"?

I think that the GC-triggering hooks I added could be useful for things other than Lua. For example, Python's reference-cycle GC is based on # of allocations, so it might need to be triggered to clean up TH's large objects (http://www.digi.com/wiki/developer/index.php/Python_Garbage_Collection).

I agree with you that for some applications you might want to override all of THAlloc/THFree, for example if you wanted to write your own special allocator for tensors, handle alignment differently, change the scheduler, etc. Personally, it seems like premature generalization since we don't have an actual use case. But I'm happy to write it that way, it's easy enough.

Let me know what you prefer!

dominikgrewe:
> I agree with you that for some applications you might want to override all of THAlloc/THFree, for example if you wanted to write your own special allocator for tensors, handle alignment differently, change the scheduler, etc. Personally, it seems like premature generalization since we don't have an actual use case. 

In cutorch we actually have a use case where we pass in a different allocator when creating a storage. This is to create pinned memory using `cudaMallocHost`. We don't completely swap out THAlloc though, but only overwrite it in certain situations.

adamlerer:
@dominikgrewe interesting. So does this argue for or against adding allocator hooks? This sort of illustrates my concern that we could build "general" allocator hooks, but they probably wouldn't exactly fit the next person's requirements (e.g. here you can't swap out `THAlloc` wholesale, because you only want to use `cudaMallocHost` in certain situations). Maybe better to just let the next person write the hooks they need when they need them.

The API I'm assuming for swapping out the allocator is

```lua
void THSetAllocator(
  void* (*allocFunction)(long size),
  void* (*reallocFunction)(void *ptr, long size),
  void* (*freeFunction)(void *ptr),
  void *allocatorData);
```

adamlerer:
I went ahead and implemented the "generic alloc hooks" approach in the branch "THAlloc_generic". You can look at it here:

https://github.com/torch/torch7/compare/master...adamlerer:THAlloc_generic?expand=1

Indeed, there seems to be minimal performance difference between the two versions for both small and large allocations:

```
Version            |  Alloc time (size 1)  |  Alloc time (size 1e9)
===================================================================
master             |   1.1e-6              |  1.5
THAlloc2           |   1.3e-6              |  1.6
THAlloc_generic    |   1.3e-6              |  1.6
```

andresy:
Thanks @adamlerer! I like both contributions. The situation is:
- The THAlloc_generic branch makes the allocator flexible. However, if one changes the allocator for another one (say jemalloc or possibly what @dominikgrewe is doing with cutorch), then one has to re-implement the GC part
- The THalloc2 branch allows you to change easily the GC hook, but not the allocator.

One could merge the two: THAlloc_generic, with an customizable GC hook in addition to the customizable malloc/realloc/free hooks.

What is your opinion @dominikgrewe?

adamlerer:
I looked at the code with @andresy and realized there's already allocator hooks in THAllocator; I think I'd just be adding needless complexity by adding a second layer of allocator indirection. So I'll just stick with the THAlloc2 branch

I also realized that someone using a custom allocator will need a way to hook into the new GC infrastructure, so I exposed the `THHeapIncr` and `THHeapDecr` functions. See the latest version of the PR.

andresy:
Looks good to me. Thanks! @dominikgrewe?

dominikgrewe:
`THHeapIncr/Decr` rely on functions like `malloc_usable_size`, which assume that the given pointer has been returned by `malloc` (or a related function). So they won't necessarily work with a custom allocator, will they? Maybe they should also take a size argument?

adamlerer:
Thanks @dominikgrewe you're absolutely right. In the spirit of providing the minimal necessary API for custom allocators, I've changed the API to just `THHeapUpdate(long size)`. (The other obvious API function would be `THCollectGarbage()` that just runs the GC hook, but a custom allocator can register the GC itself if it so desires).

dominikgrewe:
Thanks Adam. This looks good to me now!

hughperkins:
Nice!  So many questions on the torch forum are because of a missing `collectgarbage()`.  I guess for every 1 person who posts a question about this on the forum, there are 30 people who just run away, try something different :-P

soumith:
ok, crossing my fingers and merging this.

Issue #281, opened 2015-07-02
Severe side effect in lua_tostring (lua issue?)


clementfarabet:
Hey guys,

we recently noticed weird edge effects when using large numbers in functions like torch.random, torch.uniform, ... after quite a bit of debugging, we discovered that `lua_tostring` has a side effect when the input is a (large) number. This is pretty scary, as `lua_tostring` is used automatically by all the API calls generated by cwrap (so all functions in TensorMath).

Here's how to reproduce it:

```C
#include <lua.h>
#include <luaT.h>
#include <stdio.h>

static int test(lua_State *L) {
  long nb1 = (long)lua_tonumber(L, 1);
  long nb2 = (long)lua_tonumber(L, 2);
  printf("a,b = 0x%lx,0x%lx\n",nb1,nb2);

  lua_tostring(L, 2);

  nb1 = lua_tonumber(L, 1);
  nb2 = lua_tonumber(L, 2);
  printf("a,b = 0x%lx,0x%lx\n",nb1,nb2);

  return 0;
}

static const struct luaL_reg routines[] = {
   {"test", test},
   {NULL, NULL}
};

extern int luaopen_libtest(lua_State *L)
{
  luaL_openlib(L, "libtest", routines, 0);
  return 1;
}
```

Save this to `test.c` and run this (with the right paths to your lua):

```sh
#!/usr/bin/env bash

gcc -o libtest.dylib -shared test.c -I ~/local/include/ -undefined dynamic_lookup
~/local/bin/th -e "\

   local lib = require './libtest'

   print('')
   print('ok:')
   lib.test(128,256)

   print('')
   print('ok:')
   lib.test(2^32, 2^34)

   print('')
   print('wrong:')
   lib.test(2^49, 2^49)

   print('')
   print('wrong:')
   lib.test(2^52, 2^52)
"
```

With Lua 5.1 it gives this:

```
ok:
a,b = 0x80,0x100
a,b = 0x80,0x100

ok:
a,b = 0x100000000,0x400000000
a,b = 0x100000000,0x400000000

wrong:
a,b = 0x2000000000000,0x2000000000000
a,b = 0x2000000000000,0x1fffffffffffe

wrong:
a,b = 0x10000000000000,0x10000000000000
a,b = 0x10000000000000,0x10000000000004
```

Can someone confirm that he's also seeing this problem? 

Does someone know of a limit on Lua numbers, where beyond ~ 2^46 things just break?

/cc @andresy @koraykv @soumith

[...]

PR #282, opened 2015-07-02
Change lua_isstring check in torch_random

nkoumchatzky:
lua_isstring returns true for a lua_Number. Then lua_tostring was called which led to a re-formatting (#281) that generated a loss of precision for doubles who need more than 14 digits of precision.
Also avoids two useless conversions.  
Replaced by the exclusive (lua_type(L, narg) == LUA_TSTRING) check.

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -58,7 +58,7 @@ static int torch_NAME(lua_State *L)
   else if(narg >= 2 && (tname = torch_istensortype(L, luaT_typename(L, 2)))) /* second? */
   {
   }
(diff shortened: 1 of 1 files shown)


andresy:
Yes, it is good, thanks. I always remember lua_tostring has a side effect (note that if you have a number on the stack, it returns a string for this number but also converts it to a string), but I did not pay attention for lua_isstring(). Glups.

Shouldn't it be rebased though? (I see a merge)

clementfarabet:
Cool. Yes it needs a rebase.

Thanks!

PR #283, opened 2015-07-02
** Fix and improve torch.[].random(a[,b]) **

nkoumchatzky:
- Former version of torch.[].random(a[,b]) used modulo, which is not exact,
especially when (b-a) was comparable to 2^32-1.
e.g. torch.random(1,2/3 x (2^32-1)) will sample approximately twice more
in the range [1,1/3x(2^32-1)] than in the range [1/3x(2^32-1),2/3x(2^32-1)].

- Only solution is the multiplication.

- Also extends the max range diff to 2^53 from 2^32 using nested sampling
from Mersenne-Twister 32 bits.

- Also adds a test that performs a Pearson's chi-squared test on
the distribution in various cases.

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -546,32 +546,72 @@ for _,Tensor in ipairs({"ByteTensor", "CharTensor",
    if Tensor == 'ByteTensor' then -- we declare this only once
       interface:print(
          [[
+/*
+Exact random integer generation for (b-a) <= 2^53 with nested sampling.
+0.00048828101 = (2^21-1)/(2^32-1) -- Normalize first random val to [0,2^21 - 1]
+2097152 = 2^53 / 2^32 -- Normalize second random val to [0,2^53 - 2^21]
+9007199254740992 = 2^53 -- Significand precision
+*/
 static long THRandom_random2__(THGenerator *gen, long a, long b)
 {
   THArgCheck(b >= a, 2, "upper bound must be larger than lower bound");
-  return((THRandom_random(gen) % (b+1-a)) + a);
+  return (long)(((double)(THRandom_random(gen)) * 0.00048828101
+         + (double)(THRandom_random(gen)) * 2097152)
+         / 9007199254740992.
+         * (double)(b+1UL-a))
+         + a;
 }
-         
+
+/*
+Exact random integer generation for b <= 2^53 with nested sampling.
+0.00048828101 = (2^21-1)/(2^32-1) -- Normalize first random val to [0,2^21 - 1]
+2097152 = 2^53 / 2^32 -- Normalize first random val to [0,2^53 - 2^21]
+9007199254740992 = 2^53 -- Significand precision
+*/
 static long THRandom_random1__(THGenerator *gen, long b)
 {
   THArgCheck(b > 0, 1, "upper bound must be strictly positive");
-  return(THRandom_random(gen) % b + 1);
+  return (long)(((double)(THRandom_random(gen)) * 0.00048828101
+         + (double)(THRandom_random(gen)) * 2097152.)
+         / 9007199254740992.
+         * (double)(b) + 1.);
 }
          ]])
    end
 
    interface:print(string.gsub(
                       [[
+/*
+Exact tensor random integer generation for (b-a) <= 2^53 with nested sampling.
+0.00048828101 = (2^21-1)/(2^32-1) -- Normalize first random val to [0,2^21 - 1]
+2097152 = 2^53 / 2^32 -- Normalize first random val to [0,2^53 - 2^21]
+9007199254740992 = 2^53 -- Significand precision
+*/
 static void THTensor_random2__(THTensor *self, THGenerator *gen, long a, long b)
 {
   THArgCheck(b >= a, 2, "upper bound must be larger than lower bound");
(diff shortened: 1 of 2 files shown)


timharley:
Could you provide a benchmark for these changes?

nkoumchatzky:
Hey Tim,

What type of benchmark are you expecting?

timharley:
By a benchmark, I mean some speed tests. How much slower is this version?
I'd expect a factor of ~2 slowdown due to the two samples from the mersenne
twister? Getting a distribution which is actually *ahem* uniform in more
cases will be very welcome (so thanks!), but for some use cases a factor of
2 slowdown might be unwelcome.

nkoumchatzky:
@timharley Ok I see. I don't see many use cases though where torch.random(a,b) could be a bottleneck. If this happens however, it is always possible to generate one MT32 torch.random() sample directly and perform the multiplication (or the modulo version if you want to keep the buggy modulo for backward-compatibility reasons :) ). 
Other solution is to implement MT64 to get int64 random integers. I will test perf though and get back to you.

jucor:
Hi @nkoumchatzky , @timharley 

I've been advising on torch RNG behind-the-scene for a little while, happy to jump in :)

TLDR: Torch maintainers aware of this since last year, decided not to fix,  [torch-randomkit](https://github.com/deepmind/torch-randomkit) contains state-of-the-art generators (for uniform, rejection sampler).

Long version:
Thanks for your push on this, Nicolas! This bug in `torch.uniform` was flagged by a colleague late 2014, and was the subject of off-github discussions back then, and again now with your PR, with @koraykv and @dominikgrewe back then

While I am personally quite intense on good-quality generators,  and it is clear to all that the current implementation is *not* uniform, the tradeoff discussed by the maintainers (back in December) for keeping it in torch and relying on  [torch-randomkit](https://github.com/deepmind/torch-randomkit) for precise generators is to:
* avoid breaking reproducibility for the many projects who already use it
* as long as the current generator is  "good enough" and simple enough.

From there on, two possibilities:
* either there is enough people caring about quality of the default RNG to overcome the reluctance to change, in which case we can benchmark your sampler vs the rejection  sampler from randomkit vs the current,  and import the fastest (I'd be very happy to discuss the merits of the two, actually, I found your implementation very interesting).
* or we all decide that we keep with the existing policy, and we make a bigger push to advertise `torch-randomkit`.

I'm fine with both: while I love any advertisement for the oft-overlooked torch-randomkit [credit goes to the Scipy developer, who kindly agreed to us borrowing it from their C library], I do not claim to have the bigger picture that @koraykv has. In isolation, all things being equal, I'd rather go for a good default generator -- after all, Torch's very dynamic development is no stranger to breaking backward compatibility, or in this case, reproducibility [as the calls would still work].

What do you think?

jucor:
Btw, in terms of speed:
* @nkoumchatzky 's sampler has exactly two calls, all the time.
* rejection sampling has *on expectation* lambda =  (b-a)/2^math.ceil(math.log(b-a, 2))  calls, i.e. the ratio between the interval length and the closest higher power of 2. The full distribution is a [geometric distribution](https://en.wikipedia.org/wiki/Geometric_distribution) of parameter lambda. So while it makes the number of calls non-deterministic, the average number of calls will always be <= 2.

nkoumchatzky:
Hi @jucor!

I am definitely in favor of trading reproducibility for quality, at least when the original implementation is buggy. The current generator was probably good enough until now, but we're hitting the limit right now on some problems, and it might become a bad surprise for other people/groups in similar situations.

torch-randomkit looks great, the only problem is the transition period that would be needed to integrate it. The other thing is we don't need fancy distributions (which makes torch-randomkit very valuable btw, on top of its highest quality) in our deep learning stack, at least for now. In the end, the best solution might be to integrate torch-randomkit directly in torch7, but in this case we go back to modifying the current torch7 RNG stack :)

And thanks for the rejection sampling analysis, didn't cross my mind :)

jucor:
Great, thanks for the feedback. I m at ICML this week (are you? Meeting?),
where I will catch up with @koraykv and update this thread afterward.
Cheers.

nkoumchatzky:
Indeed I am! Let's catch up tomorrow then. My email is
nicolas.koumchatzky@gmail.com
Cheers

nkoumchatzky:
@clementfarabet @jucor @koraykv Looks like a good solution would be to keep both versions, this new one becoming the default. How do you guys want to play that?

Nico

soumith:
@clementfarabet @andresy @koraykv status on this? Merge / no merge / changes requested?

jucor:
@soumith @koraykv Ages after ICML, a quick report [as this issue popped up in another offline discussion]: as discussed with @nkoumchatzky and @koraykv during ICML in Lille, the best solution was  to keep both version indeed, making the new (unbiased) one as the default but allowing the old one to still be there and letting the user switch to one or the other. Why it never happened: I don't think we ever decided who would be doing it :) So that would be "changes request" to support this behaviour, I believe.

jucor:
And in the meantime, just using [randomkit.randint](http://deepmind.github.io/torch-randomkit/randomkit.html#randint) works, too [this one uses accept/reject to be unbiased].

soumith:
please re-open if you want this to be merged.

PR #285, opened 2015-07-08
MSVC corrections

diz-vara:
Some corrections that allow for compilation in MSVC:
- do not set -Werror compilation flag that MSVC does not understand
- remove TH_API from function definitions for static library compilation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,10 @@ SET(CMAKE_MODULE_PATH
   "${CMAKE_CURRENT_SOURCE_DIR}/cmake"
   "${CMAKE_MODULE_PATH}")
 
-SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror=implicit-function-declaration -Werror=format")
+IF (NOT MSVC)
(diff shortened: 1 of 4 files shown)


soumith:
rather than removing TH_API (which helps for dynamic builds), you would want to disable it for static builds with a macro, over here:
https://github.com/torch/torch7/blob/master/lib/TH/THGeneral.h.in#L25

soumith:
if you make those changes, please do reopen the Pull Request. thanks.

diz-vara:
I've removed TH_API only from definitions (in TH*.c files) - but they are present in declarations.
It just follows the pattern in  luaT.h - luaT.c 
The macro still works - but it can't puts 'dllimport' to static library files - and MSVC complains...

soumith:
oh, i see, I did not notice that. that makes sense.

hughperkins:
Note that TH_API is also needed on Windows for building dlls, by defining `__declspec(export)` or `__declspec(import)`.

soumith:
can you squash your commits, it looks like I can merge them.

soumith:
thanks for the PR!

PR #290, opened 2015-07-10
Element-wise min and max operations.

dominikgrewe:
Add cmin & cmax operation for element-wise min and max between two
tensors and between a tensor and a scalar.

### DIFF
--- a/lib/TH/generic/THTensorMath.c
+++ b/lib/TH/generic/THTensorMath.c
@@ -1066,6 +1066,30 @@ void THTensor_(cross)(THTensor *r_, THTensor *a, THTensor *b, int dimension)
                        r__data[2*r__stride] = a_data[0*a_stride]*b_data[1*b_stride] - a_data[1*a_stride]*b_data[0*b_stride];);
 }
 
+void THTensor_(cmax)(THTensor *r, THTensor *t, THTensor *src) {
+  THTensor_(resizeAs)(r, t);
+  TH_TENSOR_APPLY3(real, r, real, t, real, src,
+                   *r_data = *t_data > *src_data ? *t_data : *src_data;);
+}
+
+void THTensor_(cmin)(THTensor *r, THTensor *t, THTensor *src) {
+  THTensor_(resizeAs)(r, t);
+  TH_TENSOR_APPLY3(real, r, real, t, real, src,
+                   *r_data = *t_data < *src_data ? *t_data : *src_data;);
+}
+
+void THTensor_(cmaxValue)(THTensor *r, THTensor *t, real value) {
+  THTensor_(resizeAs)(r, t);
+  TH_TENSOR_APPLY2(real, r, real, t,
+                   *r_data = *t_data > value ? *t_data : value;);
+}
+
+void THTensor_(cminValue)(THTensor *r, THTensor *t, real value) {
+  THTensor_(resizeAs)(r, t);
+  TH_TENSOR_APPLY2(real, r, real, t,
+                   *r_data = *t_data < value ? *t_data : value;);
+}
+
 void THTensor_(zeros)(THTensor *r_, THLongStorage *size)
 {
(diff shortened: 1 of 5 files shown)


dominikgrewe:
CUDA implementations at torch/cutorch#198

lib/TH/generic/THTensorMath.c
```diff
@@ -1066,6 +1066,30 @@ void THTensor_(cross)(THTensor *r_, THTensor *a, THTensor *b, int dimension)
                        r__data[2*r__stride] = a_data[0*a_stride]*b_data[1*b_stride] - a_data[1*a_stride]*b_data[0*b_stride];);
 }
 
+void THTensor_(cmax)(THTensor *r, THTensor *t, THTensor *src) {
+  THTensor_(resizeAs)(r, t);
+  TH_TENSOR_APPLY3(real, r, real, t, real, src,
```
tkoeppe: This is Stepanov's Mistake in action (it took Stepanov 15 years to get min and max right):

You get away with this in C because you are comparing a fixed set of types, but the logic is flawed here: Suppose `<` is only a strict weak order which has equivalent elements, and suppose that `a` and `b` are equivalent but non-equal. You really want `min(a, b)` and `max(a, b)` to be two different elements!

The correct way to write these algorithms is to keep the comparison consistent (and parcimonious, using only a single operator):

    max(a, b) = a < b ? b : a
    min(a, b) = a < b ? a : b

In a nutshell, all the ordering operations (min, max, stable partition, nth element, partial sort, etc.) need to be defined consistently with respect to one another.

dominikgrewe: This code is consistent with the rest of Torch (see for example the implementations of `minall` and `maxall`). If we wanted to change it, it should be changed everywhere.

hughperkins:
@tkoeppe these are floats.  Can you give an example of two floats which are 'equivalent but not equal'?  I came up with: a pair of nans.  I think infinities are equal in IEEE floats? (certainly, in th:
```
th> a = 1/ 0
                                                                      [0.0001s]
th> b = 2 / 0
                                                                      [0.0001s]
th> a == b
true	
                                                                      [0.0001s]
th> a < b
false	
                                                                      [0.0001s]
th> a > b
false	
                                                                      [0.0001s]
th> b < a
false	
                                                                      [0.0001s]
th> b > a
false	
                                                                      [0.0001s]
th> 
```

and for nan:
```
th> a = 0/0
                                                                      [0.0001s]
th> b = 0/0
                                                                      [0.0001s]
th> a == b
false	
                                                                      [0.0001s]
th> a == a
false	
                                                                      [0.0001s]
th> b == b
false	
                                                                      [0.0001s]
th> a < b
false	
                                                                      [0.0001s]
th> a > b
false	
                                                                      [0.0000s]
th> b < a
false	
                                                                      [0.0001s]
th> b > a
false	
                                                                      [0.0001s]
```
)

tkoeppe:
Floats aren't a problem, which is why I said that you're probably fine in C. The problem is purely conceptual at this point; it is a misunderstanding of what "min" and "max" mean, rather than a miscomputation. The result of the computation happens to be right, but the understanding isn't. If someone takes the same approach and applies it to a different type of data (e.g. indirection through pointers?), they may surprisingly find that their algorithm doesn't actually work.

So no action is strictly required, but I would find emotional value in polishing the details of the high-level concepts in a math-based library :-)

dominikgrewe:
@soumith Any thoughts on this PR?

soumith:
your tests have lots of non-local variables. I think if you fix them, the PR is good to go. (the variables a,b,c,v)

dominikgrewe:
Sorry about that. That's now fixed.

soumith:
Thanks Dominik

dominikgrewe:
Thanks Soumith. There's also an equivalent cutorch PR: torch/cutorch#198

PR #291, opened 2015-07-10
do not use _aligned_malloc in WIN32: it requires _aligned_[free|realloc]

diz-vara:
Sam Gross in
9ad68c82f6d5a11db05ec667cbd78d8909c120d2
used _aligned_malloc 
but _aligned_malloc in Win32 requires _alinged_free in all other places (and in those places we do not know if we allocated with malloc - or with _aligned_malloc) - and resulted in crashes in garbage collector


colesbury:
oops! thanks for catching this

Issue #289, opened 2015-07-10
Problem with pointwise math when output tensor is zero-strided with correct size


fmassa:
When reusing a `Tensor` as the output of a pointwise math operation (like `cmul`, `add`, `cdiv` etc), the result is not properly computed if the output `Tensor` has zero stride and has the correct size.

```lua
a = torch.Tensor(1,5):expand(5,5)
b = torch.rand(5,5)
c = torch.rand(5,5)

print(a:cmul(b,c)) -- computes the wrong multiplication
print(torch.cmul(b,c)) -- correct answer
```
it outputs

```lua
th> print(a:cmul(b,c)) -- computes the wrong multiplication
 0.1019  0.0165  0.0757  0.7305  0.0191
 0.1019  0.0165  0.0757  0.7305  0.0191
 0.1019  0.0165  0.0757  0.7305  0.0191
 0.1019  0.0165  0.0757  0.7305  0.0191
 0.1019  0.0165  0.0757  0.7305  0.0191
[torch.DoubleTensor of size 5x5]

th> print(torch.cmul(b,c)) -- correct answer
 0.0222  0.0103  0.0174  0.2752  0.1085
 0.2826  0.2686  0.0294  0.5893  0.6549
 0.2514  0.0066  0.0279  0.0920  0.0991
 0.6117  0.4126  0.1853  0.0232  0.4656
 0.1019  0.0165  0.0757  0.7305  0.0191
[torch.DoubleTensor of size 5x5]

```

This happens because the `THTensor_(resizeAs)` function doesn't take into account the fact that even though the sizes match, there can be a zero stride somewhere.

Maybe a solution would be checking in `THTensor_(isSameSizeAs)` if there is one of the strides==0 and returning 0 in this case ? I don't know if there could be side effects though.

dominikgrewe:
There are probably many functions out there that use `THTensor_(isSameSizeAs)` to determine if arguments have the correct size. If you now say that two arguments are different just because one has stride=0 these functions will fail (probably unnecessarily because for inputs stride=0 doesn't matter).

I agree that stride=0 is a problem for outputs though. Not sure how to fix it without checking for it in each function...

fmassa:
The solution I proposed doesn't solve the problem. Actually, even if `THTensor_(rawResize)` was called, as the sizes are the same it wouldn't modify the strides. We would need for example to reset the sizes if there is a zero stride before calling `THTensor_(rawResize)`.

[...]

Issue #58, opened 2014-07-30
blas fails for gemm with zero strides


soumith:
```
th> torch.randn(1,2) * torch.randn(2,1):expand(2,2)
 ** On entry to DGEMM  parameter number  8 had an illegal value
0 0
[torch.DoubleTensor of dimension 1x2]
```
thanks to tudor bosman for reporting this

soumith:
fixed via #292

PR #292, opened 2015-07-13
Fix for gemm with zero strides plus add unit test for torch.mm

fmassa:
- Solves https://github.com/torch/torch7/issues/58
- Also solves the issue of gemm when the resulting tensor has zero stride (as in https://github.com/torch/torch7/issues/289)
- Adds unit test for `torch.mm`, which was missing


soumith:
Thanks franceyyy!

PR #295, opened 2015-07-15
Force resize when output tensor has zero stride

fmassa:
Simple solution to https://github.com/torch/torch7/issues/289 . Do you think that's the way to go ?


soumith:
I dont think this is the right solution. It is doing something unintended. I guess we will have to go through the math functions and cover this case.

PR #307, opened 2015-07-26
add shared memory file mapping with shm_open

koraykv:
I don't have access to a windows machine, so this is only for linux/mac.


dominikgrewe:
We should document this here: https://github.com/torch/torch7/blob/master/doc/storage.md#torchtypestoragefilename--shared--size

koraykv:
Correct! I will do it today.

koraykv:
just did through #309

PR #308, opened 2015-07-27
torch Tester: run tests in alphabetical order, and support interrupt handler

adamlerer:
It's good for tests to run in a deterministic, sensible order (alphabetical).
Also, it's nice to be able to exit the tester through Ctrl-C when the tests are slow/hang/fail. If anyone has a better way to do that then my error string grepping, please suggest it.


koraykv:
I think it is good, it might have rare false positives, but I don't have anything better.

PR #309, opened 2015-07-27
update doc for shared memory file

koraykv:


doc/storage.md
```diff
@@ -134,6 +134,11 @@ If `size` is not specified then the [size](#torch.Storage.size) of the returned
 ```
 elements.
 
+if `sharedMem` is true then, the file will be created (or mapped from) the shared
```
dominikgrewe: sharedMem isn't mentioned in the parameters above.

koraykv:
#310

PR #316, opened 2015-07-31
Tests and documentation for transpositions in LAPACK functions

GeorgOstrovski:
Added some tests for the previously failing cases of wrong transpositions for in-place LAPACK functions. Added comments to the documentation regarding the fact that return values have non-standard strides.

[ My editor clearly dislikes trailing whitespace, sorry for the noise ]


soumith:
thanks georg!

PR #320, opened 2015-08-03
Exposing the lapack function trtrs which solves triangular systems of linear equations.

yozw:
Lapack has optimized functions for solving systems of equations Ax = b in which the matrix A is upper- or lower-triangular. This pull request exposes this function through a new function torch.trtrs.

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -1036,6 +1036,24 @@ static void THTensor_random1__(THTensor *self, THGenerator *gen, long b)
                          {name=Tensor}}
                      )
       end
+      interface:wrap("trtrs",
+                     cname("trtrs"),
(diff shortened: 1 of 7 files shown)


soumith:
hey @yozw , the unit test you added seems to be failing: https://travis-ci.org/torch/torch7/jobs/73904453

yozw:
Yes, i'm not sure what went wrong there: some strings had an extra space in them. I've fixed it. The latest version should have passing tests.

soumith:
thanks. I think looks good. cc: @koraykv

koraykv:
Yes, looks good, thanks Soumith.

yozw:
Thanks, Soumith.

PR #324, opened 2015-08-06
Fix lua GC to support allocations across multiple threads

adamlerer:
With torch/threads, Tensors can be allocated in one thread and deallocated in another, so `heapSize` must not be thread-local. Note that I am leaving `heapSoftMax` as thread-local, so each thread will maintain it's own softmax.

Here's an example that this PR fixes:

```lua
```require 'torch'
local Threads = require 'threads'
Threads.serialization('threads.sharedserialize')

local a = torch.Tensor(1e7):zero()
pool = Threads(1, function() require 'torch'; end);
pool:addjob(function() for i = 1,1000 do a:add(1); end; a = nil;  collectgarbage(); end)
a = nil
collectgarbage()
pool:synchronize()
```

If you add an assert that `heapSize >= 0` prior to this PR, it will fail (in the pool thread).

### DIFF
--- a/lib/TH/THGeneral.c
+++ b/lib/TH/THGeneral.c
@@ -1,11 +1,11 @@
 #include "THGeneral.h"
+#include "THAtomic.h"
 
 #ifndef TH_HAVE_THREAD
 #define __thread
 #endif
 
-#if defined(TH_DISABLE_HEAP_TRACKING)
-#elif (defined(__unix) || defined(_WIN32))
+#if (defined(__unix) || defined(_WIN32))
 #include <malloc.h>
 #elif defined(__APPLE__)
 #include <malloc/malloc.h>
@@ -101,8 +101,10 @@ void THSetArgErrorHandler( void (*torchArgErrorHandlerFunction_)(int argNumber, 
 
 static __thread void (*torchGCFunction)(void *data) = NULL;
 static __thread void *torchGCData;
-static __thread long torchHeapSize = 0;
-static __thread long torchHeapSizeSoftMax = 300000000; // 300MB, adjusted upward dynamically
+static long heapSize = 0;
+static __thread long heapSoftmax = 300000000; // 300MB, adjusted upward dynamically
+static const double heapSoftmaxGrowthThresh = 0.8; // grow softmax if >80% max after GC
+static const double heapSoftmaxGrowthFactor = 1.4; // grow softmax by 40%
 
 /* Optional hook for integrating with a garbage-collected frontend.
  *
@@ -121,9 +123,7 @@ void THSetGCHandler( void (*torchGCFunction_)(void *data), void *data )
 }
 
 static long getAllocSize(void *ptr) {
-#if defined(TH_DISABLE_HEAP_TRACKING)
-  return 0;
-#elif defined(__unix)
+#if defined(__unix)
   return malloc_usable_size(ptr);
 #elif defined(__APPLE__)
   return malloc_size(ptr);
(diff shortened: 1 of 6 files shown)


andresy:
Looks good to me. @clementfarabet @koraykv @dominikgrewe is this ok?

koraykv:
I have some concerns about it. As far as I understand, we assume that each thread is in control of only one lua state and that lua state dies when the thread dies, I think this is quite limiting. Since the idea of introducing heap check is a proxy for Lua to have an idea of how much malloc'd space it is controlling, maybe we should push these counters to lua state and not thread local. Anyways, I think this change should go in but I think we should by default disable heap check to avoid all sorts of problems.

soumith:
disabling by default, and interested parties enabling it for a few months and reporting back with any adverse affects is a good idea. a change so fundamental makes me uncomfortable to push into master right away considering we dont have versioning.

dominikgrewe:
I agree that this should be disabled by default for now. In cutorch we have a good solution to push these counters into the Lua state with the THCState. I don't know how feasible it is to do something similar for torch.

lib/TH/THGeneral.c
```diff
@@ -139,17 +142,23 @@ static long getAllocSize(void *ptr) {
...
   }
 }
 
 // hooks into the TH heap tracking
 void THHeapUpdate(long size) {
-  torchHeapSize += size;
+  THAtomicAddLong(&heapSize, size);
+
+# ifdef TH_CHECK_HEAP_UPDATE
+  if (heapSize < 0)
```
dominikgrewe: I think we need to use atomics to _read_ the heap size too. Otherwise there's no guarantee that an atomic write by another thread is seen by this thread.
You could also use the return value from THAtomicAddLong and work with that instead of the global `heapSize`, if you're not concerned about concurrent updates by other threads during `THHeapUpdate`.

adamlerer:
@koraykv you're right, there's an implicit assumption that there's one lua_State per thread; this follows what's being done for torchErrorHandler and related. In both cases, it would be preferable to attach them to a state variable that's passed around, or attach them to the lua_State. 
Disabling GC by default for a while sounds like a good idea.

lib/TH/THGeneral.c
adamlerer: I think you're right, although I'm not sure I agree with your justification.

The atomic add will introduce a memory fence, so I think `heapSize` will have seen all prior updates at that point. However, if a thread writes `heapSize` after the add, the read could see tearing. I'll make sure I read `heapSize` atomically, thanks.

adamlerer:
I just pushed a version with some changes:
* everything's disabled by default; enable with `torch.setheaptracking(true)`
* Use atomic load for `heapSize`

I left the softmax and handlers thread-local. Regarding the softmax, I don't think it should be per-lua_State; it could even be global but I'm just worried about if multiple threads trigger the GC in parallel and end up bumping the softmax up to a huge value. Regarding the handlers, it's done that way to keep TH Lua-independent, just like the error handlers. If we want to refactor to fix this lets put it in a different commit.

lib/TH/THGeneral.c
```diff
@@ -138,20 +141,29 @@ static long getAllocSize(void *ptr) {
...
     }
   }
 }
 
 // hooks into the TH heap tracking
 void THHeapUpdate(long size) {
-  torchHeapSize += size;
-  if (size > 0)
-    maybeTriggerGC();
+  long newHeapSize = THAtomicAddLong(&heapSize, size);
```
dominikgrewe: `THAtomicAddLong` returns the previous value of `heapSize`, so we have to add `size` to our local `newHeapSize`.

dominikgrewe:
I wonder if it's better to use a compiler flag to enable this. It would give us an easy way to build a version of Torch with heap tracking, without needing to change any code. Wdyt?

lib/TH/THGeneral.c
adamlerer: *facepalm* will fix, sorry.

adamlerer:
I think that if you're worried about corners where heap tracking breaks, a dynamic flag allows the user to turn it off when it breaks. They may not want to turn it off everywhere, and maybe the organization is using a single version of the package so the user can't just turn it off with a compiler flag.

Adding one line in init.lua doesn't seem substantially more intrusive than adding one line in the Makefile.

dominikgrewe:
Sounds good. We should probably get rid of the TH_DISABLE_HEAP_TRACKING compiler flag then. No need to have two ways to control heap tracking.
In any case, I think we should get his merged soon, so it's disable by default on master.

adamlerer:
That was more to protect the #includes but I guess I'll just get rid of it.

adamlerer:
@dominikgrewe ping

dominikgrewe:
Sorry for the delay!

adamlerer:
Thanks Dominik and Ronan! The cutorch version needs to be merged next, https://github.com/torch/cutorch/pull/219

PR #334, opened 2015-08-14
Serialization at scale, fixes.

soumith:
When you theoretically write more than 268 million objects using `f:writeObject` on a single file, or `torch.save` a table (or tds.hash) with that many objects, the serialization fails right now due to LuaJIT 2GB limitations.

This is because of the duplicate object tracking we have in place right now in the serializer.

We already have a mode for file objects to disable this duplicate tracking, via:
```lua
f:referenced(false)
```
However, it was still tracking some pointers that are not actually used.

This PR fixes it.

### DIFF
--- a/File.lua
+++ b/File.lua
@@ -129,8 +129,8 @@ function File:writeObject(object)
          -- else write the object itself
          index = objects.nWriteObject or 0
          index = index + 1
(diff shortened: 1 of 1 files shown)


soumith:
merging after ronan approval.
cc: @koraykv @clementfarabet

PR #337, opened 2015-08-19
clean-up and simplify cloning/collection logic

koraykv:
clean-up and simplify cloning/collection logic

remove now unused cloning functions

adopt trtrs to new clean setup


koraykv:
This change makes torch lapack support quite simpler and more understandable. Giving time for others to check out, I will merge them tomorrow.

PR #339, opened 2015-08-23
Modified torch.isTypeOf to handle partial and composite type names

Muhaeb:
Modified isTypeOf to:
1- Do exact string match if the second parameter typeSpec is a ctor.
2- Do partial string match if the second parameter is not a ctor, now the behaviour handles the following two cases correctly:
  a- if obj is of type 'TorchDummy', then the call torch.isTypeOf(obj, 'OtherTorchDummy') returns false
  b - if obj is of type 'nn.Sequential' then the call torch.isTypeOf(obj, 'Sequential') returns true, whereas the opposite case returns false,
      i.e. if obj is of type 'Sequential' then the call torch.isTypeOf(obj, 'nn.Sequential') returns false

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -89,19 +89,51 @@ function torch.type(obj)
 end
 
 --[[ See if a given object is an instance of the provided torch class. ]]
+local function exactTypeMatch(obj_mt, typeSpecNameString)
+  return obj_mt.__typename == typeSpecNameString
+end
+
+local function partialTypeMatch(obj_mt, typeSpecNameString)
+  if #obj_mt.__typename < #typeSpecNameString then
+    return false
+  end
+  local objNamesArray = {}
+  local typeSpecNamesArray = {}
+  for name in string.gmatch(obj_mt.__typename, '([^.]+)') do
+    objNamesArray[#objNamesArray +1] = name
+  end
+  for name in string.gmatch(typeSpecNameString, '([^.]+)') do
+    typeSpecNamesArray[#typeSpecNamesArray +1] = name
+  end  
+  
+  local objNamesRevIndex = #objNamesArray
+  for typeSpecNamesRevIndex = #typeSpecNamesArray, 1, -1 do
+    if typeSpecNamesArray[typeSpecNamesRevIndex] ~= objNamesArray[objNamesRevIndex] then
+      return false
+    end
+    objNamesRevIndex = objNamesRevIndex - 1
+  end
+  return true
+end
+
+--[[ See if a given object is an instance of the provided torch class. ]]
 function torch.isTypeOf(obj, typeSpec)
   -- typeSpec can be provided as either a string, regexp, or the constructor. If
   -- the constructor is used, we look in the __typename field of the
   -- metatable to find a string to compare to.
+  local matchFunc
   if type(typeSpec) ~= 'string' then
     typeSpec = getmetatable(typeSpec).__typename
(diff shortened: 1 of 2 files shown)


init.lua
```diff
@@ -89,19 +89,51 @@ function torch.type(obj)
 end
 
 --[[ See if a given object is an instance of the provided torch class. ]]
+local function exactTypeMatch(obj_mt, typeSpecNameString)
+  return obj_mt.__typename == typeSpecNameString
+end
+
+local function partialTypeMatch(obj_mt, typeSpecNameString)
```
GeorgOstrovski: This looks rather heavy with allocations of new tables etc. I'm used to thinking of torch.isTypeOf as a cheap, almost free, operation. Would it not be cheaper to string.reverse your two strings first, then check for their longest common prefix, then check if there are '.' left after that in one of them (this seems to be essentially what you are checking, thought I might have misunderstood)

GeorgOstrovski: After re-reading: correct me if I'm wrong - this is effectively just checking that obj_mt.__typename is a suffix of typeSpecNameString ? Then that can certainly be accomplished with a single comparison of obt_mt.__typename with typeSpecNameString:sub(...) ?

Muhaeb: 1 - "effectively just checking that obj_mt.__typename is a suffix of typeSpecNameString ?"
the original bug reported complained that "TorchDummy" is erranously matching "OtherTorchDummy", which exactly the behaviour we are trying to avoid.

typeSpecNameString needs to be the last proper "as in after . separation" type name of obj_mt.__typename. so for example we want obj of type "Top.Middle.Low" to match with typeSpecNameStrings: "Middle.Low" and "Low".

2 - As for the performance issue, I am not sure if the cost of executing string.reverse would be lower than traversing in place, which is free, the tables: objNamesArray and typeSpecNamesArray are needed anyways to get the behaviour outlined in point 1 above.

I am kind of new to Lua in general, so probably this code needs work still.

GeorgOstrovski: My point was, you don't need the tables for that. All you seem to be checking is that typeSpecNameString is a suffix of obj_mt.__typename AND that this suffix is either preceded by a point '.' or by nothing. Roughly along these lines:
local lenSuffix, lenTotal = #typeSpecNameString, #obj_mt.__typename
if lenSuffix > lenTotal or 
  string.sub(obj_mt.__typename, lenTotal-lenSuffix+1) ~= typeSpecNameString then
  return false
end
if lenTotal == lenSuffix or
  obj_mt.__typename[lenTotal-lenSuffix] == '.' then
  return true
else
  return false
end

That could probably still be streamlined. Perhaps I'm still missing some subtlety but isn't that doing the same thing without any table allocations (and in fact without string reversal, too) ?

yozw: The problem with this approach is that, in its current form, isTypeOf allows the usage of wildcards. For example, if `tensor` is a `FloatTensor`, then `torch.isTypeOf(tensor, "*Tensor")` returns true. This functionality is not provided by the solution above, nor by mahaeb@'s solution.

I am very much in favor of removing the wildcard functionality altogether, but if we do, we should first try and understand whether and how this is actually used, because this can break existing code. The wildcards were originally used in `torch.isTensor`, but this usage has been removed in a [recent commit](https://github.com/torch/torch7/commit/bf269f0f9bc49dd68cf37a0504e2eb70688a7a85).

PR #350, opened 2015-09-02
More flexible torch.cat

dominikgrewe:
Adds a version of torch.cat that takes an array of tensors as inputs.

Adds TensorArray type to cwrap that reads an array of tensors from Lua
and creates a C array of tensor pointers that are passed to the C
function.

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -164,6 +164,12 @@ for _,Tensor in ipairs({"ByteTensor", "CharTensor",
                 return string.format("TH%s_nDimension(%s)", Tensor, arg.args[argn]:carg())
              end
    end
+
+   local function lastdimarray(argn)
+      return function(arg)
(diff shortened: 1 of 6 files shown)


soumith:
there's no way I can not screw up reviewing this PR. cc: @andresy to review this, and anyone else that is interested in a review.

soumith:
lgtm. thanks dominik.

szagoruyko:
awesome :+1:

szagoruyko:
so any other type raises an error, example:
```lua
torch.cat({torch.rand(8,1):float(),torch.rand(8,1):float()},2)
```
error:
```
```

dominikgrewe:
Unfortunately Torch can't deduce the right function to call in this case and therefore uses the default tensor type; see https://github.com/torch/torch7/blob/master/TensorMath.lua#L51
The function looks at the arguments to see which type to use. We could extend it to support tables if needed, i.e. if an argument is a table it checks whether the table contains tensors and uses that type.

If you specify an output tensor it should work though.

PR #351, opened 2015-09-05
Fix torch.isTypeOf(<string>, ...)

akhti:
Fix the following cases (they raise error now):
torch.isTypeOf("string", "nn.Module")

require 'nn';
m = nn.Module()
m.name = "module"
nn.utils.recursiveType(m, "torch.FloatTensor")


soumith:
Thanks!

PR #353, opened 2015-09-06
Gate THHeapUpdate on heap tracking enabled

adamlerer:
Updating heapSize is a perf problem in some multi-threaded applications.

### DIFF
--- a/lib/TH/THGeneral.c
+++ b/lib/TH/THGeneral.c
@@ -150,16 +150,18 @@ static void maybeTriggerGC(long curHeapSize) {
 
 // hooks into the TH heap tracking
 void THHeapUpdate(long size) {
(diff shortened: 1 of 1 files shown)


adamlerer:
@kmafb

kmafb:
w00t. @adamlerer is going to come up with something that solves the OOM problem without bouncing a cacheline halfway around the world every time we make a tensor.

dominikgrewe:
Doesn't this mean that the heap size can go negative now if you enable heap tracking after you've allocated a tensor?

adamlerer:
That's correct. This was a patch to avoid some serious perf problems we were seeing; these should be resolved by https://github.com/torch/torch7/pull/356. Abandoning this PR.

PR #356, opened 2015-09-08
Batch updates to global heapSize

adamlerer:
Contribution from @colesbury

Multithreaded Torch programs that do many small THAllocs were
encountering contention on updating heapSize. Instead, keep track of
allocations and frees in a thread-local variable, heapDelta, and only
update the global heapSize when heapDelta exceeds +/- 1MB.

### DIFF
--- a/lib/TH/THGeneral.c
+++ b/lib/TH/THGeneral.c
@@ -102,7 +102,9 @@ void THSetArgErrorHandler( void (*torchArgErrorHandlerFunction_)(int argNumber, 
 static __thread void (*torchGCFunction)(void *data) = NULL;
 static __thread void *torchGCData;
 static long heapSize = 0;
-static __thread long heapSoftmax = 300000000; // 300MB, adjusted upward dynamically
+static __thread long heapDelta = 0;
+static const long heapMaxDelta = 1e6; // limit to +/- 1MB before updating heapSize
+static __thread long heapSoftmax = 3e8; // 300MB, adjusted upward dynamically
 static const double heapSoftmaxGrowthThresh = 0.8; // grow softmax if >80% max after GC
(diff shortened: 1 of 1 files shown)


soumith:
With this PR, is https://github.com/torch/torch7/pull/353 still relevant?

adamlerer:
nope, I abandoned it

soumith:
I'll merge this in tomorrow unless there are objections. cc: @dominikgrewe , we hit a pretty serious perf-regression without this patch, when we have 30 threads doing hogwild.

lib/TH/THGeneral.c
```diff
@@ -139,10 +141,14 @@ static long getAllocSize(void *ptr) {
...
 static void maybeTriggerGC(long curHeapSize) {
-  if(torchGCFunction && curHeapSize > heapSoftmax ) {
+  if (torchGCFunction && curHeapSize > heapSoftmax) {
     torchGCFunction(torchGCData);
-    long newHeapSize = THAtomicGetLong(&heapSize);
-    if(newHeapSize > heapSoftmax * heapSoftmaxGrowthThresh) {
+
+    // ensure heapSize is accurate before updating heapSoftmax
+    long newHeapSize = THAtomicAddLong(&heapSize, heapDelta) + heapDelta;
+    heapDelta = 0;
```
dominikgrewe: nit: Maybe factor this out into its own function, something like `applyHeapDelta` or `updateHeapSize`? I know it's only two lines of code but it's easy to get wrong, e.g. forgetting to reset heapDelta.

dominikgrewe:
Looks good to me. I guess we should do the same in cutorch then.

dominikgrewe:
Just saw the cutorch version :)

PR #366, opened 2015-09-17
Clarify that `torch.random` result is unsigned

ninjin:
Minor clarification to the doc in that the result of `torch.random` is always unsigned.


soumith:
thanks!

PR #372, opened 2015-09-21
Track and print the serialization 'stack trace' on failure

adamlerer:
When serialization fails, it's hard to figure out what object can't be synchronized. Especially comes up in torch-threads. Any interest in this PR?

With this PR, you would now get the following error:

```
luajit: ...al/git/torch-distro/install/share/lua/5.1/torch/File.lua:115: Unwritable object <function>
at <?>.outer.theinner.baz.torch.cat
```

for this code

```lua
   local torch = require 'torch'
   local inner = {
       baz = function(a) torch.somefunc() end
   }
   local outer = {
       theinner = inner
   }
   function evil_func()
      outer.prop = 1
   end
   torch.save('saved.t7', evil_func)
end
```

### DIFF
--- a/File.lua
+++ b/File.lua
@@ -49,7 +49,7 @@ end
 function File:referenced(ref)
    -- we use an environment to keep a record of written objects
    if not torch.getenv(self).writeObjects then
-      torch.setenv(self, {writeObjects={}, writeObjectsRef={}, readObjects={}})
+      torch.setenv(self, {writeObjects={}, writeObjectsRef={}, readObjects={}, objectNameStack={}})
    end
    local env = torch.getenv(self)
    env.force = not ref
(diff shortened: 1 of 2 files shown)


hughperkins:
I'm a fan of useful error messages :-)

soumith:
i'm definitely up for this PR, quite useful to debug cryptic error messages (in the context of saving / loading, as well as threads serialization). Anyone else have any thoughts?

hughperkins:
Are there any disadvantages of this PR? Performance?

soumith:
there's no disadvantages except a thorough code review is needed. It touches upon critical code.

albanD:
I think there is a problem with the env initialization [l52](https://github.com/adamlerer/torch7/blob/3ecf3e9b23691741d9cd50604efc6b3c77a8ff02/File.lua#L52) and [l209](https://github.com/adamlerer/torch7/blob/3ecf3e9b23691741d9cd50604efc6b3c77a8ff02/File.lua#L209), these lines should be the same as [l99](https://github.com/adamlerer/torch7/blob/3ecf3e9b23691741d9cd50604efc6b3c77a8ff02/File.lua#L99).

If you perform either a `referenced` or `readObject` call on the file before doing `writeObject`, you will get an error.

For example to reproduce: if you want to use a memory file without the ref:
```lua
local file = torch.MemoryFile('rw'):binary()
file:referenced(false)

local foo = 'bar'
file:writeObject(foo)
```
will fail since [l110](https://github.com/adamlerer/torch7/blob/3ecf3e9b23691741d9cd50604efc6b3c77a8ff02/File.lua#L110) the table `objectNameStack` does not exist.

Maybe we could add this small example to the test suite.

adamlerer:
@colesbury fixed the problem pointed out by @albanD, updating on his behalf.

albanD:
@adamlerer thanks.
@soumith This can be merge now ?

soumith:
fingers crossed.

deltheil:
> Where is self:_writeObjectGetStackStr defined?

It looks like it should be replaced with `formatStack(objectNameStack))` (it was introduced in the [initial commit](https://github.com/adamlerer/torch7/blob/3ecf3e9b23691741d9cd50604efc6b3c77a8ff02/File.lua#L91)).

colesbury:
Ooops!

PR #377, opened 2015-09-26
add lua function to update thread locals

koraykv:


soumith:
thanks. this will be somewhat useful for something I am running into.
When using co-routines, the errors emitted from a co-routine are corrupted (either garbage tensors are printed from the main stack, or the stack-trace is non-sensical).

One has to update the error handlers everytime you go into a co-routine, and when you get out of it, and this puts a reasonable story around it. (still a bit more work to do).

PR #378, opened 2015-09-26
patched luaT to allow for nested modules

j-wilson:
Patch allows for declaration of 'nested' modules/classes, e.g., `torch.class('A.b.c')`, where `b` is a module in package `A` and `c` is a submodule of `b`.

Should likely check to ensure full backwards compatibility.


soumith:
cc: @andresy

j-wilson:
Added better error checking for nested modules. If a specified module path is invalid, error report will now indicate where said path goes wrong. 

By way of example: 
If module `c` is declared with path `A.b.c`, but `A.b` is not a valid module; then, error report will read as:
```lua
...: while creating metatable A.b.c: bad argument #1 (A.b. is an invalid module name)
```

soumith:
cc: @dominikgrewe

PR #382, opened 2015-09-30
patch to luaT.c allowing for nested modules

j-wilson:
Patch allows for declaration of 'nested' modules/classes, e.g., `torch.class('A.b.c')`, where `b` is a module in package `A` and `c` is a submodule of `b`.

If a specified module path is invalid, error report will now indicate where said path goes wrong. By way of example: 
If module `c` is declared with path `A.b.c`, but `A.b` is not a valid module; then, error report will read as:
```lua
...: while creating metatable A.b.c: bad argument #1 (A.b. is an invalid module name)
```

Should likely check to ensure full backwards compatibility. 

cc: @andresy, @dominikgrewe

### DIFF
--- a/lib/luaT/README.md
+++ b/lib/luaT/README.md
@@ -220,16 +220,31 @@ This function assume a table is on the stack. It creates a table field
 <a name="luat_classrootname"/>
 ### const char *luaT_classrootname(const char *tname) ###
 
-Assuming `tname` is of the form `modulename.classname`, returns
-`classname`. The returned value shall not be freed. It is a pointer
-inside `tname` string.
+Assuming `tname` is of the form `A.b.c`, returns 'c'. The returned value
+shall not be freed. It is a pointer inside `tname` string.
 
 <a name="luat_classmodulename"/>
-### const char *luaT_classmodulename(const char *tname) ###
+### int luaT_classmodulename(const char *tname, char *parent_name) ###
+Alias to `luaT_fullparentname ` for ensuring backwards compatibilty; 
+use of `luaT_fullparentname` is preferred.
 
-Assuming `tname` is of the form `modulename.classname`, returns
-`modulename`. The returned value shall not be freed. It is valid until the
-next call to `luaT_classrootname`.
+<a name="luat_fullparentname"/>
+### int luaT_fullparentname(const char *tname, char *parent_name) ###
+
+Returns a 0-1 valued integer indicating whether `tname` has a parent module.
+Assuming `tname` is of the form `A.b.c`, sets `parent_name` to `A.b`.
+
+<a name="luat_classmodulename"/>
+### int luaT_outerparentname(const char *tname, char *parent_name) ###
+
+Returns a 0-1 valued integer indicating whether `tname` has a parent module.
+Assuming `tname` is of the form `A.b.c`, sets `parent_name` to `A`.
+
+<a name="luat_classmodulename"/>
+### int luaT_innerparentname(const char *tname, char *parent_name) ###
+
+Returns a 0-1 valued integer indicating whether `tname` has a parent module.
+Assuming `tname` is of the form `A.b.c`, sets `parent_name` to `b`.
 
 <a name="luat_stackdump"/>
 ### void luaT_stackdump(lua_State *L) ###
--- a/lib/luaT/luaT.c
+++ b/lib/luaT/luaT.c
@@ -138,12 +138,12 @@ static int luaT_mt__call(lua_State *L);
 static int luaT_cmt__call(lua_State *L);
 static int luaT_cmt__newindex(lua_State *L);
 
-const char* luaT_newmetatable(lua_State *L, const char *tname, const char *parenttname,
+const char* luaT_newmetatable(lua_State *L, const char *tname, const char *parent_tname,
                               lua_CFunction constructor, lua_CFunction destructor, lua_CFunction factory)
 {
   lua_pushcfunction(L, luaT_lua_newmetatable);
   lua_pushstring(L, tname);
-  (parenttname ? lua_pushstring(L, parenttname) : lua_pushnil(L));
+  (parent_tname ? lua_pushstring(L, parent_tname) : lua_pushnil(L));
   (constructor ? lua_pushcfunction(L, constructor) : lua_pushnil(L));
   (destructor ? lua_pushcfunction(L, destructor) : lua_pushnil(L));
   (factory ? lua_pushcfunction(L, factory) : lua_pushnil(L));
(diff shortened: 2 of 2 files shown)


j-wilson:
@koraykv: Hi there, I actually opened this PR a few days ago now (had to re-open it earlier today). Would you (or anyone else) have time to give it a look over in the the near future?

lib/luaT/luaT.c
```diff
@@ -497,27 +497,77 @@ int luaT_classmodulename(const char *tname, char *module_name)
   return tname[n] == '.';
 }
 
-/* Lua only functions */
-int luaT_lua_newmetatable(lua_State *L)
+/* Method for determining name of class module */
+const char* luaT_classname(const char *tname)
```
dominikgrewe: I think we should clarify how `luaT_classrootname` and `luaT_classmodulename` should behave for nested modules. `luaT_classrootname` should probably behave just like your `luaT_classname`.

lib/luaT/luaT.c
```diff
@@ -497,27 +497,77 @@ int luaT_classmodulename(const char *tname, char *module_name)
...
+  {
+    if(!lua_istable(L, *L_idx)){
+      strncpy(term, tname_full, tname - tname_full - 1);
+      term[tname - tname_full] = '\0';
+      luaL_error(L, "while creating metatable %s: bad argument #1 (%s is an invalid module name)", tname_full, term);
+    }
+    strncpy(term, tname, n);
+    term[n] = '\0';
+    lua_getfield(L, *L_idx, term);
+    *L_idx += 1;
```
dominikgrewe: If you remove the parent table after the `lua_getfield` call, you can avoid all the index computation.

lib/luaT/luaT.c
```diff
@@ -497,27 +497,77 @@ int luaT_classmodulename(const char *tname, char *module_name)
...
   luaL_argcheck(L, lua_isnoneornil(L, 2) || lua_isstring(L, 2), 2, "parent class name or nil expected");
   luaL_argcheck(L, lua_isnoneornil(L, 3) || lua_isfunction(L, 3), 3, "constructor function or nil expected");
   luaL_argcheck(L, lua_isnoneornil(L, 4) || lua_isfunction(L, 4), 4, "destructor function or nil expected");
   luaL_argcheck(L, lua_isnoneornil(L, 5) || lua_isfunction(L, 5), 5, "factory function or nil expected");
 
-  if(is_in_module)
-    lua_getglobal(L, module_name);
+  const char* tname = luaL_checkstring(L, 1);
+  char module_name[256];
+  int is_in_module = 0;
```
dominikgrewe: I think we still want Torch to be compatible with C99 at this point. So the variables (and the ones below) must be declared at the top of the function.

lib/luaT/luaT.c
```diff
@@ -497,27 +497,77 @@ int luaT_classmodulename(const char *tname, char *module_name)
...
+  int L_idx0 = lua_gettop(L) + 1; // stack position of outermost module
+  int L_idx1 = L_idx0;            // stack position of immediate parent
+  if (is_in_module)
+    traverse_tables(L, tname, &L_idx1);
   else
     lua_pushglobaltable(L);
-  if(!lua_istable(L, 6))
-    luaL_error(L, "while creating metatable %s: bad argument #1 (%s is an invalid module name)", tname, module_name);
 
+  if(!lua_istable(L, L_idx0))
```
dominikgrewe: Don't you want to check the innermost parent, i.e. `L_idx1`?

dominikgrewe:
Sorry for the delay in reviewing this. I've add some comments.

j-wilson:
Thanks for the feedback; addressing comments now.

lib/luaT/luaT.c
j-wilson: The mentioned check happens in `traverse_table` anyhow; so, both are interchangeable. Will update when I remove index computation (as advised above).

lib/luaT/luaT.c
```diff
@@ -497,27 +496,65 @@ int luaT_classmodulename(const char *tname, char *module_name)
   return tname[n] == '.';
 }
 
-/* Lua only functions */
+void traverse_tables(lua_State *L, const char *tname, int *L_idx)
```
dominikgrewe: `L_idx` isn't updated any more, do you can just pass it by value.

dominikgrewe: In fact, you can just use relative indices in this function, couldn't you? Then you don't need to pass in the index at all. Or the function get it itself via `lua_gettop`.

lib/luaT/luaT.c
```diff
@@ -471,16 +471,15 @@ void luaT_registeratname(lua_State *L, const struct luaL_Reg *methods, const cha
 }
 
 
-/* utility functions */
-const char *luaT_classrootname(const char *tname)
+const char* luaT_classrootname(const char *tname)
```
dominikgrewe: Can you update the documentation for this method? Also, what about `luaT_classmodulename`? If you pass in `a.b.c` do you want it to return `a` (as it currently does) or `a.b`?

lib/luaT/luaT.c
j-wilson: I would imagine that passing in the index is more efficient; as it avoid either an extra call to lua_gettop() or use of relative indices (e.g. -1). That said, I can certainly update it if preferred.

lib/luaT/luaT.c
j-wilson: Good point. Currently, it's a non-issue in that `luaT_classmodulename` seems to be single use; however, I agree that clarification here would be desirable. An obvious patch would be to make seperate methods a la `luaT_outerparentname` and `luaT_innerparentname`. Note that the latter method is (in effect) what `traverse_table` accomplished. Please let me know how you would like me to proceed.

dominikgrewe: It's single use within this file but it's also a public function that can be used externally. So I think we should keep it, but maybe change its behaviour? It seems more natural to return `a.b` as the module name for a class `a.b.c` doesn't it? So I'd change the code of the function to do that and update the documentation accordingly.

j-wilson: I'd agree that returning `a.b` makes sense. My one concern there is that `a.b` isn't directly accessible from C (i.e. you can't call `_G['a.b']`).

lib/luaT/luaT.c
dominikgrewe: I think I prefer relative indices in this case as they are easier to read (-1 = what I just pushed on the stack). I'm not sure if there are any performance implications but I can't imagine luaT_newmetatable being on a critical path.

j-wilson: Very true. I'll update it accordingly.

lib/luaT/luaT.c
j-wilson: Please let me know how you'd like me to follow up on the above.

j-wilson:
@dominikgrewe Not sure what the best way to resolve class name getters. For now, I'll go ahead and add in methods s.t. given `A.b.c`:
   -  `luaT_outerparentname(A.b.c)` return `A`  (outermost parent)
   -  `luaT_innerparentname(A.b.c)` returns `b` (innermost parent)
   -  `luaT_fullparentname(A.b.c)` returns `A.b` (full parent hierarchy)

This might be a bit of overkill; but, it should ideally handle all foreseeable usage cases. Please let me know if this works for you.

lib/luaT/luaT.c
```diff
@@ -471,53 +471,136 @@ void luaT_registeratname(lua_State *L, const struct luaL_Reg *methods, const cha
...
 
-/* module_name must be a buffer at least as big as tname 
- * return true if the class is part of a module */
-int luaT_classmodulename(const char *tname, char *module_name)
+/* parent_name must be a buffer at least as big as tname.
+ * If class has a parent, returns true; and, sets
+ * parent name to that of full parent hierarchy (e.g.
+ * given class `A.b.c`, sets parent_name to `A.b`)
+ */
+int luaT_fullparentname(const char *tname, char *parent_name)
```
dominikgrewe: `luaT_classmodulename` is part of luaT's public API, so we shouldn't change that. For the `module.class` case (i.e. no nesting) this implementation behaves just like the old one, so I'd just rename this function back to `luaT_classmodulename`.

lib/luaT/luaT.c
```diff
@@ -471,53 +471,136 @@ void luaT_registeratname(lua_State *L, const struct luaL_Reg *methods, const cha
...
+
+  for(head = tail; head >= 1; head--)
+    if(tname[head-1] == '.') break;
+
+  strncpy(parent_name, tname+head, tail-head);
+  parent_name[tail-head] = '\0';
+  return 1;
+}
+
+void traverse_tables(lua_State *L, const char *tname)
```
dominikgrewe: Can you give this a more descriptive name? Maybe something like `loadinnerparent`?

j-wilson: Updated to: `luaT_getinnerparent`. Hope that works; otherwise, let me know if you'd like me to change it.

lib/luaT/luaT.c
j-wilson: I've added an alias of sorts s.t. calling `luaT_classmodulename` immediately redirects as a call to `luaT_fullparentname`. Please let me know if you'd like prefer I simply rename `luaT_fullparentname`; for the time being, I chose above recourse in that `luaT_fullparentname` seems to cohere better with `luaT_outerparentname` and `luaT_innerparentname`.

lib/luaT/luaT.c
```diff
@@ -471,53 +471,146 @@ void luaT_registeratname(lua_State *L, const struct luaL_Reg *methods, const cha
...
   lua_settop(L, 5);
   luaL_argcheck(L, lua_isnoneornil(L, 2) || lua_isstring(L, 2), 2, "parent class name or nil expected");
   luaL_argcheck(L, lua_isnoneornil(L, 3) || lua_isfunction(L, 3), 3, "constructor function or nil expected");
   luaL_argcheck(L, lua_isnoneornil(L, 4) || lua_isfunction(L, 4), 4, "destructor function or nil expected");
   luaL_argcheck(L, lua_isnoneornil(L, 5) || lua_isfunction(L, 5), 5, "factory function or nil expected");
 
-  if(is_in_module)
-    lua_getglobal(L, module_name);
+  // -------- Push immediate parent module to stack
+  is_in_module = luaT_outerparentname(tname, parent_name);
```
dominikgrewe: Do you really want outer parent name here? If `tname` is `a.b.c` but `a.b` is not a table, you will return an error saying that `a` is an invalid module name, won't you? Wouldn't full parent name be better?

lib/luaT/luaT.c
dominikgrewe: If you want to keep outer/inner parent name then I'm fine with your solution. I'm not sure if they're strictly needed though. You can always get those with `classmodulename` and `classrootname`. But I don't have a strong opinion here.

In any case, the documentation in lib/luaT/README.md must be updated.

lib/luaT/luaT.c
j-wilson: So, there are additional checks within 'luaT_getinnerparent' that are likely the source of the confusion here. As is, the code is correct; but, perhaps a touch unclear. I'll update it s.t. it's easier to understand.

lib/luaT/luaT.c
j-wilson: Updated README.

j-wilson:
Tried updating `luaT.c: line 604` along the lines you mentioned. Attempted changes produced an error however; not sure why. As is, said code should be correct however.

lib/luaT/luaT.c
dominikgrewe: `lua_getinnerparent` doesn't check that the value it leaves on the stack is a table, right? That's why the additional check below is needed. All I'm saying is that the error message would be confusing if that check fails, because it refers to the outer parent whereas it should refer to the inner parent.

Please correct me if I misunderstood something here...

lib/luaT/README.md
```diff
@@ -225,11 +225,27 @@ Assuming `tname` is of the form `modulename.classname`, returns
 inside `tname` string.
```
dominikgrewe: Can you also update the docs for `luaT_classrootname` to clarify what happens for nested modules?

j-wilson: Updated.

lib/luaT/luaT.c
```diff
@@ -471,53 +471,146 @@ void luaT_registeratname(lua_State *L, const struct luaL_Reg *methods, const cha
...
+
+  for(head = tail; head >= 1; head--)
+    if(tname[head-1] == '.') break;
+
+  strncpy(parent_name, tname+head, tail-head);
+  parent_name[tail-head] = '\0';
+  return 1;
+}
+
+/* Method for pushing a classes immediate parent to the
```
dominikgrewe: nit: "classes" -> "class's"
Or just change it to "Method for pushing the immediate parent of a class to..."

dominikgrewe:
If you fix the typo this is good to go as far as I'm concerned.

j-wilson:
Typo amended.

lib/luaT/luaT.c
j-wilson: I see what you're saying; and, agree. Earlier, I tried switching out `luaT_outerparentname` for `luaT_fullparentname` at `ln 604`; but, torch.test() fails when I did so. The easy fix would be to simply add an additional check at the end of `luaT_getinnerparent`; but, this is a bit of a hack. I'll look into the matter.

j-wilson: Found issue in `luaT_fullparentname`, was boundary case for `tname`s sans parents. Also added comments to `luaT_innerparentname` s.t. it's easier to understand.

dominikgrewe:
Thanks! Are there any other changes you want to make or can we merge now?

A tip for future PRs: It's better to not squash commits while a change is being reviewed. While we eventually want a single squashed commit, it's better to see the incremental changes during the review process.

j-wilson:
Don't have any additional changes in mind. Good tip, will keep it in mind. Please let me know if you have any further input.

dominikgrewe:
Looks good to me now. Thanks!

PR #383, opened 2015-09-30
Exposing LAPACK function potri

j-wilson:
Patch finishes partial implementation of LAPACK function potri and adds test/doc. Also introduces minor (grammatical) changes to test/doc for potrs/potrf.


soumith:
this looks good to me. cc: @koraykv

PR #386, opened 2015-10-02
Doc: remove line saying that 4Dx2D tensor multiplication is supported.

davidsaxton:
The doc is confusing - currently TensorOperator only supports 1Dx1D, 2Dx1D,
and 2Dx2D tensor multiplication (see torch_TensorOperator_(__mul__) in
TensorOperator.c).


dominikgrewe:
Thanks!

PR #390, opened 2015-10-04
Define float/double math functions in generator

szagoruyko:
so that generated functions can use corresponding functions. Patch is coming from [ztorch/lib/THZ/THZGenerateAllTypes.h](https://github.com/facebook/ztorch/blob/master/lib/THZ/THZGenerateAllTypes.h)

fix for nn functions is coming

cc @fmassa

### DIFF
--- a/lib/TH/THGenerateFloatTypes.h
+++ b/lib/TH/THGenerateFloatTypes.h
@@ -2,6 +2,24 @@
 #error "You must define TH_GENERIC_FILE before including THGenerateAllTypes.h"
 #endif
 
+#define TH_ABS fabsf
+#define TH_ACOS acosf
+#define TH_ACOSH acoshf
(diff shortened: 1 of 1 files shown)


soumith:
@szagoruyko the definitions are in the wrong place. (see where the undef for floats are for example, they come right after the defines). can you re-check the PR.

szagoruyko:
oops, I have to be more careful. fixed

fmassa:
Shouldn't you also modify [THTensorMath.c](https://github.com/torch/torch7/blob/master/lib/TH/generic/THTensorMath.c) to reflect those changes ? Like in https://github.com/torch/torch7/blob/master/lib/TH/generic/THTensorMath.c#L1843 ? More generally, change all calls to `fabs` etc to be `THAbs` etc ?

PR #393, opened 2015-10-04
static libraries no longer build by default => builds faster

hughperkins:
static libraries no longer built by default
=> halves the build time

### DIFF
--- a/cmake/TorchPackage.cmake
+++ b/cmake/TorchPackage.cmake
@@ -1,14 +1,15 @@
 # -*- cmake -*-
 
 MACRO(ADD_TORCH_PACKAGE package src luasrc)
-
   INCLUDE_DIRECTORIES(${CMAKE_CURRENT_SOURCE_DIR})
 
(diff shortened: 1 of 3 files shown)


soumith:
There's a way to actually make the static libs reuse the already compiled objects (when compiling dynamic libs). I'll try to do that.

hughperkins:
Ok, sounds good :-)  But... that will likely be linux-specific, thereby complicating the Windows builds that are being worked on?  (Just to be clear, I'm not working on Windows builds, but I have seen people working on them).

hughperkins:
Hi. What concerns do you have with this merge?  Is there something I could modify in it somehow, to avoid causing backwards-incompatible changes, that I assume you are somewhat wary of?

soumith:
I wanted to do something more refined, where a static build is just a relinking of the objects. I didn't get the time to do it.

hughperkins:
Thank you Soumith :-)

PR #394, opened 2015-10-04
parallel build => builds faster

hughperkins:
parallel build => builds faster

### DIFF
--- a/rocks/torch-scm-1.rockspec
+++ b/rocks/torch-scm-1.rockspec
@@ -22,7 +22,15 @@ dependencies = {
 build = {
    type = "command",
    build_command = [[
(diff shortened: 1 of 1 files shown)


hughperkins:
(fixed minor buggette in positioning of `-j` option)

soumith:
this needs to have a Windows fallback, like here:
https://github.com/torch/cutorch/blob/master/rocks/cutorch-scm-1.rockspec

hughperkins:
Ah, thats interesting, considering the comment I just made in the 'static libraries' PR :-D  Ok, will take a look.

hughperkins:
(Pushed update, which removes the parallel code, on Windows)

soumith:
thanks Hugh!

hughperkins:
Thanks! :-)

Issue #323, opened 2015-08-06
Use SSE from THVector


fmassa:
I've noticed that we have SSE instructions for some basic operations in https://github.com/torch/torch7/blob/master/lib/TH/THVector.h .

Is there a reason why `THVector` is not used in `TH` math operations (like `add`, `mul`, etc), at least for the contiguous case, in addition to the `omp` directives ?

soumith:
I think it is mostly because no one spent time doing it. THVector stuff only applies to 1D (contiguous) case usually (you can extend to 2D if it is contiguous for example)

fmassa:
I don't know if it would bring any improvements actually.

I played a bit with it in https://github.com/fmassa/torch-nn/tree/master/sse_optim for the `torch.mul` function using `THDoubleVector_scale` function and the micro-benchmarks I did didn't show any significant improvements over `OMP` (but maybe I did something wrong ?).
Also, the `-O3` flag seems to remove all the benefits of using `THVector`.

But then, those micro-bechmarks are almost always flawed :)

soumith:
sse and openmp are independent optimizations. one can do OpenMP + SSE for example.

fmassa:
@soumith I did exactly that in my previous link, maybe I should have chosen better names.
In my small tests, `parsseMul` stands for OpenMP + SSE https://github.com/fmassa/torch-nn/blob/master/sse_optim/sse.c#L38 , and there was not much gain in doing both compared to doing only OpenMP. But maybe I did something wrong ?

andreaskoepf:
General questing regarding use of OpenMP: Lots of code uses the [TH_TENSOR_APPLY](https://github.com/torch/torch7/blob/e271689b897780b4e4042641a20f4885931fa185/lib/TH/THTensorApply.h) methods but they seem not to have any 'open-mp over threshold path' as for example [THTensorMath.c](https://github.com/torch/torch7/blob/086ec2a760a15a5165b488e25899f3ee70e2ec6e/lib/TH/generic/THTensorMath.c) has. Most of the TH_TENSOR_APPLY using code I saw so far was effectively side-effect free and could immediately use slightly modified 'parallel for' TH_TENSOR_APPLY methods.

dominikgrewe:
I guess there are at least to considerations for using OpenMP:

[...]

PR #395, opened 2015-10-04
[POC] Adds TH_TENSOR_APPLY2_PARALLEL

fmassa:
Following the discussion in https://github.com/torch/torch7/issues/323, I've tried to add a macro `TH_TENSOR_APPLY2_PARALLEL`, which uses omp if both tensors are contiguous. For the moment, I haven't set a threshold to use omp or not.

As a proof of concept, I added it to the unary operations implemented by `LAB_IMPLEMENT_BASIC_FUNCTION` (like `abs`, `tan`, etc).

Any thoughts ?
cc @dominikgrewe

### DIFF
--- a/lib/TH/THTensorApply.h
+++ b/lib/TH/THTensorApply.h
@@ -425,4 +425,167 @@
   THFree(TENSOR##_counter); \
 }
 
+#define TH_TENSOR_APPLY2_PARALLEL(TYPE1, TENSOR1, TYPE2, TENSOR2, CODE) \
+{ \
+  long *TENSOR1##_counter = NULL; \
+  long TENSOR1##_stride = 0, TENSOR1##_size = 0, TENSOR1##_dim = 0, TENSOR1##_i, TENSOR1##_n; \
+  int TENSOR1##_isContiguous = 1; \
+  long *TENSOR2##_counter = NULL; \
+  long TENSOR2##_stride = 0, TENSOR2##_size = 0, TENSOR2##_dim = 0, TENSOR2##_i, TENSOR2##_n; \
+  int TENSOR2##_isContiguous = 1; \
+  int TH_TENSOR_APPLY_hasFinished = 0; \
+\
+  TENSOR1##_n = (TENSOR1->nDimension ? 1 : 0); \
+  for(TENSOR1##_i = 0; TENSOR1##_i < TENSOR1->nDimension; TENSOR1##_i++) \
+    TENSOR1##_n *= TENSOR1->size[TENSOR1##_i]; \
+\
+  TENSOR2##_n = (TENSOR2->nDimension ? 1 : 0); \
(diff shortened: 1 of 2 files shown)


fmassa:
Just a quick heads up on this PR.
A simple benchmark of the `log` function on a 2 cores/4 threads machine gave an speed up of 2.5 times compared to the non-threaded version, for a 2e8 elements contiguous double vector.
This could be specially useful in pointwise operations in `nn`, as for now the parallelism in some of the transfer functions has been removed https://github.com/torch/nn/pull/403

lib/TH/THTensorApply.h
```diff
@@ -425,4 +425,167 @@
...
+      { \
+        if(TENSOR2->stride[TENSOR2##_dim] == TENSOR2##_size) \
+          TENSOR2##_size *= TENSOR2->size[TENSOR2##_dim]; \
+        else \
+        { \
+          TENSOR2##_isContiguous = 0; \
+          break; \
+        } \
+      } \
+    } \
```
dominikgrewe: Feels like there's a lot of code duplication here. Can't we just call `THTensor_isContiguous`? Same for counting the number of elements. We'd need the tensor type as parameter to the macro (in addition to the value type), but that can be easily added.

dominikgrewe:
Sorry for the delay in getting back. It looks good to me in principle, but we should try to reduce the amount of code duplication.

fmassa:
Hi Dominik,
Thanks for the comments.
I agree that there are lots of code duplication. It actually comes from the original `TH_TENSOR_APPLY2` macro.
As I wanted to make this apply function as similar as possible to the non-omp one, I copied/pasted the `TH_TENSOR_APPLY2` macro and added few lines to make it work with omp, the main difference is [here](https://github.com/fmassa/torch7/blob/omp_apply/lib/TH/THTensorApply.h#L502-L518).
When I found the need of the `is_Contiguous` function, I realised it was already present in the original macro (but not explicitly set up, see [here](https://github.com/torch/torch7/blob/master/lib/TH/THTensorApply.h#L270) for example), so I just added the variable [isContiguous](https://github.com/fmassa/torch7/blob/omp_apply/lib/TH/THTensorApply.h#L432) in this [part](https://github.com/fmassa/torch7/blob/omp_apply/lib/TH/THTensorApply.h#L468).

I'll replace those repetitive parts by their respective function calls.

fmassa:
Hi @dominikgrewe ,

I was thinking again about this again, and I'm not sure we would like to have a different interface than the one in `TH_TENSOR_APPLY2`.
My initial goal with this PR was to modify `TH_TENSOR_APPLY2`, but then I realised that there are some use cases where this parallelisation woudn't work (because of reductions or breaks in the middle of the for loop). I thus decided to create a new macro (following your idea in #323 ), but keeping the interface the same to allow easily replacing the `TH_TENSOR_APPLY2` by this new one, when applicable (including in `nn`).

I agree though that there are lots of code duplication, making the compilation time quite long.

What do you think ? @soumith @andresy @koraykv @dominikgrewe

PR #404, opened 2015-10-14
added referenced option in torch.load() and torch.save()

andresy:
also improved documentation for File:referenced()


lvdmaaten:
Looks great!

soumith:
looks great!

PR #408, opened 2015-10-16
Added indexAdd

ajabri:
Added function 'indexAccum' that is similar to 'indexCopy' but instead of setting each value by index in dest, it adds the value to the existing value at the index in dest.

### DIFF
--- a/doc/tensor.md
+++ b/doc/tensor.md
@@ -122,7 +122,7 @@ FloatTensor -- contains floats
 DoubleTensor -- contains doubles
 ```
 
-Most numeric operations are implemented _only_ for `FloatTensor` and `DoubleTensor`. 
+Most numeric operations are implemented _only_ for `FloatTensor` and `DoubleTensor`.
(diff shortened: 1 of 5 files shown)


soumith:
could you add documentation in ./doc/tensor.md

andresy:
shouldn't it be indexAdd? We do not have any Accum elsewhere...

lvdmaaten:
Hmmm yeah `indexAdd` makes sense I suppose. As long as it is clear in the documentation that the `index` LongTensor may contain duplicate values, in which case the corresponding slices are indeed accumulated.

soumith:
thanks Allan!

ajabri:
My pleasure :)

On Mon, Oct 19, 2015 at 3:37 PM, Soumith Chintala <notifications@github.com>
wrote:

> thanks Allan!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/torch/torch7/pull/408#issuecomment-149322422>.
>



-- 
Thanks
Allan

PR #409, opened 2015-10-16
introduced luaT_newlocalmetatable.

andresy:
allows class creation in a local namespace, while being backward-compatible
with luaT_newmetatable.


soumith:
nice! :)

PR #417, opened 2015-10-21
Adding doc for torch.sign()

srsriram:
Also fixed some formatting errors like trailing whitespace and tabs.


soumith:
thank you!

PR #424, opened 2015-10-22
Remove the leading newline from Storage.__tostring__.

timharley:
`tostring()` on a Storage prepends an unnecessary newline character (`\n`):
```
t7> print(tostring(torch.ones(1)))
 1
[torch.DoubleTensor of size 1]

t7> print(tostring(torch.ones(1):storage()))
     <-- removing this \n
 1
[torch.DoubleStorage of size 1]

t7>
```

This pull request removes this newline, to make it consistent with `Tensor.__tostring__`. 

Each string also includes an unnecessary trailing newline, but I imagine removing this will annoy a lot of people.

### DIFF
--- a/Tensor.lua
+++ b/Tensor.lua
@@ -80,7 +80,7 @@ local function Storage__printformat(self)
 end
 
 function Storage.__tostring__(self)
-   local strt = {'\n'}
+   local strt = {}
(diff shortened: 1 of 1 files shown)


soumith:
Thanks.

PR #432, opened 2015-10-26
torch.cat checks that table is not empty.

dominikgrewe:


dominikgrewe:
I'll create the equivalent cutorch patch if everyone's happy with this solution.

soumith:
Thanks dominik!

PR #433, opened 2015-10-27
Fixed large data storing and restoring 

anastasiuspernat:
When tensors size starts to exceed around 4GB the type long can’t hold this size anymore and torch.save/torch.load act unpredictably (due to the underlying c functions - fwrite etc). Reproduced when using convolutional neural networks for big image processing.


soumith:
is everyone okay with this cc: @andresy @dominikgrewe

soumith:
an equivalent patch has to be submitted to cutorch

PR #434, opened 2015-10-27
Use proper precision for accumulators in multinomial

### LINKED ISSUE #418: torch.multinomial precision issues
torch multinomial has precision issues. When the probabilities are given as float, then torch.multinomial(res, p, n, true) does not sample rare events.

The following example should result in ~100,000 samples within 50,001-100,000 range.
With p stored in double, it does. With p stored in float it result in 0 sample within that range.

do
  p = torch.FloatTensor(100*1000):fill(1)
  p:narrow(1, 50*1000 + 1, 50*1000):fill(1e-3)
  p:div(p:sum())
  N = 100*1000

  n = 0
  c = torch.LongTensor(p:nElement()):zero()
  c_ptr = c:data() - 1
  tmp = torch.LongTensor()
  for i = 1, 1000 do
    p.multinomial(tmp, p, N, true);
    n = n + N
    tmp:apply(function(i) c_ptr[i] = c_ptr[i] + 1 end)
  end

  actual = c:narrow(1, 50001, 50000):sum()
  expected = n*p:narrow(1, 50001, 50000):sum()
  print(actual, expected)
end

albanD:
This fixes #418 by setting the accumulators to the proper type as suggested by @andresy 

The following code is now working as expected with `FloatTensor`s

```lua
do
p = torch.FloatTensor(100000):fill(1)
p:narrow(1, 50000 + 1, 50000):fill(1e-3)
p:div(p:sum())
N = 1001000

n = 0
c = torch.LongTensor(p:nElement()):zero()
c_ptr = c:data() - 1
tmp = torch.LongTensor()
for i = 1, 100 do
p.multinomial(tmp, p, N, true);
n = n + N
tmp:apply(function(i) c_ptr[i] = c_ptr[i] + 1 end)
end

actual = c:narrow(1, 50001, 50000):sum()
expected = n*p:narrow(1, 50001, 50000):sum()
print(actual, expected)
end
```

### DIFF
--- a/lib/TH/generic/THTensorRandom.c
+++ b/lib/TH/generic/THTensorRandom.c
@@ -93,7 +93,7 @@ void THTensor_(multinomial)(THLongTensor *self, THGenerator *_generator, THTenso
   for (i=0; i<n_dist; i++)
   {
     /* Get normalized cumulative distribution from prob distribution */
-    real sum = 0;
+    accreal sum = 0;
     for (j=0; j<n_categories; j++)
     {
       sum += THStorage_(get)( \
@@ -160,7 +160,7 @@ void THTensor_(multinomial)(THLongTensor *self, THGenerator *_generator, THTenso
         /* update cumulative distribution so that sample cannot be drawn again */
         real diff;
         real new_val = 0;
-        real sum;
+        accreal sum;
 
         if (sample_idx != 0)
         {


andresy:
Thanks for fixing this :+1:

DavidGrangier:
Thanks!

nicholas-leonard:
+1

nkoumchatzky:
Hey gents! I believe that patch is buggy.
If the sum of probs is a double but the cumulative distribution array is a float array, then you can end up with the last element of the cumulative distribution being < 1 after dividing each of these elements by the sum (double to float conversion followed by float by double division). Here's a simple C main that simulates why this patch can break torch.multinomial in some cases:
```C
#include <stdio.h>

int main()
{
    printf("TEST MULTINOMIAL\n");
    double maxUniform32 = 4294967295L * (1.0/4294967296.0);
    // Take a number bigger than the largest float32 consecutive integer and
    // assume it's the multinomial accreal sum of unnormalized probabilities.
    double sum = 50777218; 
    // Cast it to float as is the case when copying sum to the last value cum_dist (last value of the Tensor)
    float fsample = 50777218; 
    // Print rounded float (the new last element) -- should be 50777216.000000000000000
    printf("fsample: %15.15f\n",fsample);
    // Divide rounded float by exact sum
    fsample /= sum;
    printf("accreal sum of probs: %15.15f\n",sum);
    printf("Last element of the cumulative distribution array: %15.15f\n",fsample);
    printf("Maximum uniform sample in Torch (32 bits): %15.15f\n",maxUniform32);
    printf("Test if last element of cumulative dist array is lower than a possible uniform sample...\n");
    if(maxUniform32 > fsample)
    {
        printf("ERROR: The binary search will yield an out of bounds index!!!!!!!!!");
    }
    return 0;
}
```

So either we can change cum_dist to a double tensor, either we can revert it. The former seems to suit everyone!

Issue #166, opened 2015-03-04
LongStorage seg fault


jonathantompson:
We probably shouldn't seg fault if a long storage is indexed...

>> x = torch.LongStorage(5,2)
>> print(x)

soumith:
i actually dont know why it segfaults, what have you found when hunting the code? LongStorage(5,2) is not a valid constructor right...

deltheil:
The seg fault occurs [here](https://github.com/torch/torch7/blob/7fd571f/lib/TH/generic/THStorage.c#L168).

Since the bounds are OK this is because the `data` pointer is invalid. Indeed looking at [the constructor](https://github.com/torch/torch7/blob/7fd571f/generic/Storage.c#L32-L38) we can see that `torch.LongStorage(5, 2)` has for effect to create a storage with a data pointer set to the address of the 2nd arg:

```C
/* (x) the 2nd arg is not the address of a pre-allocated memory zone! */
real *ptr = (real *)luaL_optlong(L, 2, 0); /* (x) */
storage = THStorage_(newWithData)(ptr, size);
```

Indeed this has been introduced by commit e5342e8:

> Experimental feature: creating Storages from raw pointers, in Lua.

So we can say that `torch.LongStorage(5,2)` is a misuse here.

jonathantompson:
Right, it's definitely a misuse.  It accidentally came about because I
absentmindedly typed LongStorage rather than LongTensor... I was surprised
that we got a hard segfault, and this is something we probably should fix
(assert somewhere).

soumith:
wow, cedric thanks for the hunting, forgot that it even existed.

First step is to document it.

@jonathantompson how do we suppose we do that? We have to trust the user creating the Storage in this case, as we cant verify if a pointer is valid. There is no check we can do without the check triggering a segfault afaik

deltheil:
If the second parameter was explicitly a `cdata` object (instead of a number that holds the pointer address) don't you think it would be less error-prone? It would not avoid misuse (passing an irrelevant `cdata`, like wrong type or size) but it would be a bit more explicit.

What kind of uses do we have apart from doing something like:

```Lua
require 'torch'

local t = torch.FloatTensor(3):fill(0)
local s = torch.FloatStorage(3, tonumber(torch.data(t, true))) -- true=asnumber
```

Or:

```Lua
require 'torch'
local ffi = require 'ffi'

[...]

PR #436, opened 2015-10-27
Tensor and Storage doc update

albanD:
* Add Storage creation from pointer (clarify this error #166 )
* Clarify creation of LongTensor from LongStorage
* Clarify use of torch.data asnumber option


soumith:
finally this doc is fixed. thanks a lot :)

PR #439, opened 2015-10-29
small corner-case check for prodall

soumith:
In trunk, 

torch.Tensor({}):prod() == 1

returns true.

Fixing that small corner-case with an error instead of a surprise.

### DIFF
--- a/lib/TH/generic/THTensorMath.c
+++ b/lib/TH/generic/THTensorMath.c
@@ -405,6 +405,9 @@ accreal THTensor_(sumall)(THTensor *tensor)
 
 accreal THTensor_(prodall)(THTensor *tensor)
 {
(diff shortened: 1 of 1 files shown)


soumith:
bug pointed out by @0wu

0wu:
A side note, I checked python ```numpy.array([]).prod()```, matlab/octave ```prod([])```, they all return 1.0
I guess people see the answer as undefined and then simply choose a convenient default number for it.

dominikgrewe:
I think `prod` returns 1 for empty tensors because it's the identity element wrt multiplication. Just like `sum` returns 0.

I'm in favour of keeping it as it is.

soumith:
ok, i'll just keep it as is then.

Issue #441, opened 2015-11-02
Tensor constructor for IntTensor, etc


eulerreich:
When I do

```lua
local a = torch.IntTensor(3, 4)
print(torch.Tensor(a))
```
I get an error

```lua
bad argument #1 to '?' (expecting number or Tensor or Storage at /tmp/luarocks_torch-scm-1-2006/torch7/generic/Tensor.c:1092)
stack traceback:
	[C]: at 0x7f95f6bef560
	[C]: in function 'Tensor'
	[string "local a = torch.IntTensor(3, 4)..."]:2: in main chunk
	[C]: in function 'xpcall'
	/home/greg/torch/install/share/lua/5.1/itorch/main.lua:179: in function </home/greg/torch/install/share/lua/5.1/itorch/main.lua:143>
	/home/greg/torch/install/share/lua/5.1/lzmq/poller.lua:75: in function 'poll'
	/home/greg/torch/install/share/lua/5.1/lzmq/impl/loop.lua:307: in function 'poll'
	/home/greg/torch/install/share/lua/5.1/lzmq/impl/loop.lua:325: in function 'sleep_ex'
	/home/greg/torch/install/share/lua/5.1/lzmq/impl/loop.lua:370: in function 'start'
	/home/greg/torch/install/share/lua/5.1/itorch/main.lua:350: in main chunk
	[C]: in function 'require'
	[string "arg={'/run/user/1000/jupyter/kernel-3ec1dfcb-..."]:1: in main chunk
```

Same thing with ByteTensor, etc

albanD:
I am not sure what you are trying to do

* If you want to create a clone of `a` then you should use `torch.IntTensor(a)` because by default `torch.Tensor` is a `torch.DoubleTensor`
* If you want to create a `torch.Tensor` of size (3, 4) then two things:
    * `torch.IntTensor(3, 4)` will create a Tensor of size (3, 4) with unitialized memory, if you want it to contains the number 3 and 4 you should do `torch.IntTensor({3, 4})`
    * To construct a tensor from a tensor containing its size, the second one must be a `torch.LongStorage` see the doc [here](https://github.com/torch/torch7/blob/master/doc/tensor.md#torchtensorsizes-strides)

So in the second case, the code will be:
```lua
local a = torch.LongStorage({3, 4})
print(torch.Tensor(a))
```
And if your script gives you an `IntTensor` and you want to create a `torch.Tensor` from it:
```lua
local a = torch.IntTensor({3, 4})
[...]

PR #442, opened 2015-11-02
Improve Tensor creation error messages

albanD:
Following #441, this improves the error message containing Tensors by specifying the expected type.

I know that [these](https://github.com/torch/torch7/blob/5921fddc22a7581b11e4fe44fa80ca9239d6b7d4/generic/Tensor.c#L1114-L1119) messages may not be accurate sometimes: `LongStorage`s could be accepted when coming from the `new` function but not when coming from the `set` function. But due to the current state of `torch_Tensor_(set)`, I cannot be more precise. I will open a PR to fix `torch_Tensor_(set)` (and the corresponding error messages) asap.

### DIFF
--- a/generic/Tensor.c
+++ b/generic/Tensor.c
@@ -405,7 +405,7 @@ static int torch_Tensor_(indexSelect)(lua_State *L)
   }
   else
   {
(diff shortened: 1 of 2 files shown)


soumith:
an awesome fix for usability. thanks a lot.

PR #443, opened 2015-11-03
fix torch_Tensor_(set) arguments logic

albanD:
This prevents signatures like `torch.Tensor(sz1 [,sz2 [,sz3 [,sz4]]]])` and `torch.Tensor(sizes, [strides])` from being used when working with `tensor:set()`. The sample below show these cases.

```lua
local floattensor = torch.FloatTensor(2,3)
local longstorage = torch.LongStorage(6)
floattensor:set(longstorage) -- will now fail with a proper error message
floattensor:set(2) -- will now fail with a proper error message
floattensor:set() -- will now fail with a proper error message
```

### DIFF
--- a/generic/Tensor.c
+++ b/generic/Tensor.c
@@ -2,7 +2,7 @@
 #define TH_GENERIC_FILE "generic/Tensor.c"
 #else
 
-static void torch_Tensor_(c_readTensorStorageSizeStride)(lua_State *L, int index, int allowNone, int allowTensor, int allowStorage, int allowStride,
+static void torch_Tensor_(c_readTensorStorageSizeStride)(lua_State *L, int index, int allowNoStorage,
                                                          THStorage **storage_, long *storageOffset_, THLongStorage **size_, THLongStorage **stride_);
 
 static void torch_Tensor_(c_readSizeStride)(lua_State *L, int index, int allowStride, THLongStorage **size_, THLongStorage **stride_);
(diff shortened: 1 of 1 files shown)


generic/Tensor.c
```diff
@@ -2,7 +2,7 @@
 #define TH_GENERIC_FILE "generic/Tensor.c"
 #else
 
-static void torch_Tensor_(c_readTensorStorageSizeStride)(lua_State *L, int index, int allowNone, int allowTensor, int allowStorage, int allowStride,
```
soumith: any reason why you removed the allowTensor and allowStride options...? is it because they aren't used anywhere in the library anymore?

andresy:
hmm.. not sure where you want to go there. set() reflects what you can do with the constructor.

albanD:
@andresy ok my bad then,
My patch was to allow you to do what's in the [Tensor doc](https://github.com/torch/torch7/blob/08dccaaa692dbaf2a3fcf2d0bed3045c582ea664/doc/tensor.md#self-settensor) only.
You think it would be helpful to update the doc to add the missing use cases? (corresponding to these 2 case [here](https://github.com/torch/torch7/blob/08dccaaa692dbaf2a3fcf2d0bed3045c582ea664/doc/tensor.md#torchtensorsz1-sz2-sz3-sz4) and [here](https://github.com/torch/torch7/blob/08dccaaa692dbaf2a3fcf2d0bed3045c582ea664/doc/tensor.md#torchtensorsizes-strides))

generic/Tensor.c
albanD: This PR won't be merged but
yes the only two places where I found this function to be called are the new() and set() function where all the arguments are true. I guessed they are legacy options that are not used anymore.

soumith:
Alban, fixing the docs would work for me.

PR #447, opened 2015-11-06
tensor:free() should only free underlying storage.

clementfarabet:
In Lua, explicitly calling `:free()` on a Tensor is currently buggy - it actually frees the THTensor struct itself  in C, making any future reference to it invalid (leading to segfaults). This patch makes `:free()` only act on the referenced storage, but preserves the THTensor struct intact (and also maintains the ref counter to 1) — the result: the user can free a tensor-referenced storage from Lua, but the THTensor struct itself will always be handled by the GC.

### DIFF
--- a/doc/tensor.md
+++ b/doc/tensor.md
@@ -1570,7 +1570,7 @@ The source `tensor` should have at least as many elements as the number of 1s in
 ```lua
 x = torch.range(1,4):double():resize(2,2)
 > x
- 1  2  
+ 1  2
  3  4
 [torch.DoubleTensor of dimension 2x4]
 
@@ -1959,7 +1959,7 @@ y = torch.Tensor(2,2)
 Returns a tensor where dimensions `dim1` and `dim2` have been swapped. For 2D tensors,
 the convenience method of [t()](#torch.Tensor.t) is available.
 ```lua
-x = torch.Tensor(3,4):zero()                  
+x = torch.Tensor(3,4):zero()
 x:select(2,3):fill(7) -- fill column 3 with 7
 > x
  0  0  7  0
@@ -2392,3 +2392,13 @@ Increment the reference counter of the tensor.
 
 Decrement the reference counter of the tensor. Free the tensor if the
 counter is at 0.
+
+<a name="torch.Tensor.empty"></a>
+### empty() ###
+
+Empty the tensor, i.e. free the underlying storage and resize the tensor
+to nil. After this call, the tensor is in a state equivalent to
+`tensor = torch.Tensor()`.
+
+If the underlying storage is only referenced by this tensor, then
+the memory referenced by the storage is released immediatly.
--- a/generic/Tensor.c
+++ b/generic/Tensor.c
@@ -1004,6 +1004,28 @@ static int torch_Tensor_(free)(lua_State *L)
   return 0;
 }
 
+static int torch_Tensor_(empty)(lua_State *L)
+{
+  THTensor *tensor = luaT_checkudata(L, 1, torch_Tensor);
+
+  THLongStorage *size = THLongStorage_newWithSize(8);
+  THLongStorage *stride = THLongStorage_newWithSize(8);
+  THLongStorage_fill(size, -1);
+  THLongStorage_fill(stride, -1);
+
+  THTensor_(resize)(tensor, size, stride);
+
+  THLongStorage_free(size);
+  THLongStorage_free(stride);
+
+  if(tensor->storage) {
+    THStorage_(free)(tensor->storage);
+    tensor->storage = NULL;
+  }
+
+  return 0;
+}
+
 /* helpful functions */
 static void torch_Tensor_(c_readSizeStride)(lua_State *L, int index, int allowStride, THLongStorage **size_, THLongStorage **stride_)
 {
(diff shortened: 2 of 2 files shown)


andresy:
Hmm... no, i think it works properly. THTensor_free() is indeed called, but note that this function will actually free the tensor memory only if the refcount gets to zero.

clementfarabet:
Put this in a script:

```lua
for i = 1,100000 do
   local t = torch.Tensor(10)
   t:free()
end
```
... you'll get segfaults, sometimes. 

The issue is that in this case, the user call to `:free()` gets the refcount to 0 and triggers `THFree(tensor)`.  So once the GC gets to it, it's potentially looking at garbage.

soumith:
Clement, afaik this is completely expected. Free is not a user function
that has no side effects, it's a advanced function when you really need
it....

andresy:
i exposed free() and retain() in lua only for very specific applications (in fact: threads and tds)

retain() and free() calls should be balanced.

your code does an extra free() compared to the number of retain() which have been done (a retain is done at tensor creation() and the __gc will call the corresponding free()), so yes, it will segfault.
doing an additional t:retain() will work.

note that changing the free() code will break the current threads/tds stuff.

clementfarabet:
Well but it's literally unusable – in its current form it has no utility (advanced or not), since it's wrong (it will segfault, even in the most restricted use case. The patch I propose has restrictions – it's not entirely thread safe; but at least it will not segfault.

clementfarabet:
Ok – I see, so it's not what I was looking for.

I should rename my version: what I'm looking for is the ability to free the underlying storage, when calling `:free()`, to avoid having to rely on the GC to free up memory.

clementfarabet:
Can I expose another function called `freeStorage()` ?

clementfarabet:
OK – I modified the pull request. Forget about my comments above, this is now a separate function called `freeStorage()`. You can call it to instantly free a tensor's underlying storage.

soumith:
@clementfarabet just to make sure you know, there's a mode called "torch.setheaptracking(true)" which tracks memory usage across tensor (Cuda as well), which calls collectgarbage when the storage heap is too large. @adamlerer introduced it a few months ago as an optional argument, we've been using it internally (enabled by default) with full stability and great success. freeStorage seems useful regardless, but what you're trying to achieve seems to also be done with heap-tracking in a much more graceful manner.

soumith:
your PR btw decreases the refcount of the storage, but it does not change the size of the underlying tensor to zero, so you might have to fix that.

dominikgrewe:
I can see why this function may be useful, e.g. to explicitly remove temporary buffers before saving a network to disk. But I don't think the current implementation is safe. If you decrement the ref-count but it's not 0 yet, you will also get seg-faults because the number of references doesn't match the ref-count.

Btw, this kind of code should be in TH.

clementfarabet:
Good points — so I want to avoid collectgarbage() at all costs, the goal is to really free the underlying storage on demand. 

Let me rewrite this a bit – it shouldn't even touch the refcount of the tensor itself, you're right, it should just: (1) free the storage and (2) resize itself to 0.

timharley:
Would this be sufficient for your use-case? It frees the underlying memory for the data, though doesn't free either struct:
```
tensor:resize(0)
tensor:storage():resize(0) -- assume no other Tensors reference this storage.
```

(The retain/free are really useful as is, if one wants to manage refcounts from lua).

clementfarabet:
Does `:storage():resize(0)` actually free the memory for the storage?

I rewrote this PR, now introducing a method called `:empty()`. It resizes the Tensor to no dimensions, and then force frees the underlying storage, and un-references it. If another tensor references that storage the storage will not be freed. How does that look?

soumith:
i think it looks fine. Add some doc, and I'll merge it in.

clementfarabet:
Done

andresy:
hmfff... i go away, and returned too late. i am not sure of the code in there. if it is just about freeing the storage, you can do: tensor:set()

clementfarabet:
It's about releasing the memory instantly – doing tensor:set() will still
rely on the GC to free up the memory.

andresy:
no, it will release the storage pointer instantaneously (if, of course, there are no other pointer on it)

clementfarabet:
Oh, damn, that's true! OK this code is pointless then.

andresy:
yes (no offense)

andresy:
[and tensor:set() will also not create additional memory from C, as you do right now for size/stride]

clementfarabet:
Yeah, I just didn't know that set() could be called without args.

andresy:
in fact the purpose of set() is to mimic the constructor operations, but to avoid playing with the GC (if you have a particular tensor already allocated)

clementfarabet:
Yeah this is exactly what I was looking for, bypassing the GC and controlling alloc/dealloc. 

Reverted here:
https://github.com/torch/torch7/pull/448

clementfarabet:
While we're on the subject, have you guys kept experimenting with the refcounted lua?

soumith:
we haven't looked into refcounted lua, but the heap tracking has given us much peace, without needing collectgarbage() everywhere in the world.

clementfarabet:
Cool. With heap tracking enabled, do you have a guaranty to always free up
memory in time? Also works on CUDA?

soumith:
> Cool. With heap tracking enabled, do you have a guaranty to always free up memory in time?

Yes, within 1MB (https://github.com/torch/torch7/blob/d962e0f4e8a0ea04fcc1c03c88debddfeeac9a3a/lib/TH/THGeneral.c#L106)

More info on the internals: https://github.com/torch/torch7/blob/d962e0f4e8a0ea04fcc1c03c88debddfeeac9a3a/lib/TH/THGeneral.c#L111-L120
https://github.com/torch/cutorch/blob/master/lib/THC/THCGeneral.c#L532-L535

Works with CUDA as well: 
https://github.com/torch/cutorch/blob/master/lib/THC/THCGeneral.c#L497-L536

clementfarabet:
Neat, thanks for the pointers.

PR #448, opened 2015-11-08
No point for empty(), equivalent to set().

clementfarabet:


clementfarabet:
Reverting `:empty()` – as pointed out by @andresy, `:set()` [with no args] is exactly what I was looking for.

Issue #456, opened 2015-11-11
torch.MemoryFile writeObject() is broken


somadivad:
To reproduce:

$ torch
t7> require 'torch'
t7> file = torch.MemoryFile("rw")
t7> file:writeObject({})
/usr/share/lua/5.1/torch/File.lua:109: $ Torch: not enough memory: you tried to reallocate 144GB. Buy new RAM! at /tmp/luarocks_torch-scm-1-1680/torch7/lib/TH/THGeneral.c:251
stack traceback:
	[C]: in function 'writeInt'
	/usr/share/lua/5.1/torch/File.lua:109: in function 'writeObject'
	stdin:1: in main chunk
	[C]: at 0x00405e60

This appears to have been introduced by https://github.com/torch/torch7/pull/433, but it's not clear why.

apaszke:
I guess the problem is that someone replaced `long` with `size_t` everywhere, so it is now always an unsigned type!

It's easy to see that it's a wrong choice even if you take a look just here:
https://github.com/torch/torch7/blob/master/lib/TH/THMemoryFile.c#L372
Setting `posEol` to -1 makes it the biggest representable value and the `if`'s condition from line 383 is always true.

`ssize_t` would be probably a better choice.

Anyway, I reinstalled torch yesterday and can't reproduce the bug you're experiencing :confused:

apaszke:
@soumith Isn't this fixed?

soumith:
indeed. thx,

PR #457, opened 2015-11-11
Fixed unsigned overflows in THFile

apaszke:
Several days ago there was a commit that replaced `long` with `size_t` in THFile, THDiskFile and THMemoryFile. Because it is an unsigned type an overflow happened in several places. This commit fixes it.


dominikgrewe:
Thanks for the quick patch! This fixes #456

PR #459, opened 2015-11-12
Regression test for MemoryFile bug.

dominikgrewe:


dominikgrewe:
@soumith The tests in `test/test_writeObject` are currently not run by Travis. In fact only `torch.test()` is run. Can you fix that?

The tests in `test_writeObject` actually weren't affected by the bug reported in issue #456, I think because of the `binary` call. I therefore added a simple regression test that currently brakes. PR #457 fixes it.

soumith:
thanks Dominik. Enabling the test_writeObject tests now...

PR #463, opened 2015-11-16
Accept 64-bit offsets in THDiskFile_seek

apaszke:
Negative positions should be checked in `torch_File_seek`, because negative numbers will overflow when converted. It prints an error message "position has to be smaller than LONG_MAX", which isn't particularly helpful.

Also, if somebody has 32-bit longs then seeking beyond 2GB will fail. I've fixed that for unix systems, but I'm too inexperienced with Windows development to make it work there too.


soumith:
Thanks Adam.

PR #465, opened 2015-11-17
Detecting LAPACK on the system with LAPACKless OpenBLAS

borisfom:
Done for the installer, but useful in general. Also, my Emacs deletes trailing whitespace on save :)


andresy:
Maybe it is worth checking if the default cmake FindBLAS/FindLAPACK works properly nowadays. We shipped this custom file (now for years!) because early versions of cmake were not robust enough.

borisfom:
Well, I don't think such change is worth it at the moment, especially that you still need to express preference to openblas somehow. The change I propose just does not prevent the script from finding generic lapack in openblas case.

soumith:
Thanks Boris!

PR #467, opened 2015-11-18
Add isSetTo: simple check for shared storage.

dm-jrae:
Returns true iff Tensor is set to argument Tensor. Specifically, this is
true iff tensor shares same storage as argument tensor, with same
storage offset and identical sizes and strides.

What do people think about including such a method?


dominikgrewe:
Looks good to me. @soumith what do you think?

soumith:
looks good to me as well. Thanks @dm-jrae

PR #470, opened 2015-11-23
Added torch.equal function which performs a tensor equality check

fbesse:


timharley:
@soumith @dominikgrewe any thoughts on this?

doc/maths.md
```diff
@@ -506,6 +506,22 @@ Let `x` be a tensor and `n` a number.
 
 In this section, we explain basic mathematical operations for Tensors.
 
+<a name="torch.Tensor.equal"></a>
+### [boolean] equal([tensor1,] tensor2) ###
+
+Returns `true` iff the values of `tensor1` and `tensor2` are exactly the same.
```
timharley: Mention that this is much more efficient than `a:eq(b):all()`, as it avoids allocation of a temporary and can short circuit.

soumith:
Last time such a PR was presented, it got into a philosophical debate on what "equal" was (and whether shape matters in equality, or whether it's just the contents. Or if equal should compare the same tensor reference).
I largely like this PR because it does add usability, but dont know about the others.

soumith:
Here's the previous debate: https://github.com/torch/torch7/pull/119

timharley:
It's easy to check for equality across differing shapes using this function and a couple of `:view()` calls, so I'm inclined to go for this definition of equality over one that ignores shape. @soumith if you think this is sufficient to be merged, @fbesse could update the branch to fix conflicts and address comments?

fbesse:
@timharley Yes I can take care of updating this.

fbesse:
Updated doc and fixed conflicts.

andreaskoepf:
Regarding the name of the function we could discuss  'equal' vs. 'equals'. Regarding compare by ref and compare by value: The __eq metatable function should be reserved for compare-by-reference and an equal(s) function for compare by value (which includes the tensor dimensionality, what @timharley mentioned is right) - exactly as it is proposed by this PR. A bit OT but see what @ericlippert [wrote](http://www.informit.com/articles/article.aspx?p=2425867) about not-so-ideal parts of C#: '#9: Too much equality'.
@fbesse Since Tensor and Storage a closely related it would be a good idea to provide an equals() function for them too.

soumith:
let's see what happens :)

andreaskoepf:
I just noticed that `help(torch.equal)` in th says 'undocumented symbol'. In [doc/maths.md](https://github.com/torch/torch7/blob/master/doc/maths.md) this seems to be the only function that has the Tensor in the anchor-name `<a name="torch.Tensor.equal"></a>`. Was this done deliberately or should the 'Tensor' be removed?

fbesse:
You're right, I've made a PR to fix that: https://github.com/torch/torch7/pull/662

Issue #5, opened 2014-02-04
file serialized and saved as littleendian binary fails to load in bigendian machines, even after explicitly setting encoding


soumith:
an nn model serialized and saved as littleendian binary (with torch.save on x86) fails to deserialize properly in bigendian machines (ARM), even after explicitly setting encoding with f:littleEndianEncoding()

soumith:
fixed via #476

PR #476, opened 2015-11-26
New function longSize for files

mvitez:
This function allows binary files compatibility between 32 and 64 bit
architectures.

The only difference between 32 and 64 bit architectures is the size of the C type long. By specifying the size of long in binary files, binary files can be read back on different architectures, in the same way as bigEndianEncoding() and littleEndianEncoding() allow to exchange files between architectures with same size, but different endianness.


soumith:
oooh nice! I suppose this solves: https://github.com/torch/torch7/issues/5

soumith:
is this backward compatible?

mvitez:
Yes, it solves that issue, as that issue has nothing to do with endianness (ARMs are normally configured to work little endian, even if they can work in both modes).

This function has maximum backward compatibility, as it allows to work in three modes: use 4 byte longs, use 8 byte longs, use system default (default, as it it's now).

soumith:
Thanks a lot Marko. Very clean and nicely done.

mvitez:
Great. Thank you.

Atcold:
Should we include this into `torch.load()` and `torch.save()`, or it is considered an option for advanced users only? For example
```lua
torch.save(filename, object [, format, referenced])
[object] torch.load(filename [, format, referenced])
```
and if `format` can take `string` or `binary`, as usual, plus  `long=0` (standard binary), `long=4` and `long=8`.
I can try to implement it, @soumith, if you're OK.

soumith:
there seem to be no major takers for this. I think we will keep it at the file level for now and not include torch.load / torch.save

Atcold:
Hmm... I'm not sure there are no major takers.
Anyone who switches between 32 and 64 bit OS is going to face this annoying problem.
Let's add a `torch.load32()` and `torch.load64()` which automates the conversion.

mvitez:
:+1:

culurciello:
+1 yes this would make all our lives a lot easier please guys!

culurciello:
+1

Atcold:
Merged here #579.

PR #478, opened 2015-11-29
Error messages when bytecode loading fails

GeorgOstrovski:
This is, for example, relevant, when trying to load bytecode with Luajit 2.1 which was created with Luajit 2.0 - there is an incompatibility and previously the error would occur somewhere below (e.g. when trying to set upvalues) and give a less meaningful error message.


soumith:
Thanks!

PR #488, opened 2015-12-18
Make nBufferWritten signed for writing longs

alvinlschua:
Comparison on line 486 (nBufferWritten > -1) sometime fails when size_t resolves to unsigned int.


soumith:
Thanks for spotting this. Do we need to make the same change in DiskFile?

PR #489, opened 2015-12-18
Adding argmax index as second return value to max/min calls without specified dimension

mevGDM:
This makes a number of changes to both lua and C code to facilitate the second return value of a torch.max(input) call without specified dimension to be a LongTensor(default) that gives the indeces of the argmax of an input tensor. The C interface for the relevant functions(THTensor_(minall)/(maxall)) is changed and therefore it is not backward compatible (e.g. see line 2274 @ lib/TH/generic/THTensorMath.c). We can keep the original functions if requested in which case C interface would be compatible with earlier versions as well.

### DIFF
--- a/lib/TH/generic/THTensorMath.c
+++ b/lib/TH/generic/THTensorMath.c
@@ -358,11 +358,24 @@ accreal THTensor_(dot)(THTensor *tensor, THTensor *src)
   return sum;
 }
 
-real THTensor_(minall)(THTensor *tensor)
-{
-  real theMin;
-  real value;
+void THTensor_(flatIndexToFullIndex)(long ind, long *index_data, long nDim, long *sizes){
+    long total = 1;
+    long i;
+    for (i = 0; i < nDim; i++){
+      total = total * sizes[i];
+    }
+    for (i = 0; i < nDim; i++){
+      total = total / sizes[i];
+      index_data[i] = ind / total;
+      ind = ind - (index_data[i]) * total;
+    }
+}
 
+real THTensor_(minall)(THLongTensor *index, THTensor *tensor)
+{
+  real value;
+  real theMin;
+  long theIndex = 0;
   THArgCheck(tensor->nDimension > 0, 1, "tensor must have one dimension");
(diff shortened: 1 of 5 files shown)


soumith:
@mevGDM looks good to me. But the docs need to be updated to reflect this as well.

mevGDM:
Eliminated the problem related to the use of 'real' type in THTensorMath.c which was causing problems with obtaining the correct indices when the tensor is not of type double or float. With the changes to the documentation earlier, this should be ready for the pull.

lib/TH/generic/THTensorMath.c
```diff
@@ -358,10 +358,27 @@ accreal THTensor_(dot)(THTensor *tensor, THTensor *src)
   return sum;
 }
 
-real THTensor_(minall)(THTensor *tensor)
+void THTensor_(flatIndexToFullIndex)(long ind, long *index_data, long nDim, long *sizes){
+    double total = 1;
```
dominikgrewe: Shouldn't you use `long` instead of `double`? In the first division, `total` is a multiple of `sizes[i]` and in the second division you floor anyway. So I don't think you need double division at all.

mevGDM:
Changed the flatIndexToFullIndex to use longs instead of explicit floor calls and casts. The tests (to test.lua) were also added in the previous commits which checks the correctness of argmin/argmax return values for different tensor types and different tensor sizes.

dominikgrewe:
We also need to update nn, as it calls into minall and maxall: https://github.com/torch/nn/blob/master/generic/MultiLabelMarginCriterion.c#L32

mevGDM:
Here is the pull request for MultiLabelMarginCriterion.c https://github.com/torch/nn/pull/575

dominikgrewe:
Looks good to me. @soumith Can we merge this?

soumith:
Looks good to me. Squash and merge.

soumith:
Could yo please prepare the same for cutorch? right now torch and cutorch versions of this differ in functionality, and that is not good.

PR #491, opened 2015-12-22
File.lua: fix undefined debug function

deltheil:
See #372


soumith:
Thanks for the quick fix.

PR #496, opened 2015-12-29
k-selection implementation: torch.topk

wickedfoo:
This is an implementation of k-selection: returning the `k` highest or lowest values in a selected dimension.

For sorted results, it is functionally equivalent to:

```
local function topKViaSort(t, k, dim, dir)
  local sorted, indices = t:sort(dim, dir)
  return sorted:narrow(dim, 1, k), indices:narrow(dim, 1, k)
end
```

This torch7 CPU implementation I provide here actually just maps to the above (in effect, always sorting the results). It could be improved to use selection to find the kth element and a second scan at some point.

I am adding this as a simple, albeit functional placeholder because I have an efficient, optimized GPU cutorch implementation of this based on radix selection. I need this in torch7 in order to allow the torch.topk() bindings to work. @soumith and @andresy are aware of my cutorch implementation internally; I will provide a cutorch pull request for CudaTensor topk as soon as I fix remaining issues with cutorch sort once and for all.

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -78,7 +78,7 @@ static int torch_NAME(lua_State *L)
   }
   else if(!(tname = torch_istensortype(L, torch_getdefaulttensortype(L))))
     luaL_error(L, "internal error: the default tensor type does not seem to be an actual tensor");
-  
+
   lua_pushstring(L, "NAME");
   lua_rawget(L, -2);
   if(lua_isfunction(L, -1))
@@ -104,7 +104,7 @@ function interface:dispatchregister(name)
    end
    table.insert(txt, '{NULL, NULL}')
    table.insert(txt, '};')
-   table.insert(txt, '')   
+   table.insert(txt, '')
    self.dispatchregistry = {}
 end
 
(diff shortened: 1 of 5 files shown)


nicholas-leonard:
@wickedfoo I like the idea. I have found need for this functionality many times. Note that your commit isn't associated with your github account (needs to be done with ssh). Also, your cat pic is whack.

wickedfoo:
@nicholas-leonard should be associated now, the commit was associated with a different email address than the one on my account.

soumith:
everyone okay with checking this in as is?

fmassa:
I like the idea of the PR.
As mentioned in the description, it can be optimized by using for example the `kthvalue` function, to avoid sorting all the elements, but that can be done later I think.

wickedfoo:
I spent all my time on the GPU version, someone else can do the real CPU implementation :)

soumith:
Thanks Jeff.

wickedfoo:
GPU version is here in case anyone is curious:
https://github.com/torch/cutorch/pull/296

Issue #499, opened 2016-01-11
Stand-alone sigmoid function?


shawnjhenry:
I'm working on some custom modules that make extensive use of the sigmoid function.  Some quick messing around seems to indicate that the implementation in nn.Sigmoid is substantially faster than any possible combination of torch.exp, torch.pow, torch.cdiv and ones tensors.  Obviously I could just use nn.Sigmoid as a function instead of a module, but that seems kind of cumbersome, and it would be much nicer to have a stand-alone torch.sigmoid function, especially since it's such a common use function. Would this be useful to anyone else?

Thanks,
Shawn

fmassa:
This has been addressed in #500 and torch/cutorch#298

PR #500, opened 2016-01-12
Add torch.sigmoid, and sometensor:sigmoid()

hughperkins:
Addresses https://github.com/torch/torch7/issues/499

Adds:
```
torch.sigmoid(somefloat)
torch.sigmoid(sometensor)
sometensor:sigmoid()
```


soumith:
Thanks Hugh!

PR #503, opened 2016-01-14
maths.md: minor wording changes

jiahao:
- Matlab -> MATLAB (This is how the company spells it)
- mathematic -> mathematical
- "sizes must be relevant" -> "sizes must be conformant" (this is more or less the standard term used when defining matrix multiplication)


soumith:
Thank you!

Issue #168, opened 2015-03-09
one over X function


soumith:
A useful function to have would be 1/x of a tensor (in-place and out-of-place).

andresy:
yes, it is on my todo list

AjayTalati:
That would be really cool !!! 

Here's a concrete example of what I guess people want to do more efficiently 

`one_over_Z = torch.cdiv(torch.ones(N) , Z)`

If you can please post [here](https://groups.google.com/forum/#!topic/torch7/R_PA1Cnu6QA) when this is implemented.

Kind regards, Aj

nicholas-leonard:
`a:pow(-1)`?

andresy:
yes, that is what i do for now

hughperkins:
`:inv()`?

hughperkins:
Following Jonathan Tomson's advice, changing to `:cinv()`

soumith:
Merged into trunk!

PR #511, opened 2016-01-20
add :cinv(), which does 1/x

hughperkins:
Add :cinv(), which does 1/x.  Use like `:cinv()`, or `torch.cinv(...)`

Addresses https://github.com/torch/torch7/issues/168


hughperkins:
By the way, I also renamed the c function `sigmoid` to `TH_sigmoid`, to avoid possible conflicts, and made it `static inline`, and remove THMath.c.

jonathantompson:
Suggestion: maybe rename to cinv?  Otherwise people are going to (naturally) think that this calculates a matrix inverse (we defined cmul for the same reason).

hughperkins:
Ah, good point.

hughperkins:
ah, yes, `cinv` is perfect really, ie 'per-element inverse'.

jonathantompson:
Cool.  Thanks for the PR!  You've definitely be on fire lately with all your contributions :-)

lib/TH/generic/THTensorMath.c
```diff
@@ -1995,6 +1995,7 @@ LAB_IMPLEMENT_BASIC_FUNCTION(floor,floor)
 LAB_IMPLEMENT_BASIC_FUNCTION(round,round)
 LAB_IMPLEMENT_BASIC_FUNCTION(abs,fabs)
 LAB_IMPLEMENT_BASIC_FUNCTION(neg,-)
+LAB_IMPLEMENT_BASIC_FUNCTION(inv,TH_inv)
```
fmassa: Just wondering, shouldn't we try to be consistent in here ? [In your cutorch PR](https://github.com/torch/cutorch/pull/306/files#diff-13ee6e0e8f659bdf5f14261f7bee76a5R54), you used `1.0f /`.

hughperkins: Hmmm, yes.  That does make sense.  I'll see if that works.

hughperkins: Addressed

hughperkins:
Travis didnt like that.  Random chance?  Should change back to using `TH_cinv`, instead of `1.0 / ` in the macro?  I'll try pushing some tiny change, nad see what happens.

soumith:
Thanks Hugh!

PR #514, opened 2016-01-23
torch.topk: use quickselect + quicksort

deltheil:
Follow-up of #496

### DIFF
--- a/lib/TH/generic/THTensorMath.c
+++ b/lib/TH/generic/THTensorMath.c
@@ -1754,22 +1754,61 @@ void THTensor_(topk)(THTensor *rt_, THLongTensor *ri_, THTensor *t, long k, int 
   long sliceSize = THTensor_(size)(t, dim);
   THArgCheck(k > 0 && k <= sliceSize, 2, "k not in range for dimension");
 
-  /* Just implement in terms of sort and narrow for now */
-  THTensor* tmpResults = THTensor_(new)();
-  THLongTensor* tmpIndices = THLongTensor_new();
+  THTensor *tmpResults = THTensor_(new)();
+  THTensor_(resize1d)(tmpResults, sliceSize);
+  real *tmp__data = THTensor_(data)(tmpResults);
 
-  THLongStorage* topKSize = THTensor_(newSizeOf)(t);
+  THLongTensor *tmpIndices = THLongTensor_new();
+  THLongTensor_resize1d(tmpIndices, sliceSize);
+  long *tmpi__data = THLongTensor_data(tmpIndices);
+
+  THLongStorage *topKSize = THTensor_(newSizeOf)(t);
   THLongStorage_set(topKSize, dim, k);
   THTensor_(resize)(rt_, topKSize, NULL);
   THLongTensor_resize(ri_, topKSize, NULL);
   THLongStorage_free(topKSize);
 
-  THTensor_(sort)(tmpResults, tmpIndices, t, dim, dir);
-  THTensor_(narrow)(tmpResults, NULL, dim, 0, k);
-  THLongTensor_narrow(tmpIndices, NULL, dim, 0, k);
+  if (dir) {
+    /* k largest elements, descending order (optional: see sorted) */
+    long K = sliceSize - k;
+    TH_TENSOR_DIM_APPLY3(real, t, real, rt_, long, ri_, dim,
+                         long i;
+                         for(i = 0; i < sliceSize; i++)
+                         {
+                           tmp__data[i] = t_data[i*t_stride];
+                           tmpi__data[i] = i;
+                         }
+                         if (K > 0)
+                           THTensor_(quickselect)(tmp__data, tmpi__data, K - 1, sliceSize, 1);
+                         if (sorted)
+                           THTensor_(quicksortdescend)(tmp__data + K, tmpi__data + K, k, 1);
+                         for(i = 0; i < k; i++)
+                         {
+                           rt__data[i*rt__stride] = tmp__data[i + K];
+                           ri__data[i*ri__stride] = tmpi__data[i + K];
+                         })
+  }
+  else {
+    /* k smallest elements, ascending order (optional: see sorted) */
+    TH_TENSOR_DIM_APPLY3(real, t, real, rt_, long, ri_, dim,
+                         long i;
+                         for(i = 0; i < sliceSize; i++)
+                         {
+                           tmp__data[i] = t_data[i*t_stride];
+                           tmpi__data[i] = i;
+                         }
+                         THTensor_(quickselect)(tmp__data, tmpi__data, k - 1, sliceSize, 1);
+                         if (sorted)
+                           THTensor_(quicksortascend)(tmp__data, tmpi__data, k - 1, 1);
+                         for(i = 0; i < k; i++)
+                         {
+                           rt__data[i*rt__stride] = tmp__data[i];
+                           ri__data[i*ri__stride] = tmpi__data[i];
+                         })
+  }
 
-  THTensor_(freeCopyTo)(tmpResults, rt_);
-  THLongTensor_freeCopyTo(tmpIndices, ri_);
+  THTensor_(free)(tmpResults);
+  THLongTensor_free(tmpIndices);
 }
 
 void THTensor_(tril)(THTensor *r_, THTensor *t, long k)
--- a/test/test.lua
+++ b/test/test.lua
@@ -1358,7 +1358,7 @@ function torchtest.topK()
       return sorted:narrow(dim, 1, k), indices:narrow(dim, 1, k)
    end
 
-   local function compareTensors(t, res1, ind1, res2, ind2, msg)
+   local function compareTensors(t, res1, ind1, res2, ind2, dim, msg)
       -- Values should be exactly equivalent
       mytester:assertTensorEq(res1, res2, 0, msg)
 
@@ -1374,10 +1374,10 @@ function torchtest.topK()
    end
 
    local function compare(t, k, dim, dir, msg)
-      local topKVal, topKInd = t:topk(k, dim, dir)
+      local topKVal, topKInd = t:topk(k, dim, dir, true)
       local sortKVal, sortKInd = topKViaSort(t, k, dim, dir)
 
-      compareTensors(t, sortKVal, sortKInd, topKVal, topKInd, msg)
+      compareTensors(t, sortKVal, sortKInd, topKVal, topKInd, dim, msg)
    end
 
    local t = torch.rand(math.random(1, msize),


lib/TH/generic/THTensorMath.c
```diff
@@ -1754,22 +1754,61 @@ void THTensor_(topk)(THTensor *rt_, THLongTensor *ri_, THTensor *t, long k, int
...
+  THLongStorage *topKSize = THTensor_(newSizeOf)(t);
   THLongStorage_set(topKSize, dim, k);
   THTensor_(resize)(rt_, topKSize, NULL);
   THLongTensor_resize(ri_, topKSize, NULL);
   THLongStorage_free(topKSize);
 
-  THTensor_(sort)(tmpResults, tmpIndices, t, dim, dir);
-  THTensor_(narrow)(tmpResults, NULL, dim, 0, k);
-  THLongTensor_narrow(tmpIndices, NULL, dim, 0, k);
+  if (dir) {
```
apaszke: Wouldn't it be simpler to do it with only one macro? Something like this should work:

```C
long K = 0;
TH_TENSOR_DIM_APPLY3(real, t, real, rt_, long, ri_, dim,
                     long i;
                     for(i = 0; i < sliceSize; i++)
                     {
                       tmp__data[i] = t_data[i*t_stride];
                       tmpi__data[i] = i;
                     }

                     if (dir) {
                       /* k largest elements, descending order (optional: see sorted) */
                       K = sliceSize - k;
                       if (K > 0)
                         THTensor_(quickselect)(tmp__data, tmpi__data, K - 1, sliceSize, 1);
                       if (sorted)
                         THTensor_(quicksortdescend)(tmp__data + K, tmpi__data + K, k, 1);
                     } else {
                       /* k smallest elements, ascending order (optional: see sorted) */
                       THTensor_(quickselect)(tmp__data, tmpi__data, k - 1, sliceSize, 1);
                       if (sorted)
                         THTensor_(quicksortascend)(tmp__data, tmpi__data, k - 1, 1);
                     }

                     for(i = 0; i < k; i++)
                     {
                       rt__data[i*rt__stride] = tmp__data[i + K];
                       ri__data[i*ri__stride] = tmpi__data[i + K];
                     })

```

Also is the copy of `sliceSize` necessary, because `quickselect` only works on contiguous Tensors?

deltheil: > Wouldn't it be simpler to do it with only one macro?

That's true that is saves some repetitions and keep things readable. Also it should not be critical perf. wise to move this branch at the macro level (?) Let's see what other thinks and I can compact this out as suggested!

> Also is the copy of `sliceSize` necessary, because `quickselect` only works on contiguous Tensors

Well `quickselect` supports a stride (see last parameter). But there is a single stride parameter, so it must be the same for the data and the indices. Plus and more importantly `quickselect` works by rearranging the data (swaps) so we need a copy here (this is what `kthvalue` does BTW). Can you think about something better?

apaszke: Yeah, I've edited out the `quickselect` note, because I realised that the copy is necessary due to the fact that it rearranges elements. No, I don't think that it's possible to do the selection quickly without copying the elements :confused:

deltheil: The only alternative I first thought about was to do something closer to the former implementation and `torch.sort`, i.e. 1/ copy the whole input tensor at the beginning (`THTensor_(copy)(rt_, t);`), 2/ do each quickselect / quicksort on each slice of this `rt_` tensor (no more copy in the macro) and 3/ do a final `narrow`.

apaszke: Well it's actually more or less the same in terms of the algorithm. The only other idea that I can think of is to construct a heap for each slice and then extract k elements from it. The upside is that you get them sorted for free. I did some search and apparently (as [in here](http://blog.mischel.com/2011/10/25/when-theory-meets-practice/)) it might be faster than `quickselect` for small k (also, this post didn't include the sorting - it was only a one element selection). I think that the only way to check what's better is to benchmark :smiley:

deltheil: Heap-based solutions are definitely alternatives. I thought it was nice going one step further re-using already available blocks (`quickselect` and `quicksort`), plus it sounds like the GPU counterpart is anyway already highly optimized!

apaszke: Indeed, the heap-based code would be slightly longer. Yeah, I've read about the GPU implementation, but I didn't have time to take a look at it. If I'll find a moment maybe I'll try to write a heap based implementation and run a little benchmark.

apaszke:
@deltheil I wrote a basic implementation that uses a heap to do the selection. On a simple benchmark of taking a random 1000x1000 tensor and selecting 5 smallest elements from each row (repeated 100 times) it takes 5.5s vs quickselect's 9.5s. I'm wondering if we should include any special cases for comparison, what do you think?

deltheil:
@apaszke cool! Perhaps it would be interesting to do the comparison for large(r) k values too? Apart from that if you think the heap-based version is superior feel free to open a PR and we can of course discard this one.

apaszke:
I ran a couple of benchmarks and it seems that the heap is faster as long as `k <= n/2`, afterwards it gets slightly slower. Unsurprisingly, quickselect doesn't depend that much on chosen `k`. The numbers above run times are sizes of 2nd tensor dimension.

In the first benchmark I kept the product of `#iter * 2nd dimension == 100000` and this is what I got (first dimension was always 1000):
```
1k                          2k                         5k
heap         quickselect    heap        quickselect    heap         quickselect
===========  ===========    ==========  ===========    ===========  ===========
5:    5.47   5:    9.48     5:    5.54  5:    10.36    5:    5.67   5:    10.54
10:   5.32   10:   9.35     10:   5.15  10:   9.93     10:   5.51   10:   10.64
15:   5.35   15:   9.38     15:   5.11  15:   9.94     15:   5.47   15:   10.50
50:   5.69   50:   9.46     50:   5.37  50:   9.97     50:   5.57   50:   10.63
100:  6.02   100:  9.63     100:  5.65  100:  10.04    100:  5.67   100:  10.56
150:  6.40   150:  9.63     150:  5.82  150:  10.15    200:  5.84   200:  10.56
200:  6.72   200:  9.72     200:  6.05  200:  10.10    500:  6.30   500:  10.59
300:  7.49   300:  9.77     300:  6.46  300:  10.78    800:  6.85   800:  10.66
400:  8.19   400:  9.86     400:  6.84  400:  10.56    1000: 7.69   1000: 10.65
500:  8.89   500:  9.96     500:  7.23  500:  10.23    1200: 7.76   1200: 10.65
600:  9.60   600:  10.13    600:  7.57  600:  10.27    1500: 8.20   1500: 10.78
700:  10.22  700:  10.16    700:  7.97  700:  10.30    1750: 8.96   1750: 10.77
800:  11.49  800:  10.36    800:  8.36  800:  10.31    2000: 8.90   2000: 10.82
900:  12.94  900:  10.25    900:  8.70  900:  10.42    2500: 9.69   2500: 10.89
1000: 12.90  1000: 10.43    1000: 9.12  1000: 10.43    3000: 10.67  3000: 10.94
                            1200: 9.75  1200: 10.44    3500: 11.41  3500: 11.00
                            1500: 10.74 1500: 10.48    4000: 12.14  4000: 11.15
                            1750: 11.52 1750: 10.45    4500: 12.70  4500: 11.07
                            2000: 12.18 2000: 10.52    5000: 13.27  5000: 11.12
```

In the second benchmark I have varied only the 2nd tensor dimension (first was always equal to 1000 and I repeated the procedure 20 times):

```
1k                        2k                        10k
heap        quick         heap        quick         heap          quickselect
==========  ==========    ==========  ==========    ============  ============
5:    1.06  5:    1.90    5:    2.28  5:    4.09    5:     10.14  5:     22.07
10:   1.08  10:   1.92    10:   2.27  10:   3.98    10:    10.03  10:    22.05
15:   1.11  15:   1.92    15:   2.24  15:   3.98    15:    10.11  15:    21.95
50:   1.14  50:   1.95    50:   2.34  50:   4.00    50:    10.48  50:    21.94
100:  1.19  100:  1.98    100:  2.41  100:  4.04    100:   10.97  100:   21.96
200:  1.44  200:  1.98    200:  2.59  200:  4.05    200:   11.12  200:   22.07
500:  1.80  500:  2.10    500:  3.09  500:  4.13    500:   11.62  500:   22.12
800:  2.22  800:  2.08    800:  3.53  800:  4.18    800:   12.35  800:   22.31
1000: 2.41  1000: 2.09    1000: 3.86  1000: 4.22    1000:  12.67  1000:  22.11
                          1200: 4.08  1200: 4.29    1200:  13.15  1200:  22.46
                          1500: 4.48  1500: 4.19    1500:  13.98  1500:  22.18
                          1750: 4.79  1750: 4.27    1750:  14.48  1750:  22.19
                          2000: 5.08  2000: 4.28    2000:  14.75  2000:  24.18
                                                    2500:  15.72  2500:  24.56
                                                    3000:  16.66  3000:  23.88
                                                    3500:  16.50  3500:  23.04
                                                    4000:  17.40  4000:  22.48
                                                    4500:  18.37  4500:  22.63
                                                    5000:  19.17  5000:  22.81
                                                    6000:  20.92  6000:  22.83
                                                    7000:  22.98  7000:  22.97
                                                    8000:  24.19  8000:  23.01
                                                    9000:  25.68  9000:  23.13
                                                    10000: 26.85  10000: 23.62

```

I'm not sure what are other uses of this function apart from selecting a couple indices for top-k error calculation. For that purpose the heap implementation seems much faster. I don't have it finished yet, but I'll try to send a PR soon.

@deltheil, what do you think?

deltheil:
@apaszke good to have numbers. In terms of pure big-O it seems like it should be the other way around (see [partial sort](https://en.wikipedia.org/wiki/Partial_sorting)) but for practical / implementation reasons it is apparently not.

> I'll try to send a PR soon

:+1:

apaszke:
Uh, so I've been trying to find a good explanation of why the results looks as they do here, and I couldn't find any reasonable answer, so I double checked everything. It appears that I forgot to comment out 
`git submodule update --init --recursive` from install.sh, so in an installation that I used as quickselect in above benchmark, your commit has been reset and that was an old one, using a full sort...

It appears that this PR is a bit faster than when using a heap, even for small k, so I think we should go with it.

Sorry for that, I totally forgot about that line...

deltheil:
@apaszke thanks for these final precisions.

@soumith and others: any interest in merging this? Any remarks?

soumith:
was just waiting to see how Adam's implementation played out. Thanks for the PR.

apaszke:
Yeah, sorry for delaying this. A stupid mistake...

Issue #516, opened 2016-01-27
upvalues are untied after serialization / deserialization


soumith:
If there's an upvalue referred by the two functions, and the functions are serialized / deserialized, the upvalue references are no longer the same.

Repro:

```lua
local i = 1
local res = {}
function res.seti(j) i = j end
function res.printi() print(i) end

print("Before saving, variable i is shared")
res.seti(2)
res.printi()  -- output 2 since variable i is shared among two functions.

torch.save("./tmp", res)
local res2 = torch.load("./tmp")
print("After save/load, variable i is no longer shared")
res2.seti(3)
res2.printi() -- output 2 rather than 3
```

leonbottou:
Cute!

albanD:
This can be mitigated by using torch classes:

```lua
local i = torch.Tensor(1)
local res = {}
function res.seti(j) i[1] = j end
function res.printi() print(i[1]) end

print("Before saving, variable i is shared")
res.seti(2)
res.printi()  -- output 2 since variable i is shared among two functions.

torch.save("./tmp", res)
local res2 = torch.load("./tmp")
print("After save/load, variable i is no longer shared")
res2.seti(3)
res2.printi() -- output 3 as expected
```

Though if you change the saving to disable the `referenced` mode with `torch.save("./tmp", res, nil, false)` it breaks again.

albanD:
@soumith as a quick try [this](https://github.com/albanD/torch7/tree/upvalue_binding) kind of change solve this issue.
The idea is to check when serializing which upvalues are linked.
Then when de-serializing we link back the upvalues that used to be linked.

Positive:
* Now the upvalues stay linked after de-serialization

Negative:
* We store with each upvalue an id -> increase the size of the serialized element.

Do you think it would be worth cleaning up and adding it to the torch7 codebase?

soumith:
Leon and I were discussing this exact solution, discussing technical pitfalls etc. 

I think this is worth adding, but to maintain back-compatibility one has to introduce a LEGACY_TYPE_FUNCTION similar to how we introduced a LEGACY_TYPE_RECUR_FUNCTION: https://github.com/torch/torch7/blob/master/File.lua#L21
The increase in size of the serialized element is super minor, and the benefits outweigh it IMO.

albanD:
[...]

PR #519, opened 2016-01-29
Fix upvalue binding during serialization

albanD:
address issue #516


albanD:
@soumith FYI the travis build does not fail when the `test_writeObject.lua` test fails.
For this PR, it is expected to fail for lua 5.1.
But the error `attempt to call method 'cdata' (a nil value)` was already there for all lua build (and not reported)

soumith:
Thanks. I'm going to see what's up with contbuild.

PR #523, opened 2016-02-02
Revert "Adding argmax index as second return value to max/min calls without specified dimension"

dominikgrewe:
Reverts torch/torch7#489

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -500,23 +500,8 @@ for _,Tensor in ipairs({"ByteTensor", "CharTensor",
    for _,name in ipairs({"min", "max"}) do
       wrap(name,
            cname(name .. "all"),
-           {{name=real, creturned=true},
-            {name="IndexTensor", default=true, returned=true, noreadadd=true,
-            precall = function(arg)
-                local txt = {}
-                return table.concat(txt, '\n')
-            end,
-            postcall = function(arg)
-                 local txt = {}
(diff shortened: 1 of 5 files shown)


dominikgrewe:
We noticed this is breaking code when the output of max or min is passed to another function. Previously one argument would be passed, now two are passed, causing the function to behave differently.

dominikgrewe:
One possible solution is this: Only return the argmax if an index is passed. An index of -1 means that the max should be computed over the entire tensor and the argmax is returned as in the reverted PR. If no index is passed, the max is also computed over the entire tensor, but the argmax is not returned (just like before).

```lua
t = torch.Tensor{{1, 2, 3}, {4, 5, 6}}
= t:max()
6

= t:max(1)
 4  5  6
[torch.DoubleTensor of size 1x3]

 2  2  2
[torch.LongTensor of size 1x3]

= t:max(-1)
6

 2  3
[torch.LongTensor of size 2]
```

Any thoughts?

soumith:
@dominikgrewe isn't that already the case on trunk?

soumith:
the only difference with the version in trunk, then is that instead of returning just one of the max indices, it returns all max indices.

dominikgrewe:
Currently, if you don't specify a dimension, no index is returned.

soumith:
oh, i see that you want to introduce a -1, that seems okay to me. and it wouldn't break tons of code.
When a PR is prepared for this, can you also make sure the cutorch PR is prepared in parallel.

dominikgrewe:
Sure, will do.

PR #524, opened 2016-02-02
Lua 5.2 fix for date/time formatting in CmdLine

soumith:
Lua 5.2 does not support %F and %T, looks like it. Even though they are valid formatting in http://man7.org/linux/man-pages/man3/strftime.3.html

### DIFF
--- a/CmdLine.lua
+++ b/CmdLine.lua
@@ -74,7 +74,7 @@ function CmdLine:silent()
 end
 
 function CmdLine:addTime(name, format)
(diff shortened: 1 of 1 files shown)


deltheil:
It looks like they are supported if `LUA_USE_POSIX` is enabled: http://www.lua.org/source/5.2/loslib.c.html#LUA_STRFTIMEOPTIONS

soumith:
good to know. this patch should help windoze as well, so will keep it as is.

PR #526, opened 2016-02-03
Initialise longSize in PipeFile

kieranmilan:
The previously added longSize field was only initialised in DiskFile and
not in PipeFile. This meant uninitialised memory was used in PipeFile
construction.


soumith:
Thank you!

PR #529, opened 2016-02-03
fixing global leak in print variable

### LINKED ISSUE #528: CmdLine overwrites default print function
Once` CmdLine:log()` has been called, the global print function seems to be overwritten by the print function inside `CmdLine:log()`. I know it sounds weird, but before calling `CmdLine:log()`, doing a `print()` on a table results in nicely formatted and nested output, while after calling `CmdLine:log()`, the output is simply `table: 0x12345678`.

Here is a snippet that showcases this behaviour:
```
cmd = torch.CmdLine()
cmd:text('Testing CmdLine printing and logging')
cmd:text('Options')
cmd:option('-seed',123,'initial random seed')
cmd:option('-booloption',false,'boolean option')
cmd:option('-stroption','mystring','string option')

-- parse input params
params = cmd:parse(arg)
params.rundir = cmd:string('experiment', params, {dir=true})
paths.mkdir(params.rundir)

-- just a random table
a = {1,2,3}

-- print the table before logging
print(a)

-- create log file
cmd:log(params.rundir .. '/log', params)

-- print the table after logging
print(a)
```
The output it produces is as follows:
```
{
  1 : 1
  2 : 2
  3 : 3
}
[program started on Fri Feb  5 00:24:08 2016]
[command line arguments]
stroption	mystring
booloption	false
seed	123
rundir	experiment
[----------------------]
table: 0x417c30b8
```
Notice how printing the table no longer works as expected at the end. It seems [these lines in CmdLine.lua](https://github.com/torch/torch7/blob/master/CmdLine.lua#L193-L209) are guilty.
Changing the function name from `print` to something else, like `printLog` seems to fix this on my local install.

soumith:
Fixes #528

### DIFF
--- a/CmdLine.lua
+++ b/CmdLine.lua
@@ -190,7 +190,7 @@ end
 local oprint = print
 function CmdLine:log(file, params)   
    local f = io.open(file, 'w')
-   function print(...)
+   local function print(...)
       local n = select("#", ...)
       local arg = {...}
       if not self.silentio then


fmassa:
I think this PR breaks the logging, because the prints in the user script won't be logged anymore (that's why we were overwriting the print function, as done in the trepl).

Issue #517, opened 2016-01-27
torch.all only returns true if all numbers in the tensor are odd


dominikgrewe:
It uses a bit-wise reduction with `&=` and initialization 1, i.e. it only looks at the lowest bit of all numbers. It should use a logical and instead.

For example `return torch.ByteTensor(3):fill(2):all()` returns `false`.

andreaskoepf:
@dominikgrewe Seems to be fixed, can we close it?

PR #534, opened 2016-02-04
Fix for issue #517: torch.all

ivpopov:
Fix for "torch.all only returns true if all numbers in the tensor are odd #517"

### DIFF
--- a/lib/TH/generic/THTensorMath.c
+++ b/lib/TH/generic/THTensorMath.c
@@ -1963,12 +1963,12 @@ LAB_IMPLEMENT_BASIC_FUNCTION(abs,abs)
   { \
     THArgCheck(tensor->nDimension > 0, 1, "empty Tensor"); \
     int sum = INIT_VALUE;                               \
-    TH_TENSOR_APPLY(real, tensor, sum OP *tensor_data;); \
+    TH_TENSOR_APPLY(real, tensor, sum = sum OP *tensor_data;); \
     return sum; \
   }
 
-TENSOR_IMPLEMENT_LOGICAL_SUM(logicalall, &=, 1)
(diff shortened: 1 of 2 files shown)


lib/TH/generic/THTensorMath.c
```diff
@@ -1958,17 +1958,21 @@ LAB_IMPLEMENT_BASIC_FUNCTION(abs,abs)
...
+int THTensor_(logicalall)(THTensor *tensor)
+{
+  THArgCheck(tensor->nDimension > 0, 1, "empty Tensor");
+  int sum = 1;
+  TH_TENSOR_APPLY(real, tensor, sum = sum && *tensor_data;);
+  return sum;
+}
 
-TENSOR_IMPLEMENT_LOGICAL_SUM(logicalall, &=, 1)
-TENSOR_IMPLEMENT_LOGICAL_SUM(logicalany, |=, 0)
```
timharley: Is it not sufficient to change `&=` to `&&` and `|=` to `||` here? This would use the correct logical operators rather than bitwise operators. `sum` would probably need to be a bool in this case.

ivpopov:
I am not sure I understand your suggestion. &= and |= assign to the left operand. && and || do not.

ivpopov:
After discussing with Tim that there is no hidden magic in using |= instead of logical ||, changing macros and operators to sum = sum OP ... and && and ||

soumith:
squash your commits and this is good to merge. thnx

ivpopov:
Done

soumith:
@ivpopov still dont see a squash. did you force-push back to the same PR?

soumith:
Thanks!

PR #537, opened 2016-02-08
Fixes a memory leak in Torch/Lapack.

fastturtle:
This fix was implemented by @tkoeppe.

### DIFF
--- a/lib/TH/THLapack.h
+++ b/lib/TH/THLapack.h
@@ -3,15 +3,21 @@
 
 #include "THGeneral.h"
 
+#define THErrorHandler(X) X
+
 #define THLapack_(NAME) TH_CONCAT_4(TH,Real,Lapack_,NAME)
-#define THLapackCheck(fmt, func, info , ...)						\
-if (info < 0) {														\
+
+#define THLapackCheck(fmt, func, info , ...) \
+  THLapackCheckWithHandler(fmt, THErrorHandler(), func, info, ##__VA_ARGS__)
+
+#define THLapackCheckWithHandler(fmt, cleanup, func, info , ...)    \
+if (info < 0) {                                                     \
+  cleanup                                                           \
   THError("Lapack Error in %s : Illegal Argument %d", func, -info); \
-} else if(info > 0) {												\
-  THError(fmt, func, info, ##__VA_ARGS__);							\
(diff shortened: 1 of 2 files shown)


lib/TH/THLapack.h
```diff
@@ -3,15 +3,21 @@
...
-if (info < 0) {														\
-  THError("Lapack Error in %s : Illegal Argument %d", func, -info); \
-} else if(info > 0) {												\
-  THError(fmt, func, info, ##__VA_ARGS__);							\
-}																	\
 
+#define THLapackCheck(fmt, func, info , ...) \
+  THLapackCheckWithHandler(fmt, THErrorHandler(), func, info, ##__VA_ARGS__)
 
+#define THLapackCheckWithHandler(fmt, cleanup, func, info , ...)    \
```
tkoeppe: This should probably not be called "handler", but rather something like "cleanup" or "finally". "Handler" was just a placeholder name I came up with on short notice.

lib/TH/generic/THTensorLapack.c
```diff
@@ -513,7 +513,10 @@ void THTensor_(potrf)(THTensor *ra_, THTensor *a, const char *uplo)
 
   /* Run Factorization */
   THLapack_(potrf)(uplo[0], n, THTensor_(data)(ra__), lda, &info);
-  THLapackCheck("Lapack Error %s : A(%d,%d) is 0, A cannot be factorized", "potrf", info, info);
+  THLapackCheckWithHandler(
```
tkoeppe: There are at least three similar call sites of the erstwhile `THLapackCheck` that may require the same cleanup. They just don't have test coverage :-)

lib/TH/THLapack.h
```diff
@@ -3,15 +3,21 @@
 
 #include "THGeneral.h"
 
+#define THErrorHandler(X) X
```
tkoeppe: I have since found further leaks elsewhere, for which I need the same construction, so I propose instead to

     #define THCleanup(...) __VA_ARGS__

in `THGeneral.h`. We will then reuse that pattern both in TH and in THLapack.

tkoeppe:
I created a more comprehensive pull request in #542 which subsumes this.

PR #540, opened 2016-02-11
Add torch.load32 and torch.load64

Atcold:
Now it is possible to load objects from binary files generated on OS with both 32 and 64 bits.
This commit integrates https://github.com/torch/torch7/pull/476 missing API.

### DIFF
--- a/File.lua
+++ b/File.lua
@@ -380,7 +380,7 @@ function torch.save(filename, object, mode, referenced)
    file:close()
 end
 
-function torch.load(filename, mode, referenced)
+local function loadWithLongSize(filename, mode, referenced, size)
    assert(mode == nil or mode == 'binary' or mode == 'ascii', '"binary" or "ascii" (or nil) expected for mode')
    assert(referenced == nil or referenced == true or referenced == false, 'true or false (or nil) expected for referenced')
    mode = mode or 'binary'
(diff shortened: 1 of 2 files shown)


culurciello:
+1

soumith:
i understand that you (TeraDeep) want this into torch, can everyone from Teradeep stop posting +1 as a comment, it's not helpful. I'll discuss with the others and see whether we want this in or if we prefer to keep just longSize.

File.lua
```diff
@@ -393,6 +393,32 @@ function torch.load(filename, mode, referenced)
...
+   file:referenced(referenced)
+   file:longSize(8)
+   local object = file:readObject()
+   file:close()
+   return object
+end
+
+-- Loads binary files saved on a 32 bit OS
+function torch.load32(filename, referenced)
+   assert(referenced == nil or referenced == true or referenced == false, 'true or false (or nil) expected for referenced')
```
tkoeppe: This looks like a lot of duplication. Why not have the public functions call a private, parametrized function `loadWithLongSize(filename, references, size)`?

Atcold: Yes, absolutely right! Will do it!
What about
```lua
loadWithLongSize(filename, mode, references, size)
```
so that even `torch.load()` calls the same parametrised function?

Atcold:
@tkoeppe, done :grimacing: What do you think?

tkoeppe:
@Atcold: Just speaking in terms of code quality, I think this is an improvement. Whether this is desirable in Torch is not for me to say, but I think now you're proposing a change that feels like it fits rather than just tacks stuff on.

Can you please squash and repush?

Atcold:
OK, @tkoeppe, let's hear from @soumith :grimacing:

mijailronald23:
i install all the files but i can't use that 
============================
model=torch.load32('model32.net')
[string "model=torch.load32('model32.net')"]:1: attempt to call field 'load32' (a nil value)
stack traceback:
	[string "model=torch.load32('model32.net')"]:1: in main chunk
	[C]: in function 'xpcall'
	/usr/local/share/lua/5.1/trepl/init.lua:669: in function 'repl'
	/usr/local/lib/luarocks/rocks/trepl/scm-1/bin/th:199: in main chunk
	[C]: at 0x00406260

Atcold:
@mijailronald23, they have not yet accepted this small PR. I cannot do anything about it.

soumith:
after further thought and discussion, dont see the need to add more root functions. we are already overwhelmed by how many functions there are in torch.*. What's better is to have a mode for torch.load that might be like 'b32', 'b64'

Atcold:
OK. Do you want me to implement that?

Atcold:
@mijailronald23, the function is now integrated. See #579.

PR #541, opened 2016-02-11
Fixing a lua stack overflow.

zakattacktwitter:
If you happen to pass in a big table to torch.cat and you were near the end of the lua stack you could corrupt memory. This fix tells lua to grow the stack if it needs to.


soumith:
Thanks!

PR #542, opened 2016-02-11
Improve memory cleanup handling and fix memory leaks

tkoeppe:
Two commits in this PR create a generic facility for passing cleanup code to various error macros that do not return. The third commit adds clean-up to various `APPLY` macro uses in `THTensorMath.c`.

### DIFF
--- a/lib/TH/THGeneral.h.in
+++ b/lib/TH/THGeneral.h.in
@@ -57,13 +57,27 @@ TH_API void THSetGCHandler( void (*torchGCHandlerFunction)(void *data), void *da
 TH_API void THHeapUpdate(long size);
 
 #define THError(...) _THError(__FILE__, __LINE__, __VA_ARGS__)
-#define THArgCheck(...) _THArgCheck(__FILE__, __LINE__, __VA_ARGS__)
+
+#define THCleanup(...) __VA_ARGS__
+
+#define THArgCheck(...)                                               \
+do {                                                                  \
+  _THArgCheck(__FILE__, __LINE__, __VA_ARGS__);                       \
+} while(0)
+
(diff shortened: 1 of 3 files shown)


soumith:
This PR looks good, but I dont see the changes you made to THTensorLapack.c (in #537 ) subsumed in this PR. Is this a work-in-progress?

tkoeppe:
Yes, correct, the THTensorLapack.c changes will be a bit more substantial than in #537. I'll make a new PR when that commit is ready.

soumith:
Thank you!

nagadomi:
The first commit causes a compile error in cutorch at https://github.com/torch/cutorch/blob/master/lib/THC/THCTensorScatterGather.cu#L129.

tkoeppe:
@nagadomi: Oh no, I'm sorry, I didn't check cutorch. Do you know what the error is and whether there's a simple fix? My upcoming memory leak fixes use the new cleanup macro, so I would very much like to keep it in some form. Thanks!

tkoeppe:
@nagadomi: Ah, I see what it is. Because I changed the `THArgCheck` macro to have statement syntax, you cannot use it as an expression anymore. I think that change is an improvement in terms of consistency, but of course I should have checked users more carefully.

Would you perhaps be able to a) see if any further cutorch code uses `THArgCheck` as an expression, and b) make a change to cutorch that removes the `return`? Maybe we can "roll forward" over this breakage :-)

Thanks a lot!

soumith:
pushed fixes to cutorch, hopefully this wont happen again.

tkoeppe:
Thanks. I can try and also check out cutorch and make it use my local version of Torch to check things in the future. (Though hopefully I won't make any wide-ranging changes after this; I expect future memory leak cleanup to be purely local.)

PR #551, opened 2016-02-16
Making torch7 and his exports work with external lua; some bugfixes.

borisfom:
The changes here are about: 
- copied FindLua from qtLua (appeared to be better fit than cmake's one) - needed to properly detect location of Lua libraries in case Torch is using external Lua (like , the case of current luarocks .deb where LUA_LIBDIR is not actually pointing to Lua libs.)
- Lua variables fetched from Luarocks via FindLua (LUA_INCDIR) etc, are exported so Torch modules do not have to have the whole shebang of LuaRocks valuables passed to them. They only need to know where to find Torch cmake exports.
- Fixed a bug in TorchPackage that prevented modules like cunn be converted to use it: 
if ($src) does not work if $src i a macro, as cmake does not consider macro argument to be a variable (!).
So only the case that had an upper scope variables named $src and luasrc would work properly. With this change, this restriction is lifted.
- Some absolute paths used where relative paths used to be. 
- TorchPaths adjusted to the situation where (current) installation dir is not the same where Lua/Luarocks reside.

### DIFF
--- a/doc/storage.md
+++ b/doc/storage.md
@@ -44,11 +44,6 @@ function. By default, this "points" on `torch.DoubleStorage`.
 Returns a new `Storage` of type `TYPE`. Valid `TYPE` are `Byte`, `Char`, `Short`,
 `Int`, `Long`, `Float`, and `Double`. If `size` is given, resize the
 `Storage` accordingly, else create an empty `Storage`.
-The optional second argument `ptr` is a number whose value is a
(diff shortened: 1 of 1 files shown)


tkoeppe:
Squash and repush please? Each commit in a pull request should serve precisely one purpose and be appropriately labelled.

Issue #227, opened 2015-05-04
torch.svd allocates too large a matrix with 'S'


soumith:
torch.svd allocates too large a matrix in the setting where you take the svd of an mxn matrix A with n>m, independent of  whether the 'S' flag is used.  If it is going to return an mxm matrix for the right singular vectors (instead of an mxn "economy sized" matrix), it should return an o.n. basis for R^m, with the first n vectors in the basis spanning the range of  A; currently the last m-n columns of the right singular value matrix are junk...

Reported by @aszlam

yf225:
Instead of allocating an m x m matrix, I propose that we should return an m x n matrix by tossing out the last m-n columns of the original matrix. This way we can save quite a bit of memory, since the junk values are useless anyways. 

Thoughts? I can work on this if the change makes sense.

soumith:
cc: @koraykv @dominikgrewe we're proposing this change. Are you okay with this? what are your thoughts.

soumith:
ping ping @koraykv @dominikgrewe

dominikgrewe:
Sorry for the delay. I'm not very familiar with svd but @koraykv said he wanted to have a look.

koraykv:
Yes, in the current implementation the flags for fat matrices are not implemented, only tall case is considered. Sorry for taking long.

PR #556, opened 2016-02-23
Remove junk values from right-singular vectors matrix in SVD

yf225:
Per discussion at [https://github.com/torch/torch7/issues/227](https://github.com/torch/torch7/issues/227), we can remove junk values from right-singular vectors matrix in SVD when using the 'S' option, to reduce memory footprint.

### DIFF
--- a/lib/TH/generic/THTensorLapack.c
+++ b/lib/TH/generic/THTensorLapack.c
@@ -418,6 +418,9 @@ void THTensor_(gesvd2)(THTensor *ru_, THTensor *rs_, THTensor *rv_, THTensor *ra
                                THTensor_(free)(work);),
                            "gesvd", info);
 
+  if (*jobu == 'S')
+    THTensor_(narrow)(rv__,NULL,1,0,k);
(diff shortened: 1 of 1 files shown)


lib/TH/generic/THTensorLapack.c
```diff
@@ -418,6 +418,9 @@ void THTensor_(gesvd2)(THTensor *ru_, THTensor *rs_, THTensor *rv_, THTensor *ra
                                THTensor_(free)(work);),
                            "gesvd", info);
 
+  if (*jobu == 'S')
+    THTensor_(narrow)(rv__,NULL,1,0,k);
```
soumith: instead of doing this, wouldn't you initialize rv_ / rv__ to only be k elements in the first place, instead of allocating a bigger tensor and then narrowing (which doesn't actually save memory because this tensor is just a narrow into a larger storage).

tkoeppe:
@yf225: Squash, rebase and force-push?

PR #557, opened 2016-02-23
doc/storage: add precisions for ptr, size mode

deltheil:
In practice the storage is only [ref-counted](https://github.com/torch/torch7/blob/a8684284708badfba04723febd9247385cb2dcfc/generic/Storage.c#L74) in such a case.


soumith:
Thanks@!

PR #558, opened 2016-02-24
Fixed a bug in TorchPackage macro arguments test

borisfom:
.. "if ($src)" does not work if $src i a macro, as cmake does not consider macro argument to be a variable (!). So only the case that had an upper scope variables named $src and luasrc worked before. 
Also, C linker language specified for sources as CMake gets lost if all sources are .cu CUDA files
So now cunn can use ADD_TORCH_PACKAGE  macro (used to break on both issues above)


soumith:
Thanks Boris!

PR #562, opened 2016-02-25
Replace torch.Tester with totem.Tester + extra stuff.

davidsaxton:
This should bring a lot of benefit to code that uses torch.Tester (totem will
eventually become deprecated). Note that torch.Tester and totem.Tester once
shared the same code - this change brings it full circle.

At a glance, extra functionality includes:
- A general equality checker that accepts many different objects.
- Deep table comparison with precision checking.
- Stricter argument checking in using the test functions.
- Better output.
- torch.Storage comparison.
- Extra features for fine-grained control of testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,3 +57,4 @@ script:
 - ${TESTLUA} -ltorch -e "t=torch.test(); if t.errors[1] then os.exit(1) end"
 - cd test
 - ${TESTLUA} test_writeObject.lua
+- ${TESTLUA} test_Tester.lua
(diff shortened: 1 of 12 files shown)


davidsaxton:
Notes on improved strictness: TestSuite is added to catch tests with duplicate names, and it has caught two in test.lua: there were two non-identical tests called sin(), one of which I have renamed sin_2, and there were two tests called fxcorr3_fxcorr2_eq with identical content; I've deleted one.

Tester will now throw an error if assertTableEq is called with non-tables, which I think is desirable behaviour (better to catch potentially bad code that is suppose to produce tables). I need to create a PR to change a test in torch/nn, which calls assertTableEq to compare numerical values in a few places (see nntest.View - https://github.com/torch/nn/blob/master/test.lua#L4394).

Tester will now check tensors for dimension (or size) (so {{1}} ~= {1}). In order not to break too many tests, the default behaviour of assertTensorEq is to ignore the dimension, so that these tests still pass. However, it will still print a warning if assertTensorEq is used with tensors of different dimensions. Furthermore, the other functions that now compare tensors (assertTableEq, which before did only a shallow comparison of the data, and assertGeneralEq, a new function) will check for dimensionality too.

I've updated some of the tests so that they pass in tensors of the same dimension to avoid printing a warning.

soumith:
Thanks David!

PR #567, opened 2016-02-26
Add a hook to File:writeObject

nkoumchatzky:
* New Api is File:writeObject(object, debugname, hook)
* Backward-compatible
* Functionality: modifies objects to serialize on-the-fly before saving them while keeping track of the File:writeObject(...) nice reference-sharing. Useful when we want to save an object under a certain package config, and load it under another.

### DIFF
--- a/File.lua
+++ b/File.lua
@@ -104,7 +104,9 @@ local function formatStack(objectNameStack)
    return table.concat(parts, '.')
 end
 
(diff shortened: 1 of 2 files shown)


culurciello:
Good job!

nkoumchatzky:
Thanks @culurciello !

soumith:
looks good, thanks nico!

PR #568, opened 2016-02-27
Add math functions trunc, frac, rsqrt, lerp

andreaskoepf:
- [x] CPU implementation of trunc, frac, rsqrt, lerp
- [x] documentation
- [x] unit tests

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -10,7 +10,7 @@ interface:print([[
 #include "THMath.h"
 #include "luaT.h"
 #include "utils.h"
-                ]])
+]])
 
 -- specific to torch: we generate a 'dispatch' function
 -- first we create a helper function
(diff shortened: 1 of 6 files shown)


lib/TH/THTensor.c
```diff
@@ -7,6 +7,18 @@
 #include "THTensorDimApply.h"
 #include "THMath.h"
 
+inline double frac(double x) {
```
fmassa: I wonder if those functions shouldn't go to the newly created [THMath.h](https://github.com/torch/torch7/blob/master/lib/TH/THMath.h). What do you think ?

andreaskoepf:
@fmassa THMath.h is indeed a better place for the functions. thx

andreaskoepf:
IMO it would be better if [genericSingleOpTest](https://github.com/torch/torch7/blob/master/test/test.lua#L44) was a 'real' function that would get the functions to test as arguments instead of using string-replacement...

soumith:
they look good. squash and they're ready for commit.

andreaskoepf:
rebased & squashed

soumith:
Thanks Andreas!

PR #580, opened 2016-03-04
Fix a bug in :isSetTo

malcolmreynolds:
The case of self->storage == NULL was not checked, so any two Tensors
with no storage would effectively appear to be shared with each other.
Added a test to prevent regressions.


soumith:
Thanks!

PR #581, opened 2016-03-04
Extend torch.bernoulli()

sergomezcol:
Now it also accepts a tensor of probabilities instead of a single Lua number.

Added a test that checks that the output is binary.


soumith:
Thanks!

PR #582, opened 2016-03-07
Fix torch.cat for table arguments

JoostvDoorn:
Fixes #548.

If this is merged [/torch/cutorch/blob/master/test/test.lua#L2597](https://github.com/torch/cutorch/blob/master/test/test.lua#L2597) should be modified to torch.cat, instead of torch.CudaTensor.cat.

PR #583, opened 2016-03-07
Fix broken torch.cat{} and added tests

JoostvDoorn:
Sorry for breaking torch.cat{} in my previous pull request #582, also broke torch.cat{1}, torch.add{} etc. resulting in a segmentation fault.

Rewrote the function to check for the table being empty, and put it in a function to allow it to use the default tensor if it is empty or contains an invalid tensor. Also added additional unit tests.

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -51,6 +51,18 @@ static int torch_isnonemptytable(lua_State *L, int idx)
 }
 ]])
 
+
+interface:print([[
+static const void* torch_istensorarray(lua_State *L, int idx)
+{
+  if (!torch_isnonemptytable(L, idx)) return 0;
+
+  lua_checkstack(L, 1);
+  lua_rawgeti(L, 1, 1);
+  return (torch_istensortype(L, luaT_typename(L, -1)));
+}
+]])
+
 interface.dispatchregistry = {}
 function interface:wrap(name, ...)
    -- usual stuff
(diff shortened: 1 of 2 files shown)


TensorMath.lua
```diff
@@ -51,6 +51,18 @@ static int torch_isnonemptytable(lua_State *L, int idx)
...
 
+
+interface:print([[
+static const void* torch_istensorarray(lua_State *L, int idx)
+{
+  if (!torch_isnonemptytable(L, idx)) return 0;
+
+  lua_checkstack(L, 1);
+  lua_rawgeti(L, 1, 1);
+  return (torch_istensortype(L, luaT_typename(L, -1)));
```
dominikgrewe: Shouldn't you call `lua_remove` here instead of below? If `torch_istensortype` returns NULL, the `lua_remove` below isn't getting executed, so you leave the table element on the stack.

JoostvDoorn: You are right, this should fix it?
```c
const void* tname;
if (!torch_isnonemptytable(L, idx)) return 0;

lua_checkstack(L, 1);
lua_rawgeti(L, 1, 1);
tname = (torch_istensortype(L, luaT_typename(L, -1)));
lua_pop(L, 1);
return tname;
```

dominikgrewe: Shouldn't it be `lua_rawgeti(L, idx, 1)`? And you have to be careful with the `lua_pop` because `torch_istensortype` leaves a table on the stack if it doesn't return NULL I think. So maybe call `lua_remove` like so:

```lua
lua_rawgeti(L, idx, 1);
tensor_idx = lua_gettop(L);
tname = ...
lua_remove(L, tensor_idx)
```

JoostvDoorn: Yes, it should be `lua_rawgeti(L, idx, 1)`. I did inspect the values on the stack to make sure the value of lua_rawgeti was deleted with a stackDump function, and in all cases this seemed to work, but this is probably better.

dominikgrewe: @JoostvDoorn Are you planning to fix this? Would be great to get this merged!

JoostvDoorn: @dominikgrewe Yes I will. I have some time tomorrow to get this done.

dominikgrewe: Great, thanks! :)

PR #585, opened 2016-03-09
removing sys dependency for now

soumith:
cc: @borisfom 

for now removing the sys dependency. I've already made the require behind a pcall in an earlier commit, so the dependency is very much optional.


borisfom:
Right thing to do right now. Impressed :)

PR #587, opened 2016-03-11
Lua 5.3 integer support

luketwitter:
This shouldn't change any behavior when compiled with Lua < 5.3 or LuaJIT.

Will post the corresponding change to cwrap. This works fine independent of that, but some functions like :fill() on LongTensor won't properly support 64-bit inputs until cwrap is updated.

### DIFF
--- a/generic/Storage.c
+++ b/generic/Storage.c
@@ -2,6 +2,8 @@
 #define TH_GENERIC_FILE "generic/Storage.c"
 #else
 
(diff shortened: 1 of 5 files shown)


soumith:
@clementfarabet is this ready for merge? I haven't had a look

clementfarabet:
It's been tested on our side, I think it's safe to merge.

PR #590, opened 2016-03-12
Fix bad size evaluation

Atcold:
xmax, xmin and step are accreal. Therefore, computations are carried out in both int and double logic. The separate division does introduce a bug when int logic is used.

## Before
```lua
th> torch.range(1, 25, 5, 'torch.LongTensor')
  1
  6
 11
 16
 21
 26
[torch.LongTensor of size 6]

th> torch.range(1, 25, 5, 'torch.FloatTensor')
  1
  6
 11
 16
 21
[torch.FloatTensor of size 5]
```

## After
```lua
th> torch.range(1, 25, 5, 'torch.FloatTensor')
  1
  6
 11
 16
 21
[torch.FloatTensor of size 5]

th> torch.range(1, 25, 5, 'torch.LongTensor')
  1
  6
 11
 16
 21
[torch.LongTensor of size 5]

th> torch.range(1, 25, 5, 'torch.DoubleTensor')                                                      
  1
  6
 11
 16
 21
[torch.DoubleTensor of size 5]

th> torch.range(1, 25, 5, 'torch.IntTensor')                                                         
  1
  6
 11
 16
 21
[torch.IntTensor of size 5]

th> torch.range(1, 25, 5, 'torch.ByteTensor')                                                        
  1
  6
 11
 16
 21
[torch.ByteTensor of size 5]
```

### DIFF
--- a/lib/TH/generic/THTensorMath.c
+++ b/lib/TH/generic/THTensorMath.c
@@ -1370,7 +1370,7 @@ void THTensor_(range)(THTensor *r_, accreal xmin, accreal xmax, accreal step)
   THArgCheck(((step > 0) && (xmax >= xmin)) || ((step < 0) && (xmax <= xmin))
               , 2, "upper bound and larger bound incoherent with step sign");
 
-  size = (long)((xmax/step - xmin/step)+1);
+  size = (long) (((xmax - xmin) / step) + 1);
 
   THTensor_(resize1d)(r_, size);
 


andreaskoepf:
This PR would revert the changes introduced in https://github.com/torch/torch7/pull/252 ... maybe we should cast min, max & step to double always?

Atcold:
@andreaskoepf, I think that one simple cast would do the work. I.e.
```c
size = (long) (((double) (xmax - xmin) / step) + 1);
```

andreaskoepf:
If we want to keep the original 'fix' which especially works well if xmin/xmax are divisible by step we could use something like `size = (long)((xmax/(double)step - xmin/(double)step)+1);` ... regarding real/accreal I am not sure .. real becomes char/byte/short/int/long for the integer tensor types. accreal is probably the next larger type long for IntTensor... probably it was introduced to get double inputs for FloatTensor?

andreaskoepf:
I do not know if people want to work with really high ranges...but a double casts for values > 2^52 could of course also become a problem since double mantissa has only 52 bits. Probably the 'perfect' solution would be to use `(long)((xmax-ymin)/step+1)` for integer types and the `(long)((xmax/step - xmin/step) + 1)` for floating point... that would be possible by #ifdef  for `TH_REAL_IS_FLOAT` or `TH_REAL_IS_DOUBLE`. @Atcold what do you think?

Atcold:
I don't think it is necessary. Right now the division is already done in `double`, for the case in which it is necessary. Otherwise, `long` is fine, as long as the subtraction is done **berfore** the division.

andreaskoepf:
@Atcold My comment regarding >2^52 might be relevant only for cases in which TH is used from C or with lua > 5.3.

For luajit you see something like:
```
th> x = 1152921504606846977
th> ffi.typeof('long')(x)
1152921504606846976LL
th> ffi.typeof('long')(ffi.typeof('double')(1152921504606846977LL))
1152921504606846976LL
```

or

```
th> y = 1152921504606846977LL
th> y
1152921504606846977LL
th> ffi.typeof('long')(ffi.typeof('double')(y))
1152921504606846976LL
```

So for xmax = 1152921504606846977, xmin = 0, step = 1 I think the size calculation would be wrong (this is of course a hypothetical view since it will out-of-memory anyway).

andreaskoepf:
@dl11 maybe you want to comment on this too?

andreaskoepf:
@d11maybe you want to comment on this, too?

andreaskoepf:
@d11 maybe you want to comment on this too?

Atcold:
Isn't this supposed to be an easy merge, which reverts a change that introduced a bug?

soumith:
Thanks Alfredo!

PR #593, opened 2016-03-14
Fix multinomial regression.

nkoumchatzky:
Basically forces all the cumulative distribution to DoubleTensor and all the intermediate states to doubles as well whatever the input probs type.

Also added a perf+rare-event test based on #418 @andresy's  example for further testing but I didn't notice any major slowdown.

### DIFF
--- a/lib/TH/generic/THTensorRandom.c
+++ b/lib/TH/generic/THTensorRandom.c
@@ -75,7 +75,7 @@ void THTensor_(multinomial)(THLongTensor *self, THGenerator *_generator, THTenso
   int start_dim = THTensor_(nDimension)(prob_dist);
   long n_dist;
   long n_categories;
-  THTensor* cum_dist;
+  THDoubleTensor* cum_dist;
   int i,j,k;
 
   if (start_dim == 1)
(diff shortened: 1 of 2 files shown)


lib/TH/generic/THTensorRandom.c
```diff
@@ -103,28 +103,27 @@ void THTensor_(multinomial)(THLongTensor *self, THGenerator *_generator, THTenso
...
-      THStorage_(set)(
+      THDoubleStorage_set(
         cum_dist->storage, \
         cum_dist->storageOffset+j*cum_dist->stride[0], \
         sum \
       );
     }
-    THArgCheckWithCleanup((sum > 0), THCleanup(THTensor_(free)(cum_dist);), 2,
-                          "invalid multinomial distribution (sum of probabilities <= 0)");
+    THArgCheck((sum > 0), 2, "invalid multinomial distribution (sum of probabilities <= 0)");
```
timharley: Please use `THArgCheckWithCleanup`, as the original did.

nkoumchatzky: Good point, just did.

soumith:
Thanks!

nkoumchatzky:
Np!

lib/TH/generic/THTensorRandom.c
timharley: Thanks!

PR #602, opened 2016-03-17
Revert to pattern matching in isTypeOf()

jgehring:
This reverts the "enclosed" class matching introduced in 9c75e57, which
broke matching against patterns. In contrast to the original implementation,
non-pattern strings are always matched fully against type names to minimize
surprises caused by e.g. common prefix.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -125,47 +125,25 @@ function torch.type(obj)
    return class
 end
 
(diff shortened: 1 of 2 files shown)


andresy:
Thanks for fixing this! :+1:

Issue #607, opened 2016-03-20
mod and cmod bugs


j-wilson:
Bug in recently added mod and cmod functions:

__Correct Behavior__
```
th> a, b = -10, 3
th> a % b
2
th> a - b*math.floor(a/b)
2
```
Note that [maths.md](https://github.com/torch/torch7/blob/master/doc/maths.md#res-torchmodres-tensor-value) explicitly defines `mod` and `cmod` as performing this second computation

__Trouble Behavior__
```
th> a, b = -10, 3
th> torch.mod(torch.Tensor{a}, b)
-1
[torch.DoubleTensor of size 1]

th> torch.cmod(torch.Tensor{a}, torch.Tensor{b})
-1
[torch.DoubleTensor of size 1]
```

soumith:
fixed via #620

PR #608, opened 2016-03-21
Fix torch.mod and torch.cmod behaviors.

liboyue:
I met this issue #607 too. The function `fmod` in the math library behaves a little bit different with the `%` operator.
I replaced `fmod(a, b)` by `a - b*floor(a/b)` and modified test. Computing 100,000,000 remainders by
 `a - b*floor(a/b)`  will cost 0.1s more than `fmod(a, b)`, so this change will not affect the performance too much.

### DIFF
--- a/DiskFile.c
+++ b/DiskFile.c
@@ -62,10 +62,18 @@ static int torch_DiskFile_longSize(lua_State *L)
   return 1;
 }
 
+static int torch_DiskFile_noBuffer(lua_State *L)
+{
+  THFile *self = luaT_checkudata(L, 1, "torch.DiskFile");
+  THDiskFile_noBuffer(self);
+  lua_settop(L, 1);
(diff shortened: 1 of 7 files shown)


andreaskoepf:
@j-wilson @liboyue Thank for reporting and providing a patch.

Should we rename the original impl to fmod in order to provide the behaviour of the standard C fmod() function as well? Note that the lua math module itself also defines fmod to offer the symmetric modulo operation. 

There are probably a lot of situations where users expect a `-10.1 modulo 10` operation to return `-0.1` instead of `9.9`.

liboyue:
@andreaskoepf Yes, I think we need to add `fmod()` and `modf()` for the two types of mod, and leave `mod()` as it is for compatibility. I can provide a new patch later :)

PR #611, opened 2016-03-22
Improvements to torch.{linspace, logspace, range}.

davidsaxton:
Allow end < start in linspace and logspace (no mathematical reason to forbid
this, and it makes it match numpy behaviour).

Also don't raw-resize tensor unnecessarily. This makes it work with
non-contiguous tensors (e.g., can now do a range on a slice of a tensor).


soumith:
Thanks David!

Issue #610, opened 2016-03-22
bug in operator overloading in torch.class


iamalbert:
Torch class system seems to have bug handling overloaded operators when the instance is on the right-hand-side

```lua 
local cb =  function(x,y)                       
    print("\ncalled: ", x, y, "\n")            
end 

torch.class('TestClass') 

local mt = torch.getmetatable('TestClass')          
mt.__add__ =  cb
                                               
local m = TestClass()                               
                                               
print( m )            -- ok
print( m + 2 )      -- ok                   
print( 2 + m )      -- error                   
```

which has error
```
internal error in __add: no metatable
```

soumith:
it's not a bug of the torch class system. 2 + m will call the __add__ of the "number" metatable. The LHS determines which metatable is called.

iamalbert:
OK, but this is an contradiction to the original behavior of overloading operator in Lua

soumith:
hmm, re-reading the Programming in Lua page, this does seem like a bug: http://www.lua.org/pil/13.1.html

Thanks for pointing it out

soumith:
I'll try to trace where this comes from, if you have any pointers feel free to let me know.

iamalbert:
I suppose that the handler is in the implementation of `static int luaT_mt__add(lua_State *L);`
( https://github.com/torch/torch7/blob/master/lib/luaT/luaT.c#L1199 ). It is an macro defined at https://github.com/torch/torch7/blob/master/lib/luaT/luaT.c#L1171 , which only checks the LHS for metatable. (https://github.com/torch/torch7/blob/master/lib/luaT/luaT.c#L1174)

we can provide another version for all binary operators.

Atcold:
Wow! This has been there since the beginning! And I thought as well it was Lua's fault!
Good to know Lua supports mixed input!

PR #612, opened 2016-03-22
separate handler for binary/unary operators in luaT

iamalbert:
#610 
For binary operators, follow the original Lua behavior. (http://www.lua.org/manual/5.1/manual.html#2.10.1)

 First, try the first operand. If its type does not have a metatable defining a handler for such operation, then try the second operand.

### DIFF
--- a/lib/luaT/luaT.c
+++ b/lib/luaT/luaT.c
@@ -1167,12 +1167,17 @@ static int luaT_mt__newindex(lua_State *L)
   return 0;
 }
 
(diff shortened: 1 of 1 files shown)


soumith:
Thanks Albert. I think now one just has to fix __div__ and __mod__ to support this case. https://github.com/torch/torch7/blob/master/generic/TensorOperator.c#L5 , __add__, __sub__ and __mul__ already handle it correctly

PR #615, opened 2016-03-24
Add DiskFile:noBuffer()

colesbury:
The noBuffer function disables read and write buffering via setvbuf.

This is particularly useful for random reads from files on networked storage, which otherwise results in much larger network transfers.


soumith:
i'm surprised that setvbuf exists on Windoze. Looks good to me.

soumith:
Thanks!

Issue #617, opened 2016-03-25
torch.chunk fails to split correctly


vivanov879:
th> torch.rand(11):chunk(5)
{
  1 : DoubleTensor - size: 3
  2 : DoubleTensor - size: 3
  3 : DoubleTensor - size: 3
  4 : DoubleTensor - size: 2
}

vivanov879:
splits in only 4 chunks, while 5 is specified

andreaskoepf:
@vivanov879 I created an alternative impl for chunk which always tries to return the a table with the number of entries requested by the user. My implementation has two 'effects' which need to be discussed: 

1. The size of more than the last tensor in the result table can vary.
2. There are cases with empty tensors if the desired chunk count exceeds the size of the source dimension.

vivanov879:
works ok

andreaskoepf:
After discussion with @soumith & @nicholas-leonard the original torch.chunk() behavior is considered 'by-design'. For anybody interested a solution that always returns exactly the requested number of chunks I created [this gist](https://gist.github.com/andreaskoepf/0412f4a5bfe0531f226071b34802376f).

PR #619, opened 2016-03-27
Alternate implementation of torch.chunk()

andreaskoepf:
This PR tries to address https://github.com/torch/torch7/issues/617. The new implementation returns a table with nChunk entries, even in the case that the tensor has < nChunk entries in the specified dimension. 

TODO: If new behaviour is as intended the documentation has to be updated since it currently says chunk() would use split().

Old behaviour:
```
th> torch.rand(11):chunk(5)
{
  1 : DoubleTensor - size: 3
  2 : DoubleTensor - size: 3
  3 : DoubleTensor - size: 3
  4 : DoubleTensor - size: 2
}

th> torch.rand(10):chunk(7)
{
  1 : DoubleTensor - size: 2
  2 : DoubleTensor - size: 2
  3 : DoubleTensor - size: 2
  4 : DoubleTensor - size: 2
  5 : DoubleTensor - size: 2
}

th> torch.rand(5):chunk(10)
{
  1 : DoubleTensor - size: 1
  2 : DoubleTensor - size: 1
  3 : DoubleTensor - size: 1
  4 : DoubleTensor - size: 1
  5 : DoubleTensor - size: 1
}
```

New behaviour:
```
th> torch.rand(11):chunk(5)
{
  1 : DoubleTensor - size: 3
  2 : DoubleTensor - size: 2
  3 : DoubleTensor - size: 2
  4 : DoubleTensor - size: 2
  5 : DoubleTensor - size: 2
}

th> torch.rand(10):chunk(7)
{
  1 : DoubleTensor - size: 2
  2 : DoubleTensor - size: 1
  3 : DoubleTensor - size: 2
  4 : DoubleTensor - size: 1
  5 : DoubleTensor - size: 2
  6 : DoubleTensor - size: 1
  7 : DoubleTensor - size: 1
}

th> torch.rand(5):chunk(10)
{
  1 : DoubleTensor - size: 1
  2 : DoubleTensor - empty
  3 : DoubleTensor - size: 1
  4 : DoubleTensor - empty
  5 : DoubleTensor - size: 1
  6 : DoubleTensor - empty
  7 : DoubleTensor - size: 1
  8 : DoubleTensor - empty
  9 : DoubleTensor - size: 1
  10 : DoubleTensor - empty
}
```

### DIFF
--- a/Tensor.lua
+++ b/Tensor.lua
@@ -514,8 +514,20 @@ function Tensor.chunk(result, tensor, nChunk, dim)
       result = {}
    end
    dim = dim or 1
-   local splitSize = math.ceil(tensor:size(dim)/nChunk)
-   return torch.split(result, tensor, splitSize, dim)
+   local n = tensor:size(dim)
+   local lo = 0
+   local z              -- empty tensor variable used when nChunk > tensor:size(dim)
+   for i=1,nChunk do    -- (loop is skipped if nChunk < 1, returning empty table)
+      local hi = math.min(math.ceil(i*n/nChunk), n)
+      if lo < hi then
+        table.insert(result, tensor:narrow(dim, lo+1, hi-lo))
+        lo = hi
+      else
+        z = z or torch.Tensor():typeAs(tensor)
+        table.insert(result, z)
+      end
+   end
+   return result
 end
 torch.chunk = Tensor.chunk
 
--- a/test/test.lua
+++ b/test/test.lua
@@ -3102,7 +3102,7 @@ function torchtest.chunk()
    local result = {}
    local tensor = torch.rand(4,7)
    local nChunk = 3
-   local targetSize = {{4,3},{4,3},{4,1}}
+   local targetSize = {{4,3},{4,2},{4,2}}
    local dim = 2
    local splits = tensor:chunk(nChunk, dim)
    local start = 1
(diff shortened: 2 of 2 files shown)


vivanov879:
works great -- thanks

andreaskoepf:
@nicholas-leonard I think you added `torch.chunk()` originally ... what do you think?

soumith:
@nicholas-leonard can you please see if returning interleaved empty tensors as a behavior is okay? I am leaning towards it being not intuitive.

soumith:
The alternative is to update the docs of torch.chunk to reflect that it's a best-effort chunking, and wont guarantee exactly those many chunks.

nicholas-leonard:
@soumith I think you are right that we should just update the docs.

soumith:
@andreaskoepf what do you think, do you think my POV is valid? I feel like it is super unintuitive to me that i would get a bunch of chunks with empty tensors (especially in between). One option is to go towards what people do in image resizing and aspect ratios. The default is that you put a best-effort-forward wrt chunking, and there's an option in the function signature that forces you to return the exact desired chunks, even if there's empty tensors).

andreaskoepf:
OK, #617 solved as 'by-design'.

PR #620, opened 2016-03-29
Add fmod(), remainder(), remove mod() and fix tensor operator % behavior

liboyue:
To resolve https://github.com/torch/torch7/issues/607, I added `torch.fmod()`, `torch.cfmod()` for computing remainder of division (rounded towards zero) and `torch.remainder()`, `torch.cremainder()` for computing `a % b = a - b * floor(a/b)`.

1. `torch.fmod()` and `torch.cfmod()` are the same as `torch.mod()` and `torch.fmod()`.
2. `torch.remainder()` and `torch.cremainder()` compute remainders by `a - b * floor(a/b)` and it will return `nan` if `b == 0`. The function `remainder()` in the math library computes by `a - b * round(a/b)`. In my laptop, `torch.remainder()` is faster than `torch.fmod()`.
```lua
th> a = torch.Tensor(10000, 10000):uniform(-10, 10);
                                                                      [0.9528s]
th> b = torch.fmod(a, 2);
                                                                      [1.0508s]
th> c = torch.remainder(a, 2);
                                                                      [0.5180s]
```

3. The tensor operator uses `THTensor_(mod)` as its backend. To make sure it's the same as Lua `%` operator I changed it with `THTensor_(remainder)`.

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -311,8 +311,9 @@ for _,Tensor in ipairs({"ByteTensor", "CharTensor",
          {name=Tensor, method={default=1}},
          {name=real}})
  
+   -- mod alias
    wrap("mod",
-        cname("mod"),
+        cname("fmod"),
         {{name=Tensor, default=true, returned=true, method={default='nil'}},
          {name=Tensor, method={default=1}},
          {name=real}})
@@ -363,8 +364,9 @@ for _,Tensor in ipairs({"ByteTensor", "CharTensor",
          {name=Tensor, method={default=1}},
          {name=Tensor}})
 
+   -- cmod alias
    wrap("cmod",
-        cname("cmod"),
+        cname("cfmod"),
         {{name=Tensor, default=true, returned=true, method={default='nil'}},
          {name=Tensor, method={default=1}},
          {name=Tensor}})
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -299,6 +299,12 @@ for _,Tensor in ipairs({"ByteTensor", "CharTensor",
          {name=Tensor, method={default=1}},
          {name=real}})
 
+   wrap("fmod",
+        cname("fmod"),
+        {{name=Tensor, default=true, returned=true, method={default='nil'}},
+         {name=Tensor, method={default=1}},
+         {name=real}})
+
    wrap("mod",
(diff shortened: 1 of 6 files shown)


liboyue:
@andreaskoepf Is this PR better?

andreaskoepf:
@liboyue `remainder` reminds me very much of the C [remainder()](http://www.cplusplus.com/reference/cmath/remainder/) function which implements the [IEEE 754-1985](https://en.wikipedia.org/wiki/IEEE_754-1985#Standard_operations) remainder standard operation. The standard `remainder()` rounds towards the *nearest* integer. From Wikipedia:

> Floatingpoint remainder. This is not like a normal modulo operation, it can be negative for two positive numbers. It returns the exact value of x–(round(x/y)·y).
 
I think it would be confusing to introduce a remainder function with a different meaning. Also I am not really happy with introducing an alias and declaring mod as deprecated...

There are three different [modulo-styles](https://en.wikipedia.org/wiki/Modulo_operation):

- `r = a - b * trunc(a/b)` (truncated)
- `r = a - b * floor(a/b)` (floored)
- `r=a - b * round(a/b)` (Euclidean)

If we really want to offer all three flavours I wonder whether it would be better to have a single mod() function with a default behaviour (e.g. truncated/fmod style) and an optional mode-argument (similar to the mode of `torch.eig`) which would allow to switch to floored/Euclidean mode. This would be also fully backward compatible.

Maybe @soumith or @andresy could comment on this?

liboyue:
@andreaskoepf In fact, I am confused by the names too... 
How about this, just offer `torch.fmod()` and `%` operator. I don't think we need all the four mod, when people need other mods, they can implement it in Lua. 
But It's important to clarify that `torch.mod()` is different with `%`, so in my opinion, we should mark it as deprecated...

andreaskoepf:
Since C does not have a % operator for floating point values I would be fine with defining it as floored-mod for tensors.

liboyue:
Yes.
```lua
th> (-1) % 2
1
th> torch.Tensor({-1}) % 2
-1
[torch.DoubleTensor of size 1]
```
`%` should have the same behavior.

soumith:
Thanks a lot 李博约!

PR #632, opened 2016-04-07
Update maths.md

lejlot:
Current implementation of .eig and .symeig returns eigenvalues such that `A = V diag(e) V'`, not `A = V' diag(e) V`.


soumith:
Thanks!

PR #634, opened 2016-04-10
Make THInf type-specific

tkoeppe:
I have not tested this change with libraries other than `torch7` itself. From a search though our internal codebase, I believe that whatever code does break (e.g. in `cunn`) is actually incorrect in its current form and needs to be fixed (usually something like `float x = -THInf;`, which needs to become `float x = -FLT_MAX;`).

What do you think?

### DIFF
--- a/lib/TH/THGeneral.h.in
+++ b/lib/TH/THGeneral.h.in
@@ -31,8 +31,6 @@
 # define TH_API TH_EXTERNC
 #endif
 
-#define THInf DBL_MAX
-
 #define TH_INLINE @TH_INLINE@
 
 #ifndef __cplusplus
--- a/lib/TH/THGenerateAllTypes.h
+++ b/lib/TH/THGenerateAllTypes.h
@@ -5,6 +5,7 @@
 #define real unsigned char
 #define accreal long
 #define Real Byte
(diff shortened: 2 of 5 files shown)


tkoeppe:
@soumith, @dominikgrewe: I'd be very grateful for your advice and guidance on these PRs.

noa:
I think this broke cutorch:

/tmp/luarocks_cutorch-scm-1-9578/cutorch/lib/THC/THCTensorMath.cu(280): error: identifier "THInf" is undefined

/tmp/luarocks_cutorch-scm-1-9578/cutorch/lib/THC/THCTensorMath.cu(295): error: identifier "THInf" is undefined

2 errors detected in the compilation of "/tmp/tmpxft_00007690_00000000-7_THCTensorMath.cpp1.ii".
CMake Error at THC_generated_THCTensorMath.cu.o.cmake:264 (message):
  Error generating file
  /tmp/luarocks_cutorch-scm-1-9578/cutorch/build/lib/THC/CMakeFiles/THC.dir//./THC_generated_THCTensorMath.cu.o

soumith:
Should be fixed now.

noa:
@soumith thanks

PR #642, opened 2016-04-14
get rid of extra NULL character (serialize)

deltheil:
This extra slot comes from [here](https://github.com/torch/torch7/blob/814836bb3115fa0a0b3f3d6982122e08f9f6530d/lib/TH/THMemoryFile.c#L230). IMHO we should get rid of it in the context of `serialize` / `serializeToStorage` so that archives are identical to those produced by `save`.

### DIFF
--- a/File.lua
+++ b/File.lua
@@ -424,6 +424,8 @@ function torch.serializeToStorage(object, mode)
    f = f[mode](f)
    f:writeObject(object)
    local storage = f:storage()
+   -- the storage includes an extra NULL character: get rid of it
+   storage:resize(storage:size()-1)
    f:close()
    return storage
 end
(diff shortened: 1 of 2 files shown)


soumith:
i assume this poses no problems for loading serialized files, as this is a different codepath

deltheil:
Yes. Also using the current `deserializeFromStorage` with or without this extra slot is transparent, e.g. (before this fix):

```Lua
> x = torch.serializeToStorage("foo")
> print(torch.deserializeFromStorage(x))
foo
> x:resize(x:size()-1)
> print(torch.deserializeFromStorage(x))
foo
```

I was just wondering why the deserialization part adds an extra slot ([here](https://github.com/torch/torch7/blob/814836bb3115fa0a0b3f3d6982122e08f9f6530d/File.lua#L437)) but it's another question.

soumith:
Thanks Cedric.

PR #647, opened 2016-04-18
Check for overflows when converting pointers in luaT

apaszke:
`luaT_lua_pointer` didn't check for overflows when converting pointers, and it used a signed long type, that could overflow and stay negative even after a cast to double. This addresses torch/threads#55.


apaszke:
There are integers greater than 2^53 that are representable by doubles, but I don't really know how to check if we're dealing with one at the moment, so I made it a hard threshold.

PR #650, opened 2016-04-19
Make torch.Generator serializable

kosklain:
This implements read/write naively as a block of bytes for a given torch.Generator (in the style of set/getRNGstate, which use memcpy).

### DIFF
--- a/test/test.lua
+++ b/test/test.lua
@@ -2483,6 +2483,20 @@ function torchtest.RNGStateAliasing()
     mytester:assertTensorEq(target_value, forked_value, 1e-16, "RNG has not forked correctly.")
 end
 
+function torchtest.serializeGenerator()
+   local generator = torch.Generator()
+   torch.manualSeed(generator, 123)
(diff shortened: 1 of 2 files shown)


test/test.lua
```diff
@@ -2483,6 +2483,20 @@ function torchtest.RNGStateAliasing()
     mytester:assertTensorEq(target_value, forked_value, 1e-16, "RNG has not forked correctly.")
 end
 
+function torchtest.serializeGenerator()
+   torch.manualSeed(123)
+   local generator = torch.Generator()
+   torch.manualSeed(124)
```
timharley: Both of these manualSeed calls operate on the global generator, not these local ones. These generators will be non-deterministically initialised with a [seed from the clock](https://github.com/torch/torch7/blob/master/lib/TH/THRandom.c#L69).

Do this instead:
`torch.manualSeed(generator, 123)`

kosklain:
Thanks for the comment, done.

soumith:
Thanks!

PR #652, opened 2016-04-19
Fix pointer to long conversions in luaT

apaszke:
I came across more segfaults when debugging threads. Apparently there's more of this conversion stuff 😕


soumith:
Thanks!

PR #653, opened 2016-04-19
allow THAllocator to have a NULL realloc

soumith:
This is a pre-diff that's needed to fix cutorch's THCudaHostAllocator_realloc which at the moment has a bug where it does not copy the old memory over to the new memory.

https://github.com/torch/cutorch/blob/master/lib/THC/THCAllocator.c#L21-L31

cc: @tudor


soumith:
tested via locally setting the default allocator's realloc to NULL and running tests.
Reviewed by @tudor

PR #657, opened 2016-04-25
Remove some warning when compiling TH

albanD:
These were introduce when adding these two features to TH: #13 and #213


soumith:
Thanks!

PR #660, opened 2016-04-26
Fixed the Visual Studio CMake build.

Ark-kun:
* Added missing CMake function include.
* snprintf returns int, not ssize_t. ssize_t is not standard C++.
* Added arguments for the variadic macro parameters. (We should probably remove the variadic parameter from the macro as it's never used)
* Avoided preprocessor directives inside macro arguments.
* Not linking to LibM when building under MSVC as the math.h implementations are already available in the msvcrt*.dll. (Does torch need the LibM or just some math.h implementation?)


lib/TH/THMemoryFile.c
```diff
@@ -177,7 +177,7 @@ static int THMemoryFile_mode(const char *mode, int *isReadable, int *isWritable)
       size_t i;                                                           \
       for(i = 0; i < n; i++)                                            \
       {                                                                 \
-        ssize_t nByteWritten;                                           \
+        int nByteWritten;                                               \
```
albanD: For the `Byte` and `Char` types, `nByteWritten` can be affected the value of `n` which is of type `size_t`. Can't this overflow an `int`?

Ark-kun: This was the case before the change: 'ssize_t' is often defined as 'int' while 'size_t' can be 'unsigned long int' in the same header. The 'ssize_t' is not defined in the C++ standard, so it's not easy to reason about its size. I could use the 'intmax_t' type, but it might be slower.

albanD: I was asking because on my x86_64 ubuntu, `ssize_t` is actually a `long`.
I am really not sure what should be done here since this variable can be assigned either an `int` or a `size_t`.

soumith: i think this should def be long.

Ark-kun: I've fixed the PR.

soumith:
fixed via #694

Issue #671, opened 2016-05-03
baddbmm with not contiguous res tensor


JeremyMorvan:
Hi all,

I've just encountered an bug with the baddbmm function. It seems to be due to the non-contiguousity of the result tensor. Here is how to reproduce it : 

```
a = torch.randn(1,3,2)
b = torch.randn(1,2,6)
c = torch.randn(1,3,6,2):select(4,1)
cc = c:clone()

print((c-cc):abs():max())
c:baddbmm(a,b)
cc:baddbmm(a,b)
print((c-cc):abs():max())
```

Same thing happens in Cuda.

Thank you,
Jérémy

akashmehra:
You probably need to use index (followed by a reshape) instead of select (select doesn't change the stride). index will give you a new tensor though and will be slow.

JeremyMorvan:
Yes, I'll find another way to do what I want to do (probably slower) but I posted this message because I don't think this is due to my usage of baddbmm : 
This looks like an operation that does not give the same result for two tensors that have the same content and the same size.

fmassa:
This is indeed a bug. I'm fixing it. Thanks for the bug report !

PR #672, opened 2016-05-04
Fix addmm for non-contiguous result tensor

fmassa:
`addmm` didn't support properly non-contiguous result tensors.
The goal of this part of the code is to copy the contents of `r_` into `r__` such that `r__` contains the transpose of `r_` contiguous in memory.
The way it was doing was by considering `c_inv = torch.Tensor(c:size(2),c:size(1)):copy(c)`, which doesn't transpose the tensor but only change its shape.

This problem was introduced in https://github.com/torch/torch7/commit/0d534e64f5569fea2992a2754f0b87e27b3c1d8c .

I'll add more unit tests later if you want.

Should fix https://github.com/torch/torch7/issues/671

Also, something similar should be added to `cutorch`.

### DIFF
--- a/lib/TH/generic/THTensorMath.c
+++ b/lib/TH/generic/THTensorMath.c
@@ -808,8 +808,8 @@ void THTensor_(addmm)(THTensor *r_, real beta, THTensor *t, real alpha, THTensor
   {
     transpose_r = 'n';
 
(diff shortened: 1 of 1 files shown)


soumith:
Thanks Francisco!

JeremyMorvan:
Thank you !

PR #673, opened 2016-05-04
[TensorOperator.c] Make unary negation expression have real type, not double

tkoeppe:
Before this change, the expression `-luaL_checknumber(L, ...)` has type `double`, and its value may not be representable as `real`, leading to undefined behaviour. After this change, the unary negation is applied to a value of type `real`, and the arithmetic promotion and conversion rules give this well-defined behaviour.


soumith:
Thanks!

PR #675, opened 2016-05-06
Fix memory leak in addmm

fmassa:
I'm really sorry about that, but I think I introduced a memory leak in https://github.com/torch/torch7/pull/672
This should address the leak, but please someone double check in case I'm missing something.


soumith:
Thanks Francisco!

PR #676, opened 2016-05-07
Add function to get the amount of memory used by torch tensors

fmassa:
Allows to get the amount of memory used by torch tensors.

In conjunction with `collectgarbage('count')`, we can have some good estimates of the total amount of memory used (except if other data structures are used, like `tds`).

**Note**: as `heapSize` is only updated by batches of ~ 1MB, so creating really small tensors won't show any immediate difference.

I will add docs if we agree to merge this.

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -227,6 +227,12 @@ static int torch_updateerrorhandlers(lua_State *L)
   return 0;
 }
 
+static int torch_getmemory(lua_State *L)
+{
+  lua_pushinteger(L, THGetHeapSize());
+  return 1;
+}
+
 static const struct luaL_Reg torch_utils__ [] = {
   {"getdefaulttensortype", torch_lua_getdefaulttensortype},
   {"isatty", torch_isatty},
@@ -250,6 +256,7 @@ static const struct luaL_Reg torch_utils__ [] = {
   {"pointer", luaT_lua_pointer},
   {"setheaptracking", torch_setheaptracking},
   {"updateerrorhandlers", torch_updateerrorhandlers},
+  {"getmemory", torch_getmemory},
   {NULL, NULL}
 };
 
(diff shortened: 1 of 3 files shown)


utils.c
```diff
@@ -250,6 +256,7 @@ static const struct luaL_Reg torch_utils__ [] = {
   {"pointer", luaT_lua_pointer},
   {"setheaptracking", torch_setheaptracking},
   {"updateerrorhandlers", torch_updateerrorhandlers},
+  {"getmemory", torch_getmemory},
```
apaszke: How about `memoryused`? It might be more self-explanatory.

fmassa: I agree that simply `getmemory` is not a good name.
`memoryused` is indeed better.
Any other opinions ? `cutorch` has a `getMemoryUsage` (which prints the total memory available and the memory used for the GPU). Maybe it would be good to have some uniformity in there.

fmassa:
Also, I could add `heapDelta` to `heapSize` to have a better estimate of the count of memory.

utils.c
apaszke: `getMemoryUsage` sounds good to me as well 😄 I agree that it might be better to keep it similar.

fmassa: but there is a difference between `getMemoryUsage` and this PR in the sense that this PR aims at showing the memory used by torch, and not the total memory used. BTW, the same PR could be done in `cutorch`, in which case we would have different functions which behave differently but with a similar purpose.

apaszke: Hm, ok, you're right. Do you have any suggestions for the name? `getTorchMemory[Usage]`?

soumith: torch.getHeapSize ?

apaszke: Sounds a bit as if it would return Lua's heap size as well, but I like it.

colesbury: torch.tensormemoryusage_if_youre_not_using_too_many_threads_plusOrMinus100MB()

PR #680, opened 2016-05-13
added THAtomicSetLong

lake4790k:
Added the missing long version for atomic set. I'm adding an atomic counter in `tds` and adding support for that in `threads`, this is needed for that.


soumith:
Thanks!

PR #686, opened 2016-05-18
torch.Tester early abort and disable: make them work when used together

davidsaxton:
Previously: if early abort was enabled, then it would abort on any disabled tests.


soumith:
Thanks david!

PR #694, opened 2016-05-31
fix compilation issues under Windows

elikosan:
With these fixes i can install torch7 with luarocks on Windows.
Compiler: Visual Studio 2015
OS: Windows 7 Pro


fmassa:
Thanks for the fixes!
This seems quite similar to https://github.com/torch/torch7/pull/660 though.
@soumith maybe its time to merge one of the PRs ?

lib/TH/THMemoryFile.c
```diff
@@ -479,7 +479,7 @@ static size_t THMemoryFile_writeLong(THFile *self, long *data, size_t n)
     size_t i;
     for(i = 0; i < n; i++)
     {
-      ssize_t nByteWritten;
+      size_t nByteWritten;
```
soumith: changing the ssize_t to size_t is likely breaking the contbuilds. See https://travis-ci.org/torch/torch7/builds/134239495

soumith:
This looks better than #660 but fails contbuilds. made in-line comments.

elikosan:
I just committed a fix to the ssize_t issue. Let's see if Travis likes it better...

soumith:
Thanks! :)

PR #695, opened 2016-06-04
fix "macro redefined" warning (th_isnan)

deltheil:
See #694


soumith:
thanks, was about to do it myself.

PR #720, opened 2016-07-20
Make TH more usable and consistent

apaszke:


soumith:
LGTM! thanks!

ghostcow:
what does this mean for everyday usage?

fmassa:
You can more easily use it in standalone C applications, because you don't need to convert the returned indices from 1-based to 0-based at every function call.
But if you only use the lua interface, it doesn't change anything.
BTW, @apaszke , do you plan do do the same for `cutorch`?

Issue #718, opened 2016-07-15
torch.histc buggy implementation for float


codeAC29:
`torch.histc` works incorrectly on float tensors. Here is a working example reproducing the bug:
```lua
x = torch.Tensor(3, 2):uniform():mul(10):add(1):floor()
x[1][1] = 38
x:histc(38)[38]           -- Correct
x:float():histc(38)[38]   -- Wrong
```

Workaround:

```lua
x:float():histc(39, 1, 39)[38]  -- Correct
```

andreaskoepf:
I had a view at the code. The problem as in [THTensorMath.c#L2489](https://github.com/torch/torch7/blob/58b9acd72bc39b3ea3b0426fc52fd28ffb0d7f97/lib/TH/generic/THTensorMath.c#L2489) which assumes that a subtraction of a float value results in something smaller than the original value. Unfortunately `38 - 1e-6` with single precision floating point (23 bit mantissa) simply stays 38.

PR #723, opened 2016-07-25
Fix torch.histc problem #718

andreaskoepf:
No longer cloning input tensor nor using tensor operations but computing destination bin index directly in TH_TENSOR_APPLY() block and ensuring that bin index does not exceed nbins-1.


soumith:
Thanks a lot Andreas!

PR #733, opened 2016-08-08
cmake: add soversion for luaT

CDLuminate:
Imported from Debian package.

https://anonscm.debian.org/cgit/debian-science/packages/lua-torch-torch7.git/tree/debian/patches/luaT-cmake-add-version

PR #738, opened 2016-08-11
cmake: add soversion for libTH

CDLuminate:
Imported from Debian package.


CDLuminate:
This is similar to https://github.com/torch/torch7/pull/733

Original patch can be found here

https://anonscm.debian.org/cgit/debian-science/packages/lua-torch-torch7.git/tree/debian/patches/TH-cmake-add-version

luaT and TH both have soversion now.

CDLuminate:
Adding soversion to these shared libraries brings convenience to distributions.

soumith:
Thanks!

PR #745, opened 2016-08-23
clean up equations in documentation

temerick:
No new content. Just standardizing the formatting of the numbers and equations in the random docs.


soumith:
Thanks!

Issue #641, opened 2016-04-13
Back-compatibility deserialisation issue


Atcold:
Here there is a model I've trained some time ago.

https://engineering.purdue.edu/elab/smartEYE-data/net/17cate9filter.tar.gz

With the latest version of `torch` / `cunn` the object is no longer deserialisable.
This is awful. I can no longer access my previously trained models.

Reverting back `torch` / `cunn` (and many more packages) to, say, 1 year ago, fixes the problem.

```lua
th> require 'cunn';
                                                                      [0.9739s]	
th> net = torch.load('model-127.net')
/home/atcold/torch/install/share/lua/5.1/torch/File.lua:301: Failed to load function from
 bytecode: (binary): cannot load incompatible bytecode
stack traceback:
	[C]: in function 'error'
	/home/atcold/torch/install/share/lua/5.1/torch/File.lua:301: in function 'readObject'
	/home/atcold/torch/install/share/lua/5.1/torch/File.lua:369: in function 'readObject'
	/home/atcold/torch/install/share/lua/5.1/nn/Module.lua:154: in function 'read'
	/home/atcold/torch/install/share/lua/5.1/torch/File.lua:351: in function 'readObject'
	/home/atcold/torch/install/share/lua/5.1/torch/File.lua:369: in function 'readObject'
	/home/atcold/torch/install/share/lua/5.1/torch/File.lua:369: in function 'readObject'
	/home/atcold/torch/install/share/lua/5.1/nn/Module.lua:154: in function 'read'
	/home/atcold/torch/install/share/lua/5.1/torch/File.lua:351: in function 'readObject'
	/home/atcold/torch/install/share/lua/5.1/torch/File.lua:369: in function 'readObject'
	/home/atcold/torch/install/share/lua/5.1/torch/File.lua:369: in function 'readObject'
	/home/atcold/torch/install/share/lua/5.1/nn/Module.lua:154: in function 'read'
	/home/atcold/torch/install/share/lua/5.1/torch/File.lua:351: in function 'readObject'
	/home/atcold/torch/install/share/lua/5.1/torch/File.lua:409: in function 'load'
	[string "net = torch.load('model-127.net')"]:1: in main chunk
	[C]: in function 'xpcall'
	/home/atcold/torch/install/share/lua/5.1/trepl/init.lua:669: in function 'repl'
	...cold/torch/install/lib/luarocks/rocks/trepl/scm-1/bin/th:199: in main chunk
	[C]: at 0x00406670	
[...]

Issue #741, opened 2016-08-18
Compatibility issue while loading data in Torch


aparnamhnty:
I am new to Torch.....I want to implement the code given at http://torch.ch/blog/2015/07/30/cifar.html
While trying to execute train.lua........it initially shows the parameters and configures the model, but while loading the data, it complains that 'Failed to load function from bytecode: (binary): cannot load incompatible bytecode.
What shall I do to solve this problem.
![capture](https://cloud.githubusercontent.com/assets/9334941/17765583/69021fe8-6544-11e6-888f-ac8ead36c5bb.PNG)

Atcold:
You already posted this error here https://github.com/torch/torch7/issues/641
Why are you making duplicates?

PR #748, opened 2016-08-24
changing function load errors to warnings

soumith:
This seems like a needed change, now that we have LuaJIT 2.1, 2.0, Lua 5.2 and soon Lua 5.3 used by torch users.

The models which are serialized usually have no functions, but occasionally have some functions which usually are not needed.

Fixes https://github.com/torch/torch7/issues/641 and #741

### DIFF
--- a/File.lua
+++ b/File.lua
@@ -275,7 +275,7 @@ function File:readObject()
        local dumped = self:readChar(size):string()
        local func, err = loadstring(dumped)
        if not func then
-          error(string.format('Failed to load function from bytecode: %s', err))
+          io.stderr:write(string.format('Warning: Failed to load function from bytecode: %s', err))
        end
        local upvalues = self:readObject()
        for index,upvalue in ipairs(upvalues) do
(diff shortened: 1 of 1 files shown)


soumith:
@szagoruyko @fmassa what do you think?

fmassa:
In general, I'm ok with this change.

But this might also lead to subtle bugs in some cases.
For example, in `optnet`, I overwrite the backward pass with a function which raises an error if the network was optimized for inference (to avoid computing wrong gradients). With this change, a network saved in `inference` mode could potentially be deserialized without this `error` function, leading to completely wrong gradients without any note or warning that the gradients might be wrong.

Maybe the best would be to change `optnet` to avoid overwriting functions as I'm currently doing, but I'm not yet sure how to do it in a robust manner.

soumith:
when you load the model it does print a warning. I could make the warning to be more scary. "Warning: bad things might happen!!!!!!!!! :-# :-#"

fmassa:
Yes, it does print a warning. No need to scare people that much I think :)

I think I'll add an error while serializing models with optimization on, forcing the user to remove the optimizations before saving.

PR #755, opened 2016-09-04
Added dyanamic dispatch for x86

rguthrie3:
This should allow distribution of binaries which will use SSE instructions for THVector operations when available on the host hardware, and back off to a default implementation otherwise.  The following bullets summarize the changes:

* It looks like the vector functions were exported in the header TH.h publicly as inline functions or macros.  obviously the dynamic dispatch functions can't be inlined, so I changed THVector.h to simply include a generic/THVector.h header which provides declarations for all of the functions with all of the data types to be exported publicly (this might not be necessary if you don't care about exporting them publicly).  These will still be included as macros / inlines if compiled for ARM.

* I renamed the SSE functions to, e.g THVector_(fill_SSE) and the scalar versions to THVector_(fill_SCALAR) (which are still in generic/THVector.c).  The SSE ones are in a file I named THVectorImpls.h.  I also wrapped the macros with functions (so that I can take the functions address).

* generic/THVectorDispatch.c defines the function pointers that point to the most optimized function available on the host, the tables describing what implementations are available, and the code for initializing these function pointers (which needs to be called by the user).  It also has the entry points for client code calling THVector functions (which just calls the pointer to the actual implementation).

* With all that above, THTensor.c just includes THVectorImpls.h (which generates all SSE and SCALAR functions), then generates generic/THVectorDispatch.c, which uses the definitions of the SSE and SCALAR functions.


apaszke:
Is there any reason for keeping the SSE and SCALAR function definitions in header files instead of `.c`? The whole point of having automatic dispatch is to figure out the best implementation available automatically, instead of hardcoding it. What's more, they will have to be compiled multiple times if used in several files. Also, it results in this weird pattern that `THTensor.c` includes `THVectorImpl.h` that includes `generic/THVector.c`. I'm not sure if it's *that* important to keep them as inlines (especially that they are no longer inlines on x86).

IMHO it would be better to use automatic dispatch for all architectures. Separating x86 and ARM results in quite weird conditional includes and requires changing the outside code (`vectorDispatchInit` is undefined on ARM). You could just generate all of the implementations that can be generated, and use preprocessor conditionals to include or exclude them from dispatch tables. `SCALAR` versions would be always defined. This could greatly simplify the code.

lib/TH/THVectorImpls.h
```diff
@@ -0,0 +1,602 @@
...
+            :                                                           \
+            :"r" (y),"r" (x), "r"(caddr),"r"(n)                         \
+            : "cc", "r0", "r1", "r2", "r5", "memory",                   \
+              "q0", "q1", "q2", "q3", "q14",                            \
+              "d0", "d1", "d2", "d3", "d4", "d5", "d6", "d7",           \
+              "d16", "d17", "d18", "d19", "d20", "d21", "d22", "d23", "d28", "d29" \
+            );                                                          \
+    }
+
+static inline void THDoubleVector_fill(double *x, const double c, const long n) {
```
apaszke: This is just code copied from `generic/THVector.c`. Adding NEON to dispatch tables would eliminate this duplication. If we don't choose to go this way, it would be better to generate them just for double from `generic/THVector.c` anyway.

lib/TH/generic/THVectorDispatch.c
```diff
@@ -0,0 +1,99 @@
...
+
+static void (*THVector_(fill_DISPATCHPTR))(real *, const real, const long) = &THVector_(fill_SCALAR);
+static FunctionDescription THVector_(fill_DISPATCHTABLE)[] = {
+#if defined(TH_REAL_IS_DOUBLE) || defined(TH_REAL_IS_FLOAT)
+  FUNCTION_IMPL(THVector_(fill_SSE), SIMDExtension_SSE),
+#endif
+  FUNCTION_IMPL(THVector_(fill_SCALAR), SIMDExtension_SCALAR)
+};
+void THVector_(fill)(real *x, const real c, const long n) {
+  assert(THVector_(fill_DISPATCHPTR));
```
apaszke: Are these asserts really needed? You explicitly initialize the pointers a couple lines above. I think it makes sense to do any of the two (with 1. being a bit better I think):
1. Don't initialize the pointer explicitly, leave the assert - this allows to catch errors if someone forgets to add `INIT_DISPATCH_PTR` below.
2. Initialize the pointer explicitly, remove the assert - it won't be triggered anyway.

rguthrie3:
Good points.  The reason I did not implement anything for ARM dynamic dispatch yet is that I don't have any ARM experience or hardware (although I could set up an environment for simulation).  The obnoxious conditional compilation for ARM was only intended to be temporary, as I believe Soumith said he was facing a deadline that needed dynamic dispatch code in order to distribute binaries.  ARM dynamic dispatch would not be hard to add on to the code, provided I can set up my environment to test it.

I also agree that moving all implementations to a .c file is fine.  There is no good reason to keep them in .h files since they cannot be inlined.

I will move the impls around tonight and push a change.  I will also try and get ARM functions in the dispatch tables.  If there are no problems testing it, I can push it tonight as well.

Thanks!

apaszke:
Also, since the implementations can be moved to .c files, there's no longer any reason to put the code in macros.

I have a computer with an ARM CPU supporting NEON, so I can try to test your new changes.

rguthrie3:
I pushed some changes.  This code will probably not build on ARM, but I can do the dynamic dispatch for ARM tomorrow most likely.

*  I renamed the _SCALAR functions to _DEFAULT.  THVectorScalar sounds very confusing.
* I put declarations for all implementations in the header THVectorImpls.h, which is included by THVectorDispatch.c (which just generates all the types for generic/THVectorDispatch.c, so this file is compiled separately of THTensor.c now).
* SIMD impls are in THVectorSIMD.c, and default impls are generics in a file I renamed generic/THVectorDefault.c.  Similar for the generic header that declares the default impls.
* I made the pointers NULL and put the asserts there.  I originally was initializing them to NULL and had the asserts for that reason, then changed to initializing them to the default, but I think the bug catching will be better by just leaving the asserts and initializing to NULL

Let me know what you think.  I will get ARM dynamic dispatch implemented as soon as possible.

lib/TH/THTensor.c
```diff
@@ -1,6 +1,11 @@
 #include "THAtomic.h"
 #include "THTensor.h"
+
+#include "THVectorSIMD.c"
```
apaszke: I believe these includes have nothing to do with tensor logic and are better off in `THVectorDispatch.c`. Maybe it'd be better to rename it to `THVector.c` and compile both the dispatch logic and implementations together.

apaszke:
This looks much better to me. The only things I'd change would be to compile the vector code separately from tensors and isolate NEON and SSE implementations into different files e.g. `vector/NEON.c` and `vector/SSE.c`.

So:
* `THVectorDispatch.c` -> `THVector.c` (including arch-specific implementations definitions)
* `THVectorSIMD.c` -> divide into multiple files, each for one SIMD extension
* All generic files look good to me.

After adding the ARM dispatch it should be much cleaner. Thanks!

rguthrie3:
I pushed changes that add the NEON functions to the dispatch tables conditioning on the __NEON__ macro,  and factored the vector functions into their own files.  I also got rid of the weird conditional compilation of the default implementations, so ARM NEON should be "dynamic dispatched", which I put quotes around because this still depends on determining the availability of NEON at compile time.  Should there be an ARM macro that is defined whether or not the host supports NEON so that this choice can be deferred until runtime?

apaszke:
I see one last problem. I don't think the checking for `__NEON__` and otherwise assuming you're on x86 is a good idea. TH tries to be quite portable. It would be better to use the `USE_SSE*` macros for this reason both in THVector.h and in THVectorDispatch.c (and omit the else clause - just compile whatever you can):
```C
#ifdef __NEON__
#include "vector/NEON.c"
#endif

#if defined USE_SSE2 || defined USE_SSE3 || defined USE_SSSE3 \
  || defined USE_SSE4_1 || defined USE_SSE4_2
#include "vector/SSE.c"
#endif
```

Also, a smal nit: `NEON` functions are `static`, `SSE` functions are not. It would be better to stay consistent.

rguthrie3:
I just pushed both changes you requested.  Let me know if there is anything else

apaszke:
The code looks good to me 👍
But before the merge you have to make the tests pass, apparently they [all segfault on the same test](https://github.com/torch/torch7/blob/master/test/test_Tester.lua#L213-L229). There must be some small bug somewhere.

rguthrie3:
Oh, the reason for that is the null pointer initialization of the dispatch pointers.  I can fix this probably by just adding the correct vectorDispatchInit() call somewhere (or just changing the pointers back to being initialized to the default implementation)

apaszke:
Well, merging the PR should never cause segfaults. It's better to add `vectorDispatchInit()`  [here](https://github.com/torch/torch7/blob/master/generic/Tensor.c#L1315).

soumith:
@rguthrie3 i think it's better to make the dispatch pointers be initialized to the default implementation, and add a call to `THVector_(vectorDispatchInit)()`  [here](https://github.com/torch/torch7/blob/master/generic/Tensor.c#L1315).

rguthrie3:
Yes sorry about that, I wasn't sure where to add the init call.  I have added it, and also changed the dispatch pointer initialization, so I think that all looks good now.

soumith:
Thanks Robert! This is really great.

lib/TH/generic/simd/simd.h
```diff
@@ -0,0 +1,91 @@
...
+  return SIMDExtension_NEON;
+}
+
+#else // x86
+
+static inline void cpuid(uint32_t *eax, uint32_t *ebx, uint32_t *ecx, uint32_t *edx)
+{
+  uint32_t a = *eax, b, c, d;
+  asm volatile ( "cpuid\n\t"
+                 : "+a"(a), "=b"(b), "+c"(c), "=d"(d) );
```
BTNC: According to the context (cupid is called in detectHostSIMDExtensions), ecx is only used as input, and local variable c is not initialized. Thus I think "+c" should be "=c". ps: below is my change in order to compile with msvc:
--- a/lib/TH/generic/simd/simd.h
+++ b/lib/TH/generic/simd/simd.h
@@ -2,6 +2,9 @@
 #define TH_SIMD_INC

 #include <stdint.h>
+#ifdef _MSC_VER
+#include <intrin.h>
+#endif

 // Can be found on Intel ISA Reference for CPUID
 #define CPUID_AVX2_BIT 0x10       // Bit 5 of EBX for EAX=0x7
@@ -56,13 +59,23 @@ static inline uint32_t detectHostSIMDExtensions()

 static inline void cpuid(uint32_t *eax, uint32_t *ebx, uint32_t *ecx, uint32_t *edx)
 {
+#ifndef _MSC_VER
   uint32_t a = *eax, b, c, d;
   asm volatile ( "cpuid\n\t"
-                 : "+a"(a), "=b"(b), "+c"(c), "=d"(d) );
+                 : "+a"(a), "=b"(b), "=c"(c), "=d"(d) );
   *eax = a;
   *ebx = b;
   *ecx = c;
   *edx = d;
+#else
+  // above asm code is some how misleading
+  uint32_t cpuInfo[4];
+  __cpuid(cpuInfo, *eax);
+  *eax = cpuInfo[0];
+  *ebx = cpuInfo[1];
+  *ecx = cpuInfo[2];
+  *edx = cpuInfo[3];
+#endif
 }

 static inline uint32_t detectHostSIMDExtensions()
diff --git a/lib/TH/vector/SSE.c b/lib/TH/vector/SSE.c
index f909907..c47e28d 100644
--- a/lib/TH/vector/SSE.c
+++ b/lib/TH/vector/SSE.c
@@ -1,4 +1,8 @@
+#ifndef _MSC_VER
 #include <x86intrin.h>
+#else
+#include <intrin.h>
+#endif


 static void THDoubleVector_fill_SSE(double *x, const double c, const long n) {

aybassiouny: Same problem here with msvc. Any plans for a pr?

PR #758, opened 2016-09-08
Various improvements to THMapAllocator

apaszke:
* Fixed segfault when calling THError
* Improved `shm_open` and `shm_unlink` cmake feature tests
* Created `incref`, `decref` functions for refcounted shared memory objects
* Fixed an OS X bug in THMapAllocator (writing to fd to shm object is not supported)


soumith:
Thanks!

lib/TH/CMakeLists.txt
```diff
@@ -70,28 +70,6 @@ IF (CORTEXA9_FOUND)
   SET(CMAKE_C_FLAGS "-mcpu=cortex-a9 ${CMAKE_C_FLAGS}")
 ENDIF (CORTEXA9_FOUND)
 
-IF(UNIX)
-  INCLUDE(CheckFunctionExists)
```
CDLuminate: FindBlas needs this include. This caused build failure on my side. I'll fix it later.

CDLuminate: https://github.com/torch/torch7/pull/761

PR #768, opened 2016-09-19
fix cpuid ecx; change to compile with msvc; compile with MKL on windows

BTNC:
Hi,

This small change contains three parts:
1. fix one line cupid code from "+c" to "=c"
2. change master branch to compile with msvc
3. compile against mkl on windows with msvc

This change only affect those who use msvc, please let me know if any change is improper or can be improved.

Thanks


soumith:
Thank you so much for this patch!

PR #774, opened 2016-09-26
bug fix for read/writeLong in THMemoryFile and 2 test files change for windows

BTNC:
Hi,

There is a bug in THMemoryFile read/writelong for following cases:
(1) sizeof(long) is 8 while longSize is set to 4;
(2) sizeof(long) is 4 while longSize is set to 8;

This change comes along with 2 test files change for windows.

Thanks,


lib/TH/THMemoryFile.c
```diff
@@ -383,12 +384,12 @@ static size_t THMemoryFile_readLong(THFile *self, long *data, size_t n)
     {
       int i, big_endian = !THDiskFile_isLittleEndianCPU();
       size_t nByte = 8*n;
-      long *storage = (long *)(mfself->storage->data + mfself->position);
+      int32_t *storage = (int32_t *)(mfself->storage->data + mfself->position);
```
soumith: i believe this has to be int64_t, if i'm not mistaken

lib/TH/THMemoryFile.c
```diff
@@ -459,8 +460,8 @@ static size_t THMemoryFile_writeLong(THFile *self, long *data, size_t n)
     {
       int i, big_endian = !THDiskFile_isLittleEndianCPU();
       size_t nByte = 8*n;
-      long *storage = (long *)(mfself->storage->data + mfself->position);
       THMemoryFile_grow(mfself, mfself->position+nByte);
+      int32_t *storage = (int32_t *)(mfself->storage->data + mfself->position);
```
soumith: this should be int64_t as well

lib/TH/THMemoryFile.c
BTNC: It should be int32_t. This else branch is to handle sizeof(long)!=8 while longSize==8 case (for example type long is of 4 bytes). Current implementation has a trick to use two 4bytes integers to represent one 8bytes long. Please check the main implementation in the following lines.
PS. I do not think this case is useful as long as the input "data" is of "long *". Do you know an use case for this?

lib/TH/THMemoryFile.c
BTNC: It should be int32_t. The reason is same with last comment.

soumith:
Thank you!

PR #776, opened 2016-09-27
windows high resolution timer with a few makefile changes

BTNC:
Hi,

This pr contains:
(1) windows high resolution timer with a test file (the old impl's precision is seconds);
(2) makefile changes for msvc;

Thanks,


soumith:
Thank you!

PR #784, opened 2016-10-02
replace long with ptrdiff_t for memory size/offset etc

BTNC:
Hi,

Type long is used extensively in torch mainly for (1) memory size/offset, element counts; (2) representing data; (3) meeting 3rd party api in/out specification. This PR is about long usage in (1).

Impact:
(1) for users on linux/mac where rang of long is same as range of ptrdiff_t, this PR is expected to introduce no behavior change.
(2) for users on windows/arm x64 etc where long is 32bit, the maximum managed memory by torch used to be 2G, but now the limit is extended to 2^63, same as linux/mac users.

Please let me know if any part in this PR is not about usage (1) or this PR miss any change about usage (1).

For usage (2), my proposal is to have a torch long type like th_Long which can be configured for integer types other than long. So windows/arm x64 users can also use 64bit integer. THGenerateIntTypes.h is now providing very limited configuration.

Thanks,


soumith:
hi, i see that you have built your commit "replace long with ptrdiff_t for memory size/offset etc" on top of your other branch that fixes time. 
Hence there are lot of merge conflicts.

I am reviewing that commit now, but I think you should isolate that commit on top of master.

soumith:
ok the PR largely looks good. I will re-review it again in a bit.
I'll also test its downstream dependencies like cutorch.
Are you planning to send an equivalent PR to cutorch, or is torch7 your main interest?

For now, please close this PR, and isolate and rebase the commit "replace long with ptrdiff_t for memory size/offset etc" onto torch7/master and send a new PR.
Thank you :)

BTNC:
I have not tried the new cuda toolkit8 which may finally compile with vs2015. But as long as I can try it out, I will send an equivalent PR for cutorch. I will close this PR and resend a new PR.

apaszke:
Actually `ptrdiff_t` it's a correct type for this purpose. You should probably use `size_t`, as that's what's used by `malloc` and co.

BTNC:
I did find many places that should better use size_t, but after evaluation I found it may require code logic change in many places due to sign/unsign (which can easily introduce bugs if not addressed properly), and may introduce behavior changes for most users(Linux/Mac). As a result, I decided to separate the update to several phases instead of one go.

To be clear, my plan was: (1) categorize long usage and address one by one.  As described in this PR description, this PR is to address long usage for memory size/offset. (2) addressing each in small phases until the best shape. Replacing long with ptrdiff_t is a relatively smooth change, though identifying which long should be replaced is laborious. size_t change can be on top of the ptrdiff_t change, i.e. replacing some ptrdiff_t with size_t. Changing to size_t is not my current concern since I am on 64bit systems, but 32bit users may benefit from it.

PR #785, opened 2016-10-03
replace long with ptrdiff_t for memory size/offset, element counts etc

BTNC:
Hi,

Type long is used extensively in torch mainly for (1) memory size/offset, element counts; (2) representing data; (3) meeting 3rd party api in/out specification. This PR is about long usage in (1).

Impact:
(1) for users on linux/mac where rang of long is same as range of ptrdiff_t, this PR is expected to introduce no behavior change.
(2) for users on windows/arm x64 etc where long is 32bit, the maximum managed memory by torch used to be 2G, but now the limit is extended to 2^63, same as linux/mac users.

Replacing long with ptrdiff_t is a relatively smooth change, though identifying which long should be replaced is laborious. For places that should better use size_t, size_t change can be on top of the ptrdiff_t change, i.e. replacing some ptrdiff_t with size_t. I did not replace some long with size_t because after evaluation I found replacement with size_t requires code logic changes in many places which could complicate this PR and introduce big behavior change for most users, and I am uncertain about that impact.

Please let me know if any part in this PR is not about usage (1) or this PR miss any change about usage (1).

For long usage (2), my proposal is to have a torch long type like th_Long which can be configured for integer types other than long. So windows/arm x64 users can also use 64bit integer. THGenerateIntTypes.h is now providing very limited configuration.

Thanks,


soumith:
Thanks a lot @BTNC . Whoever you are you are a godsend for the windows users.

PR #800, opened 2016-10-17
Expose OpenMP num threads through TH lib

colesbury:
Expose omp_set_num_threads and similar APIs through the TH lib. This
means a third-party libaries using TH don't need to be compiled with
OpenMP support just to control the number of TH OMP threads.


soumith:
Thanks!

Issue #762, opened 2016-09-10
[Port] Portability problem introduced by THVector.c


CDLuminate:
Check buildlog here:
https://buildd.debian.org/status/package.php?p=lua-torch-torch7&suite=experimental#problem-1

All architectures fails to build except for x86.

soumith:
@rguthrie3 i guess we need an elseif x86 and then finally an else for all other processors....

rguthrie3:
Yeah it looks like it.  The fastest fix I think would be to make vectorDispatchInit just return and do nothing (leaving the dispatch pointers initialized to the default implementations) if the SSE macros are not defined.  This should at least allow it to build and work properly everywhere.  The better fix would be to have the x86 macro and change generic/simd/simd.h's conditional compilation accordingly.

dusty-nv:
Any update on building for non-x86 architectures?

soumith:
fixed in https://github.com/torch/torch7/pull/793 thanks to @howard0su

CDLuminate:
Thanks!

CDLuminate:
@soumith Please re-open this issue -- as there is still problem on protability of THVector.c

https://buildd.debian.org/status/package.php?p=lua-torch-torch7&suite=experimental

howard0su:
Can u try the following patch? https://github.com/torch/torch7/pull/801

I don't know how to run as your log shows. so the testing is only limited to normal build.

from what I can tell from log, seems amd64 is a different issue other than other failed platform. this patch only address the issue on build failure on several platforms.

CDLuminate:
@howard0su Thank you, I'll look into it later. But I have to say the amd64 build failure you see was not caused by the same reason as for arm or etc, but invalid simd instruction set -- the build machine doesn't support SSE4.1 (I guess it is quite old a machine but it's indeed amd64).

howard0su:
That's my understanding as well. this patch doesn't address that problem. I may look at it later after build break is fixed first.

howard0su:
@CDLuminate  any updates on it? after this build problem is confirmed to be fixed, I can look at amd64 problem.

CDLuminate:
@howard0su i think i will be available to handle it after Nov 11.

CDLuminate:
@howard0su I have solution to the amd64 build failure. The root of trouble lays in packaging.

Erhannis:
@CDLuminate Any luck implementing said solution?  Or do you have any other hints about what the problem is?

[...]

PR #801, opened 2016-10-17
Fix build on other platform than x86/arm with extension support.

howard0su:
try to fix the build issue reported by #762 

I only has access to amd64 and arm box. so I only tested it on that platforms.


soumith:
this should be fixed in master by other PRs

PR #803, opened 2016-10-19
add read/writePointer to be used in threads (de)serializePointer

BTNC:
Hi,

Added read/writePointer together with a test case 'readWritePointer' in test/test_sharedmem.lua. The implementation of (de)serializePointer in threads will be replaced with these new functions instead of relying on readWriteLong/Double.

Thanks,

### DIFF
--- a/File.c
+++ b/File.c
@@ -155,6 +155,47 @@ static int torch_File_writeString(lua_State *L)
   return 1;
 }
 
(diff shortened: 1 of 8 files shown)


BTNC:
To resend a new pr in case commits to resolve *nix compilation error pollute master's log.

PR #804, opened 2016-10-19
add read/writePointer to be used in threads (de)serializePointer

BTNC:
Hi,

Added read/writePointer together with a test case 'readWritePointer' in test/test_sharedmem.lua. The implementation of (de)serializePointer in threads will be replaced with these new functions instead of relying on readWriteLong/Double.

Travis-ci fail for osx with following error during test, but the compilation does succeed.
/Users/travis/torch/install/bin/luajit: /Users/travis/torch/install/share/lua/5.1/torch/init.lua:13: cannot load '/Users/travis/torch/install/lib/lua/5.1/libtorch.so'

Thanks,

### DIFF
--- a/doc/file.md
+++ b/doc/file.md
@@ -68,6 +68,9 @@ If a `Storage` is given, the method will attempt to read a number of elements
 equals to the size of the given storage, and fill up the storage with these elements.
 The number of elements actually read is returned.
 
+A convenient method exists to read one pointer as a integer: `[integer] readPointer()`. It reads
+in `"%p"` format if `File` is of ascii encoding. This method does not support `n` or `Storage`.
+
 In case of read error, these methods will call the `Lua` error function using the default
 [pedantic](#torch.File.pedantic) option, or stay quiet with the [quiet](#torch.File.quiet)
 option. In the latter case, one can check if an error occurred with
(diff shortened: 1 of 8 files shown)


soumith:
thanks. dont worry about the OSX build, it's been failing on travis, i have to take a look at it.

doc/file.md
```diff
@@ -68,6 +68,9 @@ If a `Storage` is given, the method will attempt to read a number of elements
 equals to the size of the given storage, and fill up the storage with these elements.
 The number of elements actually read is returned.
 
+A convenient method exists to read one pointer as a integer: `[integer] readPointer()`. It reads
```
gchanan: nit: rename convenient to convenience.

doc/file.md
```diff
@@ -109,6 +112,9 @@ in the storage.
 
 These methods return the number of elements actually written.
 
+A convenient method exists to write one pointer: `writePointer(integer)`. It writes
```
gchanan: here too.

BTNC:
Yes, it seems only useful for MemoryFile. However if remove this from DiskFile, it would be against the purpose of virtual table and will complicate the implementation and maintenance. By using virtual table (as in c++), it is not expected some sub class have null function pointers in virtual table. To remove this from DiskFile, the function pointers have to be initialized as null pointers and the callers have to check if the function pointers are null pointers before calling them. This special logic will make the implementation less straightforward and less maintainable. Besides, it will add unnecessary null pointer check for the normal case, the MemoryFile case.

Issue #809, opened 2016-10-21
Shared storage allocation error


tworec:
I have problem creating new file for FloatStorage

```
In [1]:
s=torch.FloatStorage('/tmp/2rec', true, 10000)
```
gives me

```
unable to stretch file </tmp/2rec> to the right size at /home/tworec/torch/pkg/torch/lib/TH/THAllocator.c:262
stack traceback:
    [C]: at 0x7fd308444330
    [C]: in function 'FloatStorage'
    [string "s=torch.FloatStorage('/tmp/2rec', true, 10000..."]:1: in main chunk
    [C]: in function 'xpcall'
    /home/tworec/torch/install/share/lua/5.1/itorch/main.lua:210: in function </home/tworec/torch/install/share/lua/5.1/itorch/main.lua:174>
    /home/tworec/torch/install/share/lua/5.1/lzmq/poller.lua:75: in function 'poll'
    /home/tworec/torch/install/share/lua/5.1/lzmq/impl/loop.lua:307: in function 'poll'
    /home/tworec/torch/install/share/lua/5.1/lzmq/impl/loop.lua:325: in function 'sleep_ex'
    /home/tworec/torch/install/share/lua/5.1/lzmq/impl/loop.lua:370: in function 'start'
    /home/tworec/torch/install/share/lua/5.1/itorch/main.lua:389: in main chunk
    [C]: in function 'require'
    (command line):1: in main chunk
    [C]: at 0x00406670
```

I think this change is flawed
https://github.com/torch/torch7/commit/d49aee3fa631eac9d5483c600e9553aa15442fde#diff-7da47904ec3be10d37adb2af15fdc839R276

When I move one or two commits before this commit and install torch everything seems to work well.

apaszke:
Apparently the problem is that `lseek` resizes the file for you if you try to seek after its end, while `fstat` only reads the size. I'll send a fix for this issue sometime today. Sorry!

soumith:
That pr indeed fixes another bug i guess

soumith:
@apaszke confirms that the pr fixes this bug as well

tworec:
yeah, I confirm it too.

PR #817, opened 2016-10-26
fix bug in mmaping

soumith:
fixes https://github.com/torch/torch7/issues/809


tworec:
This commit doesn't change condition which was cause of #809 but I'll check

apaszke:
It doesn't change the condition, but it adds a call that performs a side effect of lseek.

tworec:
I confirm. This resolved #809

PR #818, opened 2016-10-26
torch.HalfTensor mostly working

borisfom:
@soumith, @szagoruyko , @ngimel:
Here, first cut, as we discussed in https://github.com/torch/cutorch/issues/484.
Math for HalfTensor should be extended later - only for ARM and other CPUs with native __fp16 type.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,13 @@ IF(MSVC)
   ADD_DEFINITIONS(-D_CRT_SECURE_NO_DEPRECATE=1)
 ENDIF(MSVC)
 
+#
+# Temporary: flag to enable new 'half' type in THGenerateAllTypes.h
+# Enable for Torch build (disabled by default).
+# Will make default and remove once all modules are updated accordingly
+#
+ADD_DEFINITIONS(-DTH_GENERIC_USE_HALF=1)
+
 # OpenMP support?
 SET(WITH_OPENMP ON CACHE BOOL "OpenMP support if available?")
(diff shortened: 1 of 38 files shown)


ngimel:
torch.HalfTensor will also have to be added to cutorch, in particular Tensor.lua and generic/CTensor.c

borisfom:
@ngimel : absolutely! I am also preparing to push a commit with (temporary) config switches so that the rest of modules build.

borisfom:
@soumith: I think I have fixed it all as of commit 84d766c, should be ready to go!

borisfom:
@soumith: actually, please hold on on merging this: after intergrating with cutorch I think the type should go into cwrap to avoid duplication (torch's torchcwrap is not exported). Also the C type name should be the same - not so hard to do.

borisfom:
@soumith: it you think we should leave cwrap alone, then I suppose, exporting torchcwrap is in order - please let me know!

soumith:
i think the type being here is fine. it's not a fundamental type and does not have to go into cwrap.

borisfom:
@soumith : right, I reached the same conclusion about the location of 'half' - should be in pkg/torch.
I am finishing my tweaks to use the same type name - it looks cleaner and let us avoid a few dummy type conversions, but it's a bit more tedious as I can't simply move it from cutorch as it relies on CUDA header so the definition needs to be conditional depending of where THHalf.h is included from.

soumith:
cc: @gchanan

borisfom:
Superceded by https://github.com/torch/torch7/pull/826

PR #820, opened 2016-10-27
Updated maskedCopy example #1

jdmartin86:
Original call to maskedCopy() only included y, not the mask. Update added the mask to the first argument.


soumith:
Thanks!

PR #821, opened 2016-10-28
fix one func call for lua52/53 on win; update rockspec for win

BTNC:


BTNC:
The (void) cast should be checked in because lua_pushstring is returning char * instead of void while lua_pushnil is still returning void from lua52 on which cause msvc fail to compile, but the rockspec change is not needed.

PR #823, opened 2016-10-28
Use -O3 when compiling TH

colesbury:
On my machine, this speeds up FloatTensor copies by 4x.


lib/TH/CMakeLists.txt
```diff
@@ -387,6 +387,8 @@ IF (BLAS_FOUND AND BLAS_INSTALL_LIBRARIES)
   ENDIF()
 ENDIF()
 
+SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O3")
```
soumith: can you set this only in RELEASE mode

PR #828, opened 2016-11-02
Add more size checks and improve some LAPACK error messages

apaszke:
Also, linear system solvers can now accept vectors (they had to be reshaped to Nx1 matrices before).

### DIFF
--- a/lib/TH/generic/THTensorLapack.c
+++ b/lib/TH/generic/THTensorLapack.c
@@ -103,12 +103,23 @@ static THTensor *THTensor_(cloneColumnMajor)(THTensor *self, THTensor *src)
 
 void THTensor_(gesv)(THTensor *rb_, THTensor *ra_, THTensor *b, THTensor *a)
 {
+  int free_b = 0;
   if (a == NULL) a = ra_;
   if (b == NULL) b = rb_;
-  THArgCheck(a->nDimension == 2, 2, "A should be 2 dimensional");
-  THArgCheck(b->nDimension == 2, 1, "B should be 2 dimensional");
-  THArgCheck(a->size[0] == a->size[1], 2, "A should be square");
-  THArgCheck(a->size[0] == b->size[0], 2, "A,b size incompatible");
+  THArgCheck(a->nDimension == 2, 2, "A should have 2 dimensions, but has %d",
+      a->nDimension);
+  THArgCheck(b->nDimension == 1 || b->nDimension == 2, 1, "B should have 1 or 2 "
+      "dimensions, but has %d", b->nDimension);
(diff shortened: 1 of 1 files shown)


lib/TH/generic/THTensorLapack.c
```diff
@@ -103,12 +103,22 @@ static THTensor *THTensor_(cloneColumnMajor)(THTensor *self, THTensor *src)
...
+      a->nDimension);
+  THArgCheck(b->nDimension == 1 || b->nDimension == 2, 1, "B should have 1 or 2 "
+      "dimensions, but has %d", b->nDimension);
+  THArgCheck(a->size[0] == a->size[1], 2, "A should be square, but is %ldx%ld",
+      a->size[0], a->size[1]);
+  THArgCheck(a->size[0] == b->size[0], 2, "A,B size incompatible - A has %ld "
+      "rows, B has %ld", a->size[0], b->size[0]);
+
+  if (b->nDimension == 1) {
+    b = THTensor_(newWithStorage2d)(b->storage, b->storageOffset, b->size[0], -1, 1, -1);
```
fmassa: Hum, does this work if the tensor is non-contiguous with a stride > 1?
Maybe it would be better to do something like the following ? (caution, not tested!)
```C
b = THTensor_(newWithStorage2d)(b->storage, b->storageOffset, b->size[0], b->stride[0], 1, -1);
```

apaszke: Good point. I'll change that. Your approach should work fine. I'm not sure what will the second stride be, but I guess it doesn't matter, since it's a singleton dim.

lib/TH/generic/THTensorLapack.c
```diff
@@ -103,12 +103,23 @@ static THTensor *THTensor_(cloneColumnMajor)(THTensor *self, THTensor *src)
...
+  THArgCheck(b->nDimension == 1 || b->nDimension == 2, 1, "B should have 1 or 2 "
+      "dimensions, but has %d", b->nDimension);
+  THArgCheck(a->size[0] == a->size[1], 2, "A should be square, but is %ldx%ld",
+      a->size[0], a->size[1]);
+  THArgCheck(a->size[0] == b->size[0], 2, "A,B size incompatible - A has %ld "
+      "rows, B has %ld", a->size[0], b->size[0]);
+
+  if (b->nDimension == 1) {
+    b = THTensor_(newWithStorage2d)(b->storage, b->storageOffset, b->size[0],
+            b->stride[0], 1, 0);
```
fmassa: It may indeed not matter what value you put for the second stride, but I wonder if a value of 1 wouldn't be more in line with current behaviour of torch?
Anyway, I think this is more a matter of taste here now (specially because this tensor won't be exposed to the external interface anyway).

soumith:
Thanks!

PR #834, opened 2016-11-11
Fix implementation of logNormal

killeent:
Current implementation was broken. See numpy:

https://github.com/numpy/numpy/blob/c90d7c94fd2077d0beca48fa89a423da2b0bb663/numpy/random/mtrand/distributions.c#L691

for a correct implementation, which we adopted here.

Test Plan:

- Add unit test, stolen from cutorch

### DIFF
--- a/lib/TH/THRandom.c
+++ b/lib/TH/THRandom.c
@@ -255,10 +255,8 @@ double THRandom_cauchy(THGenerator *_generator, double median, double sigma)
    M'enfin. */
 double THRandom_logNormal(THGenerator *_generator, double mean, double stdv)
 {
-  double zm = mean*mean;
-  double zs = stdv*stdv;
   THArgCheck(stdv > 0, 2, "standard deviation must be strictly positive");
-  return(exp(THRandom_normal(_generator, log(zm/sqrt(zs + zm)), sqrt(log(zs/zm+1)) )));
+  return(exp(THRandom_normal(_generator, mean, stdv)));
 }
 
 int THRandom_geometric(THGenerator *_generator, double p)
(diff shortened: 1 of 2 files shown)


fmassa:
Are you sure about this? Torch's implementation  seems to be consistent with Matlab's https://fr.mathworks.com/help/stats/lognstat.html

soumith:
okay so it seems this:

numpy and curand generate a lognormal distribution taking in the mean and stdv of the underlying normal distribution.

torch (and seems like matlab) generate a lognormal distribution taking in the mean and stdv of the currently generated numbers.

soumith:
actually, @fmassa you are wrong.
Torch has gotten logNormal wrong.

In [Matlab](https://www.mathworks.com/help/stats/lognrnd.html), [SciPy](https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.lognormal.html) and in [curand](http://docs.nvidia.com/cuda/curand/host-api-overview.html#generation-functions), logNormal does the same thing. It generates a distribution with mean and stdv given which are of the underlying normal distribution.
I'll merge this in.

fmassa:
@soumith sorry, I was wrong and the behaviour of matlab, numpy and curand seems to be consistent, my bad.

I got confused because I thought that `torch.logNormal` was supposed to generate samples from a log normal distribution with the mean and std that is provided by the user.
So if one do
```
N = 10000
mean = 2
std = 4
t = torch.Tensor(N):logNormal(mean, std)
print(t:mean())
print(t:std())
```
I would expect `t:mean()` to be close to 2, and `t:std()` to be close to 4, and that's what the old behaviour of torch used to do.

But that's not what matlab/numpy/curand do, so it's better to be consistent with them.

PR #842, opened 2016-11-18
Adding macros to control generic types selection

borisfom:
Minimal changes to allow reduces set of types + refactoring of GenerateAllXX headers


soumith:
i'm sorry but you misunderstood me. we should only do this for cutorch, and not for torch.
torch binaries are < 2 MB, this stuff is not necessary at all.

A PR like this for cutorch, with THC_GENERIC_NO_* defined is fine with me.

borisfom:
No problem! You're right, not worth the trouble in this module.

PR #851, opened 2016-11-28
2 trivial changes to compile with msvc

BTNC:
Hi,

1. THTensorLapack.c : to be consistent with THLapackCheckWithCleanup
2. luaT.c : `lua_pushstring()` returns `const char *` instead of `void `since lua 5.2

Thanks,


soumith:
thanks!

PR #852, opened 2016-11-28
Adding AVX Optimizations with Runtime Dispatch

soumith:
Work in Progress.

Initial work by Roberth Guthrie (@rguthrie3) and more additions ongoing by Keren Zhou (@jokeren)

This speeds up our Tensor Math ops. On add, we have up to 11x speedup already, and more to come.


Jokeren:
Hi @soumith,

Regarding the coding style, I notice that many functions, such as `cadd`, `remainder`, and `mul`, do not follow the two-space indentation rule. Hence, should modify their styles?

soumith:
@Jokeren you can modify their styles, yes.

Jokeren:
@soumith , Should I include <omp.h> when using Clang?

Jokeren:
@apaszke, I used <omp.h> to calculate start and end indices for each thread. But it appears that Clang does not support openmp well, at least for the environment on travis-ci. Could you help fix the problem?

apaszke:
@Jokeren What's the exact issue you're having?

apaszke:
Ok, I've check the travis build. You shouldn't assume that the lib will be always compiled in an environment that has OpenMP. Use `#ifdef _OPENMP` to guard these parts [like in the other files](https://github.com/torch/torch7/blob/608712bfb2ec27f939b6892fcf76a89e9e1d7e99/lib/TH/THGeneral.c#L4-L6). You could only hide the `#pragma omp parallel ...` sections and have the `i` and `i_end` always computed as the whole array, if compiled without OpenMP.

generic/Tensor.c
```diff
@@ -1319,6 +1319,7 @@ void torch_Tensor_(init)(lua_State *L)
   luaT_setfuncs(L, torch_Tensor_(_), 0);
   lua_pop(L, 1);
   THVector_(vectorDispatchInit)();
+  THTensor_(tensorMathDispatchInit)();
```
apaszke: Also, could you please merge these into a single function? We should have one entry point for initializing the function pointers, or we'll have people forgetting about them.

Jokeren: Hi, @apaszke , thanks for your kind help! Please see my recent changes.

I plan to merge `THTensorMathSIMD.c` into `SSE.c`, which you requested some days ago.

    If contiguous
      multi-thread fork-join
        THVector_op
    else
      TH_TENSOR_APPLY

I use THVector to apply the corresponding operation when data is contiguous. So in this paradigm, I merge all vector functions into `SSE.c` and `NEON.c` and maintained non-contiguous tensor operations. But notice that I also modify the original `add` function to `cadd` by doing this. 

If the design is ok, I will move on.

lib/TH/generic/THTensorMath.c
```diff
@@ -458,15 +462,24 @@ void THTensor_(add)(THTensor *r_, THTensor *t, real value)
...
+      size_t num_threads = omp_get_num_threads();
+      size_t tid = omp_get_thread_num();
+      #else
+      size_t num_threads = 1;
+      size_t tid = 0;
+      #endif
+      ptrdiff_t i = tid * (sz / num_threads);
+      ptrdiff_t i_end = tid == num_threads - 1 ? sz : i + sz / num_threads;
+      THVector_(add)(rp+i, tp+i, value, i_end-i); 
+    }
```
adamlerer: It looks like this pattern is copy-pasted across this file. How about wrapping this in a macro?

```
TH_TENSOR_APPLY2_CONTIG(add, tp, rp, sz);
```

lib/TH/generic/THTensorMath.c
```diff
@@ -1488,75 +1546,75 @@ static void THTensor_(quicksortascend)(real *arr, long *idx, long elements, long
 
   while(!done) {
       /* Use median of three for pivot choice */
-      P=(L+R)>>1;
-      BOTH_SWAP(P, L+1);
-      if (ARR(L+1) > ARR(R)) { BOTH_SWAP(L+1, R); }
-      if (ARR(L) > ARR(R)) { BOTH_SWAP(L, R); }
-      if (ARR(L+1) > ARR(L)) { BOTH_SWAP(L+1, L); }
+    P=(L+R)>>1;
```
adamlerer: All these changes to sort are just indentation?

lib/TH/generic/THVectorDefault.c
```diff
@@ -17,68 +17,100 @@ void THVector_(fill_DEFAULT)(real *x, const real c, const ptrdiff_t n) {
     x[i] = c;
 }
 
+void THVector_(cadd_DEFAULT)(real *z, const real *x, const real *y, const real c, const ptrdiff_t n)
+{
+  ptrdiff_t i = 0;
+
+  for(; i<n-4; i+=4)
```
adamlerer: Can we get rid of this manual loop unrolling (everywhere)? I'd be shocked if on a modern compiler this actually helps. But not sure what compilers we're targeting here...

lib/TH/vector/AVX.c
```diff
@@ -0,0 +1,243 @@
...
+#else
+#include <intrin.h>
+#endif
+
+static void THDoubleVector_fill_AVX(double *x, const double c, const ptrdiff_t n) {
+  ptrdiff_t i;
+  ptrdiff_t off;
+  __m256d YMM0 = _mm256_set_pd(c, c, c, c);
+  for (i=0; i<=((n)-16); i+=16) {
+    _mm256_storeu_pd((x)+i  , YMM0);
```
adamlerer: The compiler doesn't do loop unrolling for you? I'm surprised you're seeing a benefit from manually unrolling these. Any idea why?

lib/TH/generic/THVector.h
```diff
@@ -3,10 +3,12 @@
...
 TH_API void THVector_(fill)(real *x, const real c, const ptrdiff_t n);
+TH_API void THVector_(cadd)(real *z, const real *x, const real *y, const real c, const ptrdiff_t n);
 TH_API void THVector_(add)(real *y, const real *x, const real c, const ptrdiff_t n);
-TH_API void THVector_(diff)(real *z, const real *x, const real *y, const ptrdiff_t n);
-TH_API void THVector_(scale)(real *y, const real c, const ptrdiff_t n);
-TH_API void THVector_(mul)(real *y, const real *x, const ptrdiff_t n);
+TH_API void THVector_(cmul)(real *z, const real *x, const real *y, const ptrdiff_t n);
+TH_API void THVector_(mul)(real *y, const real *x, const real c, const ptrdiff_t n);
+TH_API void THVector_(cdiv)(real *z, const real *x, const real *y, const ptrdiff_t n);
+TH_API void THVector_(div)(real *y, const real *x, const real c, const ptrdiff_t n);
```
adamlerer: Have you tried adding `THVector_(copy)` here? I know it sounds silly but I think AVX copy may be faster. And it's more common than these other ops.

lib/TH/generic/THTensorMath.c
Jokeren: Yes, I think the correct indentation style is 2 spaces

lib/TH/generic/THVector.h
Jokeren: OK, I will add it in the next commits

lib/TH/vector/AVX.c
Jokeren: Well, actually I reference `SSE.c` in which loops are unrolled manually. I was also confused about the implementation, and test results show no difference between unrolled version and the original version.

I paste a piece of code in `SSE.c` for you:

```
static void THDoubleVector_cdiv_SSE(double *z, const double *x, const double *y, const ptrdiff_t n) {
  ptrdiff_t i;                                                                                                                                               
  __m128d XMM0, XMM1, XMM2, XMM3;
  for (i=0; i<=((n)-4); i+=4) {
    XMM0 = _mm_loadu_pd(x+i);
    XMM1 = _mm_loadu_pd(x+i+2);
    XMM2 = _mm_loadu_pd(y+i);
    XMM3 = _mm_loadu_pd(y+i+2);
    XMM2 = _mm_div_pd(XMM0, XMM2);
    XMM3 = _mm_div_pd(XMM1, XMM3);
    _mm_storeu_pd(z+i, XMM2);
    _mm_storeu_pd(z+i+2, XMM3);
  }
  for (; i<(n); i++) {
    z[i] = x[i] / y[i];
  }
}
```

Jokeren:
Hi, @adamlerer, I have rebased the branch.

pavanky:
@Jokeren something went wrong. There are commits appearing multiple times.

Jokeren:
Thanks, @pavanky , I have noticed the problem.

I just rebased the branch again by `git rebase -i`. So it appears that all the previous commits are appended. Could you provide me some advises to fix it?

pavanky:
I tried benchmarking the following code on master as well as the branch on this PR. I am getting identical performance.

```
local n = 1 << 24
local a = torch.DoubleTensor(1 << 24):uniform()
local b = torch.DoubleTensor(1 << 24):uniform()
local timer = torch.Timer()
local niter = 100
for i = 1, niter do
   a:add(b)
end
local ts = timer:time().real
print('Time taken (Size: ' .. n .. ', niter: '  .. niter .. '): ' .. ts .. ' s')
```

Is this may be related to cmake not finding AVX1 (which defines `USE_AVX`) but finding AVX2 (which defines `USE_AVX2` but not `USE_AVX`) ? Or am I doing the benchmark incorrectly ?

This is the output I am seeing on i7-5500U with gcc 5.4.0.

```
-- Performing Test C_HAS_SSE1_1
-- Performing Test C_HAS_SSE1_1 - Success
-- Performing Test C_HAS_SSE2_1
-- Performing Test C_HAS_SSE2_1 - Success
-- Performing Test C_HAS_SSE3_1
-- Performing Test C_HAS_SSE3_1 - Failed
-- Performing Test C_HAS_SSE3_2
-- Performing Test C_HAS_SSE3_2 - Success
-- Performing Test C_HAS_SSE4_1_1
-- Performing Test C_HAS_SSE4_1_1 - Failed
-- Performing Test C_HAS_SSE4_1_2
-- Performing Test C_HAS_SSE4_1_2 - Success
-- Performing Test C_HAS_SSE4_2_1
-- Performing Test C_HAS_SSE4_2_1 - Failed
-- Performing Test C_HAS_SSE4_2_2
-- Performing Test C_HAS_SSE4_2_2 - Success
-- Performing Test C_HAS_AVX_1
-- Performing Test C_HAS_AVX_1 - Failed
-- Performing Test C_HAS_AVX_2
-- Performing Test C_HAS_AVX_2 - Success
-- Performing Test C_HAS_AVX2_1
-- Performing Test C_HAS_AVX2_1 - Failed
-- Performing Test C_HAS_AVX2_2
-- Performing Test C_HAS_AVX2_2 - Success
-- Performing Test CXX_HAS_SSE1_1
-- Performing Test CXX_HAS_SSE1_1 - Success
-- Performing Test CXX_HAS_SSE2_1
-- Performing Test CXX_HAS_SSE2_1 - Success
-- Performing Test CXX_HAS_SSE3_1
-- Performing Test CXX_HAS_SSE3_1 - Failed
-- Performing Test CXX_HAS_SSE3_2
-- Performing Test CXX_HAS_SSE3_2 - Success
-- Performing Test CXX_HAS_SSE4_1_1
-- Performing Test CXX_HAS_SSE4_1_1 - Failed
-- Performing Test CXX_HAS_SSE4_1_2
-- Performing Test CXX_HAS_SSE4_1_2 - Success
-- Performing Test CXX_HAS_SSE4_2_1
-- Performing Test CXX_HAS_SSE4_2_1 - Failed
-- Performing Test CXX_HAS_SSE4_2_2
-- Performing Test CXX_HAS_SSE4_2_2 - Success
-- Performing Test CXX_HAS_AVX_1
-- Performing Test CXX_HAS_AVX_1 - Failed
-- Performing Test CXX_HAS_AVX_2
-- Performing Test CXX_HAS_AVX_2 - Success
-- Performing Test CXX_HAS_AVX2_1
-- Performing Test CXX_HAS_AVX2_1 - Failed
-- Performing Test CXX_HAS_AVX2_2
-- Performing Test CXX_HAS_AVX2_2 - Success
```

Jokeren:
Hi, @pavanky , I have tried your benchmark scripts.

`a:add(b)` this statement will call `THBlas_(axpy)` in the `THTensor_(cadd)` function. Thus, suppose you have `MKL` installed, it will automatically invokes the `mkl_axpy_avx2` function no matter which branch you use.

If you want to test the optimization effects of `cadd` (contiguous add), you should allocate another array as the sum of `a` and `b`. You can reference my sample here:

https://github.com/Jokeren/torch-tensor-benchmark/blob/master/bench_cadd_contiguous.lua

pavanky:
@Jokeren using your script

On master:

```
contiguous float cadd single thread:	
0.0051441192626953	
contiguous float cadd multi-thread:	
0.11501288414001	
contiguous double cadd single thread:	
0.0079038143157959	
contiguous double cadd multi-thread:	
0.21997499465942
```

On simd-opt

```
contiguous float cadd single thread:	
0.0054259300231934	
contiguous float cadd multi-thread:	
0.10986804962158	
contiguous double cadd single thread:	
0.0080790519714355	
contiguous double cadd multi-thread:	
0.21401214599609	
```

Jokeren:
@pavanky , regarding your question, I dig into the source code and asm, trying to find out where the problem lies.

1. In fact, on my `E5-2680v3` with a single thread, there is a tiny speedup with the `simd-opt` branch.

contiguous float cadd single thread:
0.0071110725402832
contiguous float cadd multi-thread:
0.048126935958862
contiguous double cadd single thread:
0.0081689357757568
contiguous double cadd multi-thread:
0.1191349029541

contiguous float cadd single thread:
0.0081889629364014
contiguous float cadd multi-thread:
0.056603908538818
contiguous double cadd single thread:
0.0092580318450928
contiguous double cadd multi-thread:
0.12701296806335

2. Another thing I found annoying is that the `detectHostSIMDExtensions` function fails to detect the `AVX2` flag. So the `cadd_AVX` implementation is called instead of `cadd_AVX2`. Could you please examine whether `AVX2` is detected on your computer. Whatever, FMA instructions do not help a lot in my tests.

pavanky:
@Jokeren I posted what cmake is detecting in an earlier comment. It is detecting AVX2 but not AVX.

Jokeren:
@pavanky , Yes, I know that cmake can enable `USE_AVX2` which I previously added.

But what I want to check is the statement in `lib/TH/generic/simd/simd.h`:

```
cpuid(&eax, &ebx, &ecx, &edx);
if (ebx & CPUID_AVX2_BIT)
  hostSimdExts |= SIMDExtension_AVX2;
```

Whether `hostSimdExts |= SIMDExtension_AVX2` will be executed or not?

Sorry that I have not point it out for you.

pavanky:
I ran the benchmark with the following patch applied:

```
@@ -120,15 +120,21 @@ static inline uint32_t detectHostSIMDExtensions()
   // Check for AVX2. Requires separate CPUID
   eax = 0x7;
   cpuid(&eax, &ebx, &ecx, &edx);
-  if (ebx & CPUID_AVX2_BIT)
-    hostSimdExts |= SIMDExtension_AVX2;
+  if (ebx & CPUID_AVX2_BIT) {
+      printf("------Has AVX2 BIT------------\n");
+      hostSimdExts |= SIMDExtension_AVX2;
+  }
 
   eax = 0x1;
   cpuid(&eax, &ebx, &ecx, &edx);
-  if (ecx & CPUID_AVX_BIT)
-    hostSimdExts |= SIMDExtension_AVX;
-  if (edx & CPUID_SSE_BIT)
-    hostSimdExts |= SIMDExtension_SSE;
+  if (ecx & CPUID_AVX_BIT) {
+      printf("------Has AVX BIT------------\n");
+      hostSimdExts |= SIMDExtension_AVX;
+  }
+  if (edx & CPUID_SSE_BIT) {
+      printf("------Has SSE BIT------------\n");
+      hostSimdExts |= SIMDExtension_SSE;
+  }
 
   return hostSimdExts;
 }
```

This is the output:

```
------Has AVX BIT------------
------Has SSE BIT------------
------Has AVX BIT------------
------Has SSE BIT------------
------Has AVX BIT------------
------Has SSE BIT------------
------Has AVX BIT------------
------Has SSE BIT------------
------Has AVX BIT------------
------Has SSE BIT------------
------Has AVX BIT------------
------Has SSE BIT------------
------Has AVX BIT------------
------Has SSE BIT------------
contiguous float cadd single thread:	
0.0044651031494141	
contiguous float cadd multi-thread:	
0.11115121841431	
contiguous double cadd single thread:	
0.007871150970459	
contiguous double cadd multi-thread:	
0.21293210983276	
```

pavanky:
@Jokeren forgot to mention your name in previous commit ^

Jokeren:
@pavanky , thanks a lot for your check!

So I guess it is a bug of detecting AXV2. Although we have not solved the mystery of performance, at least we find a bug😜. I will try to optimize or account for it in the next days. But I have to go to sleep now.

Jokeren:
I will report that bug, seeing how to fix it, as it is not written by me originally.

Jokeren:
@pavanky I think commit logs are correct now?

Jokeren:
The AVX2 support issue:

https://github.com/torch/torch7/issues/916

Jokeren:
@pavanky , for the performance problem, I find it hard to beat cadd with compiler optimizations. In our case, three contiguous arrays should be loaded but only a single FMA is applied. Hence the arithmetic intensity is extremely low, and the performance is dominated by AVX loads. You can remove the FMA instruction to test. I think the results will be only about 30% better.

The compiler, by default, uses SSE and unrolls the loop for several times (I use `objdump` to see the assembly codes). Because multiple ports exist in a core, the speed will be similar to AVX.

pavanky:
@Jokeren That is what I expected. The `-O3` optimizations are usually very hard to beat for vector operations.

> This speeds up our Tensor Math ops. On add, we have up to 11x speedup already, and more to come.

I saw this in the description and was surprised that the speedup can be that high. I wanted to check this out for myself.

lib/TH/generic/THTensorMath.c
Jokeren: @adamlerer , it is hard to enable openmp by inserting `#pragma` inside `#define`

fmassa: Maybe you could use the `_Pragma` function https://github.com/torch/torch7/pull/395/files#diff-7c0d3aa11c041d012ec0251c0252f017R508 ?

Jokeren: Thanks @fmassa. Yes, I know the usage. What I am worried about is the support of c99 to enable _Pragma. In the current version, I see only parts of codes are compiled with c99.

adamlerer:
@Jokeren using your script, the multi-thread version is very sensitive to OMP_NUM_THREADS:

```
/data/users/alerer/git/torch7$ OMP_NUM_THREADS=1 th cadd.lua
contiguous float cadd single thread:
0.0061922073364258
contiguous float cadd multi-thread:
0.075101137161255
contiguous double cadd single thread:
0.011461019515991
contiguous double cadd multi-thread:
0.20637106895447
/data/users/alerer/git/torch7$ OMP_NUM_THREADS=10 th cadd.lua
contiguous float cadd single thread:
0.006270170211792
contiguous float cadd multi-thread:
0.0083858966827393
contiguous double cadd single thread:
0.011456966400146
contiguous double cadd multi-thread:
0.016303777694702
/data/users/alerer/git/torch7$ OMP_NUM_THREADS=40 th cadd.lua
contiguous float cadd single thread:
0.0062191486358643
contiguous float cadd multi-thread:
0.48840308189392
contiguous double cadd single thread:
0.011584043502808
contiguous double cadd multi-thread:
0.51699090003967
```

The second thing to note, is that you're claiming that the pointwise operations are memory-bound, but you're looking at `cadd` which calls `mkl_blas_avx_xdaxpy`, so of course it's going to be efficient.

On `master`, I find that `add` is not memory bound. Looking at the timings below, note that for `cadd`, the float version is twice as fast, so memory (or SIMD) bound; whereas for `add`, they are both the same speed (and slower), meaning it's bound by executing non-SIMD instructions.

```
th> fs = torch.FloatTensor(100000):zero()
                                                                      [0.0004s]
th> ds = torch.DoubleTensor(100000):zero()
                                                                      [0.0006s]
th> for i = 1,100000 do fs:add(fs,fs) end
                                                                      [2.7250s]
th> for i = 1,100000 do ds:add(ds,ds) end
                                                                      [5.1225s]
th> for i = 1,100000 do fs:add(1) end
                                                                      [7.9280s]
th> for i = 1,100000 do ds:add(1) end
                                                                      [7.9726s]
```

However, on `simd_opt`, it looks like we *are* memory-bound (or SIMD-op bound)

```
th> for i = 1,100000 do fs:add(1) end
                                                                      [2.3672s]
th> for i = 1,100000 do ds:add(1) end
                                                                      [5.2245s]
```

and perf top tells me I'm now executing `THFloatVector_add_AVX`. I guess still waiting on the AVX2 bit to be fixed.

Bottom line: the SIMD does seem to be speeding things up a bit, but it seems we can't get a *ton* of speedup because of memory constraints. 

This is running at 20GB/s which is supposedly the max main memory bandwidth per core (although I would expect that 400K of data will be sitting in cache). When I really force it to go to main memory by using a 10+GB tensor, I'm seeing closer to 10GB/s

```
th> f = torch.FloatTensor(2.5e9):zero()
                                                                      [6.6841s]
th> d = torch.DoubleTensor(2.5e9):zero()
                                                                      [10.9034s]
th> f:add(1);
                                                                      [1.5745s]
th> d:add(1);
                                                                      [2.2922s]
```

pavanky:
@adamlerer You are right, I am seeing a 50% speedup on simd-opt branch for `:add(1)` at large sizes.

adamlerer:
So, @colesbury and I took a look at the source code of the two versions (gcc 4.8.5), and the truth is stranger than expected. Here's the inner loop of THFloatVector_add_DEFAULT:

```
  244710:       c5 fc 10 1c 06          vmovups ymm3,YMMWORD PTR [rsi+rax*1]
  244715:       48 83 c1 10             add    rcx,0x10
  244719:       c4 c1 7c 10 14 02       vmovups ymm2,YMMWORD PTR [r10+rax*1]
  24471f:       c5 e4 58 d9             vaddps ymm3,ymm3,ymm1
  244723:       c5 ec 58 d1             vaddps ymm2,ymm2,ymm1
  244727:       c5 fc 11 1c 07          vmovups YMMWORD PTR [rdi+rax*1],ymm3
  24472c:       c4 c1 7c 11 14 01       vmovups YMMWORD PTR [r9+rax*1],ymm2
  244732:       48 83 c0 40             add    rax,0x40
  244736:       4c 39 c1                cmp    rcx,r8
  244739:       7c d5                   jl     244710 <THFloatVector_add_AVX+0x20>
```

And here's the inner loop of THFloatVector_add_AVX (I think... there are a couple loops):

```
  23c05b:       c5 f8 10 0c 0e          vmovups xmm1,XMMWORD PTR [rsi+rcx*1]
  23c060:       49 83 c0 01             add    r8,0x1
  23c064:       c4 e3 75 18 4c 0e 10    vinsertf128 ymm1,ymm1,XMMWORD PTR [rsi+rcx*1+0x10],0x1
  23c06b:       01
  23c06c:       c5 f4 58 ca             vaddps ymm1,ymm1,ymm2
  23c070:       c5 f8 11 0c 0f          vmovups XMMWORD PTR [rdi+rcx*1],xmm1
  23c075:       c4 e3 7d 19 4c 0f 10    vextractf128 XMMWORD PTR [rdi+rcx*1+0x10],ymm1,0x1
  23c07c:       01
  23c07d:       48 83 c1 20             add    rcx,0x20
  23c081:       49 39 c0                cmp    r8,rax
  23c084:       72 d5                   jb     23c05b <THFloatVector_add_AVX+0x5b>
```

First thing to notice is they both use `vaddps ymmX ymmX ymmX` which is a 256-bit AVX add (!) But you can see that while the AVX version does 256-bit loads (vmovups on ymmX), the DEFAULT version does 128-bit loads (vmovups on xmmX, and vinsertf128/vextractf128 to read/write the high bits). I'm not sure why gcc can't do 256-bit loads automatically.

If I unroll the DEFAULT verison further to do 16 at a time, the assembly is also unrolled (see below), but still only does 128-bit loads and stores. *shrug*.
```
   23c040:       c5 f8 10 4c 06 20       vmovups xmm1,XMMWORD PTR [rsi+rax*1+0x20]
  23c046:       c5 f8 10 14 06          vmovups xmm2,XMMWORD PTR [rsi+rax*1]
  23c04b:       c4 e3 75 18 4c 06 30    vinsertf128 ymm1,ymm1,XMMWORD PTR [rsi+rax*1+0x30],0x1
  23c052:       01
  23c053:       c4 e3 6d 18 54 06 10    vinsertf128 ymm2,ymm2,XMMWORD PTR [rsi+rax*1+0x10],0x1
  23c05a:       01
  23c05b:       c5 f4 58 cb             vaddps ymm1,ymm1,ymm3
  23c05f:       c5 ec 58 d3             vaddps ymm2,ymm2,ymm3
  23c063:       c5 f8 11 4c 07 20       vmovups XMMWORD PTR [rdi+rax*1+0x20],xmm1
  23c069:       c4 e3 7d 19 4c 07 30    vextractf128 XMMWORD PTR [rdi+rax*1+0x30],ymm1,0x1
  23c070:       01
  23c071:       c5 f8 11 14 07          vmovups XMMWORD PTR [rdi+rax*1],xmm2
  23c076:       c4 e3 7d 19 54 07 10    vextractf128 XMMWORD PTR [rdi+rax*1+0x10],ymm2,0x1
  23c07d:       01
  23c07e:       48 83 c0 40             add    rax,0x40
  23c082:       48 39 c8                cmp    rax,rcx
  23c085:       75 b9                   jne    23c040 <THFloatVector_add_AVX+0x40>
```

adamlerer:
P.S. The upshot of all this is you can probably get 2-3x speedup by writing an AVX version of contiguous copy, if src and dst are in cache.

(blocked transpose is more important though)

pavanky:
@adamlerer unless I am mistaken, isn't AVX just 128 bit while AVX2 is 256 bit?

Jokeren:
@adamlerer , @pavanky 
Thanks for discuss!
I am sorry that the benchmark is not as accurate as I expect. Something to claim:

1. I am talking about the `cadd` branch that handles three different pointers. It will only call `MKL` when the dest address is the same as one of the source addresses. 

2. `THFloatVector_add_AVX` and `THFloatVector_add_AVX2` are exactly the same thing. Because AVX2 instruction set does not change arithmetic operations for float and double data types. 

3. `the multi-thread version is very sensitive to OMP_NUM_THREADS`, this is caused by the setting of my benchmark. I fixed the array size so that with the growth of the number of threads, each of them gets smaller granularity. Anyway, I am messed up with the benchmark, finding it hard to model CPU and memory behaviors. Many variables, such as the array size, number of threads, iterations, and hardware features, affect the final performance.  

But I may finish `AVX copy` first. And I will look back into the problem, analyze assembly codes, build an accountable benchmark, and associate hardware features with sources code. I believe a benchmark is whatever meaningful for us to compare with numpy or other libraries.

Jokeren:
Notice that after changing `THVectorDispatch.c`, some of the `VSX` vector operations do not work because they are designed for old `THVectorDispatch.c`. For instance, I remove `diff` and `scale` functions that can be implemented by `cadd` and `mul`. In this way I match the names of vector functions with tensor operations.

lib/TH/vector/AVX.c
```diff
@@ -123,6 +136,19 @@ static void THDoubleVector_add_AVX(double *y, const double *x, const double c, c
   }
 }
 
+static void THFloatVector_copy_AVX(float *y, const float *x, const ptrdiff_t n) {
+  ptrdiff_t i;
+  ptrdiff_t off;
+  for (i=0; i<=((n)-16); i+=16) {
+    _mm256_storeu_ps(y+i, _mm256_loadu_ps(x+i));
+    _mm256_storeu_ps(y+i+8, _mm256_loadu_ps(x+i+8));
```
Jokeren: stream ops cannot be applied for memory alignment issues. In fact, I am curious about why not assign aligned memory for tensors?

fmassa: I thought they were aligned for sufficiently large tensors https://github.com/torch/torch7/blob/master/lib/TH/THGeneral.c#L233

Jokeren: @fmassa, Thank you!

I tried this. But it appears that some test cases, like `conv2` and `conv3`, cannot pass after using stream instructions. And I think a flag to indicate a tensor is aligned or not will be helpful.

soumith: even if a tensor is aligned, a narrowed tensor is likely not guaranteed to be aligned.

a = torch.ByteTensor(1024)
b = a:narrow(1, 1, 10) -- b is not aligned...

Jokeren:
@soumith , This branch is not activated?

soumith:
@Jokeren looks like it is: https://travis-ci.org/torch/torch7/builds/196482699

lib/TH/generic/THTensorMath.c
```diff
@@ -483,9 +491,7 @@ accreal THTensor_(prodall)(THTensor *tensor)
 void THTensor_(add)(THTensor *r_, THTensor *t, real value)
 {
   THTensor_(resizeAs)(r_, t);
-  if (((THTensor_(isContiguous)(r_) && THTensor_(isContiguous)(t)) ||
-    (THTensor_(isTransposed)(r_) && THTensor_(isTransposed)(t))) &&
-    THTensor_(nElement)(r_) == THTensor_(nElement)(t)) {
```
Jokeren: I rolled back these kinds of wrong optimizations.

First I think it is necessary to add more checks into torch.test, since the current test cannot detect these errors.

Then, to apply transposed optimizations. I should separate different situations. For **TH_TENSOR_APPLY2**, there are nine cases for `r_` and `t`--a tensor can be entire transposed, with holes, or contiguous. Thus, nine strategies should be applied. While I think transposed tensors could be handled in the same way as tensors with holes without loss of speed.

Jokeren:
Tensor bind the type of a tensor with methods such as ` isContiguous`, but indeed they are independent of types.

Jokeren:
Benchmark codes and results are updated. @adamlerer , @soumith, @pavanky, I hope this time the benchmark is more rigorous and reasonable.

**contiguous**
https://github.com/Jokeren/torch-tensor-benchmark/tree/master/contiguous

**noncontiguous**
https://github.com/Jokeren/torch-tensor-benchmark/tree/master/noncontiguous

**transposed**
https://github.com/Jokeren/torch-tensor-benchmark/tree/master/transposed

It seems that three times speedup is achieved with regard to transposed tensor operations, not to mention hundreds times speedup with the `fill` operation.

We have applied three ideas:

1. **Dimension collapsing** that compresses the dimension counter array.
2. **Multi-threading** for `TH_TENSOR_APPLY` with transposed tensors.
3. **Indexing reduction** for contiguous tensors in `TH_TENSOR_APPLY`s.

Jokeren:
By the way, I am glad to say that the transposed copy is faster than numpy. Source code used for comparison is shown as below:

```
import numpy as np
import time

a = np.ndarray((10000, 1000))
b = np.ndarray((1000, 10000))

start = time.time()
for i in range(1,100):
    b = np.transpose(a).copy()
end = time.time()
print(end - start)                    
```

lib/TH/generic/THTensor.c
```diff
@@ -510,6 +510,28 @@ void THTensor_(squeeze1d)(THTensor *self, THTensor *src, int dimension)
   }
 }
 
+int THTensor_(isTransposed)(const THTensor *self)
```
adamlerer: 1) I'm not sure if this code is correct in general. I think what you're asking is "is this a non-contiguous tensor with no holes". Is that what you meant? I think in that case it works except for a corner case when you have a stride of 0. For example,

```
t = torch.Tensor(10,1,32)
u = t:select(2,0):expand(10,32)
```
In this case you have size=[10,32] and stride=[32,0]. It does have holes.

2) This function is never used anywhere. Can you remove it? Since it's not used anywhere I don't know what it's semantics should be.

lib/TH/generic/THVectorDefault.c
```diff
@@ -17,68 +32,100 @@ void THVector_(fill_DEFAULT)(real *x, const real c, const ptrdiff_t n) {
     x[i] = c;
 }
 
+void THVector_(cadd_DEFAULT)(real *z, const real *x, const real *y, const real c, const ptrdiff_t n)
```
adamlerer: I think we should get rid of the hand-unrolling, but I'm okay saving it for a future PR.

lib/TH/generic/THVectorDefault.c
```diff
@@ -17,68 +32,100 @@ void THVector_(fill_DEFAULT)(real *x, const real c, const ptrdiff_t n) {
...
+{
+  ptrdiff_t i = 0;
+
+  for(; i<n-4; i+=4)
```
adamlerer: nit: spacing `i < n-4; i += 4` to be consistent (see throughout)

lib/TH/generic/THVectorDefault.c
```diff
@@ -17,68 +32,100 @@ void THVector_(fill_DEFAULT)(real *x, const real c, const ptrdiff_t n) {
...
   for(; i < n-4; i += 4)
   {
-    y[i] *= x[i];
-    y[i+1] *= x[i+1];
-    y[i+2] *= x[i+2];
-    y[i+3] *= x[i+3];
+    y[i] = x[i] / c;
+    y[i+1] = x[i+1] / c;
+    y[i+2] = x[i+2] / c;
+    y[i+3] = x[i+3] / c;
```
adamlerer: The naming cleanup here is great. 👍

lib/TH/vector/AVX.c
adamlerer: What do you mean by "stream instructions"?
If you want to do aligned instructions, can't you just check if the pointers are aligned?

```
if (!(x & 0x10) && !(y & 0x10)) {
  // do the aligned thing
} else {
  // something else
}
```

lib/TH/generic/THTensorMath.c
adamlerer: That's unfortunate, but I guess you're right.

lib/TH/THTensorApply.h
```diff
@@ -4,15 +4,17 @@
 #define TH_TENSOR_APPLY3(TYPE1, TENSOR1, TYPE2, TENSOR2, TYPE3, TENSOR3, CODE) \
 { \
   TYPE1 *TENSOR1##_data = NULL; \
```
adamlerer: wow, missed this file. I'll take a look tomorrow.

lib/TH/THTensorApply.h
```diff
@@ -370,77 +501,84 @@
  * Tensor. But because we are guaranteed the subsequent data is contiguous in memory, we
  * can simply loop for sizeof(A) iterations and perform the operation, without having to
  * follow the order described by the strides of A.
+ *
```
adamlerer: I think that 'outermost' should be called 'innermost' in line 485, and outer/inner switched in general. Usually the first index is referred to as outermost.

lib/TH/THTensorApply.h
```diff
@@ -370,77 +501,84 @@
...
+ * 3. We find the contiguous sections in a tensor and compress the counter array.
+ * For instance, if A is a 3x3x4x3 tensor, from which we get A' (3x3x3x3). Then, the
```
adamlerer: 3. As an optimization, we merge dimensions of A that are contiguous in memory. For example, if A is a 3x3x3x3 tensor narrowed from a 3x3x4x3 tensor, then the first two dimensions can be merged for the purposes of APPLY, reducing the number of nested loops.

lib/TH/THTensorApply.h
```diff
@@ -370,77 +501,84 @@
...
+ * original counter for A' is the same as A. But actually the first dimension and the
+ * second dimension can be collapsed into a single one.
```
adamlerer: As an additional optimization, you could sort dimensions in order of decreasing stride. This would mean you'd get maximally contiguous sections. But even more importantly, it guarantees maximum data locality, which is very important for cache efficiency.

I know you already handle transposed tensors above with a special case, but for example you might have
```
t = torch.Tensor(3,3,4,3).narrow(2,0,3).transpose(1,3)
```

lib/TH/THTensorApply.h
```diff
@@ -370,77 +501,84 @@
...
-    { \
-      if(TENSOR->size[TENSOR##_dim] != 1) \
-        break; \
-    } \
-    TENSOR##_stride = (TENSOR##_dim == -1 ? 0 : TENSOR->stride[TENSOR##_dim]); \
-\
-    /* what is the largest contiguous section? size will store the size of this section */ \
-    TENSOR##_size = 1; \
-    for(TENSOR##_dim = TENSOR->nDimension-1; TENSOR##_dim >= 0; TENSOR##_dim--) \
+    /* Find the dimension of contiguous sections */ \
```
adamlerer:
```
/* Find the number of contiguous sets of dimensions */

File: lib/TH/THTensorApply.h [hunk 18]
```diff
@@ -370,77 +501,84 @@
...
+    TENSOR##_strides[TH_TENSOR_dim_index] = TENSOR->stride[TENSOR->nDimension-1]; \
     /* TENSOR##_counter tracks where we are in the storage. The offset into the */ \
     /* storage is given by storage_offset + (i * j), where i is the stride */ \
     /* vector and j is tensor_counter vector. This sets the starting position for the loop. */ \
-    TENSOR##_counter = (long*)THAlloc(sizeof(long)*(TENSOR##_dim+1)); \
-    for(TENSOR##_i = 0; TENSOR##_i <= TENSOR##_dim; TENSOR##_i++) \
+    for(TENSOR##_i = TENSOR##_dim-1; TENSOR##_i >= 0; --TENSOR##_i) { \
       TENSOR##_counter[TENSOR##_i] = 0; \
+    } \
+    for(TENSOR##_i = TENSOR->nDimension-2; TENSOR##_i >= 0; --TENSOR##_i) { \
```
adamlerer: this is awesome

lib/TH/THTensorApply.h
```diff
@@ -286,16 +409,19 @@
 \
     if(TENSOR1##_i == TENSOR1##_size) \
     { \
-      if(TENSOR1##_dim == -1) \
+      if(TH_TENSOR1_contiguous == 1) \
+	break; \
```
adamlerer: This is a tab. Did you mix tabs/spaces elsewhere?

lib/TH/CMakeLists.txt
```diff
@@ -101,15 +101,21 @@ ENDIF(C_SSE2_FOUND)
 IF(C_SSE3_FOUND)
   SET(CMAKE_C_FLAGS "${C_SSE3_FLAGS} -DUSE_SSE3 ${CMAKE_C_FLAGS}")
 ENDIF(C_SSE3_FOUND)
+IF(C_AVX_FOUND)
+  SET(CMAKE_C_FLAGS "${C_AVX_FLAGS} -DUSE_AVX ${CMAKE_C_FLAGS}")
+ENDIF(C_AVX_FOUND)
+IF(C_AVX2_FOUND)
+  SET(CMAKE_C_FLAGS "${C_AVX2_FLAGS} -DUSE_AVX2 ${CMAKE_C_FLAGS}")
+ENDIF(C_AVX2_FOUND)
```
Jokeren: Yes, I agree

lib/TH/generic/THTensor.c
Jokeren: Well, I tried `t:select(2,0):expand(10,32)`, but it appears that torch does not allow to do that.

> stdin:1:` bad argument #2 to 'select' (out of range at

lib/TH/generic/THVectorDefault.c
Jokeren: Sure

lib/TH/THTensorApply.h
```diff
@@ -34,70 +36,131 @@
...
-      if(TENSOR1->size[TENSOR1##_dim] != 1) \
-        break; \
-    } \
-    TENSOR1##_stride = (TENSOR1##_dim == -1 ? 0 : TENSOR1->stride[TENSOR1##_dim]); \
     TENSOR1##_size = 1; \
-    for(TENSOR1##_dim = TENSOR1->nDimension-1; TENSOR1##_dim >= 0; TENSOR1##_dim--) \
-    { \
-      if(TENSOR1->size[TENSOR1##_dim] != 1) \
+    TENSOR1##_stride = 1; \
+    for(TENSOR1##_i = TENSOR1->nDimension-1; TENSOR1##_i >= 0; TENSOR1##_i--) { \
```
Jokeren: Yes. During the compilation period, torch generates different types of `isCoutiguous` function. That is to say, `double`, `float` or `byte` are appended before `isContiguous`. However, indeed the function is independent of types. Therefore, we cannot use `isContiguous` in `THTensorApply2` or `THTensorApply3` when array types are different

lib/TH/THTensorApply.h
Jokeren: My mistake. I'll examine and fix that. Is there a quick way to fix code styles?

lib/TH/THTensorApply.h
Jokeren: Sure. I will add the optimization later. Maybe not in this PR.

Jokeren:
Dear @adamlerer , thanks a lot for your kind review!

I will answer some questions here as Github do not allow me to comment. I do not know why.

> My main question is, how are you testing all the SIMD extensions? Since you have runtime dispatch, I expect you should be able to have the relevant test (or all tests) run in all the different modes (AVX, SSE, default). Are there tests to do that already?

I downloaded torch without AVX and benchmark it to compare with the AVX version I used. So Actually I compare AVX with SSE.

> On a different note, can you explain why fill performance improved so much relative to the other ops? 1000x speedup is a bit surprising... and this corresponds to 200GB/s of fill, which I don't believe :)

The surprising speedup is caused by cache readings. I tried `fill(1.0)` that results in the same running time. My processor--E5 2680v3 has a large L3 cache (about 30MB), so when I increase the test size from 1e7 to 1e9, 40GB/s bandwidth is achieved.

> What do you mean by "stream instructions"?

I mean the stream load and store instructions that directly load and store data between main memory and registers, but I think they are not necessary as we need to use cache mostly.

> This may be a stupid question, but I don't understand why your changes should speed up apply for transposed 2-dimensional tensors.

This is a good question. Let's compare two versions of `THTensorApply`. Take a look at the `TENSOR##_size` variable.

Original

```
    TENSOR##_size = 1; \
    for(TENSOR##_dim = TENSOR->nDimension-1; TENSOR##_dim >= 0; TENSOR##_dim--) \
    { \
      if(TENSOR->size[TENSOR##_dim] != 1) \
      { \
        if(TENSOR->stride[TENSOR##_dim] == TENSOR##_size) \
          TENSOR##_size *= TENSOR->size[TENSOR##_dim]; \
        else \
          break; \
      } \
    } \
```

Optimized

`TENSOR3##_size = TENSOR3##_sizes[TENSOR3##_dim-1]`

So in the original version, `TENSOR##_size` is assigned to the size of the **longest contiguous region**, while `TENSOR##_size` is assigned to the size of the **last region** in the optimized version. Think of a transposed matrix of which the stride is transpoed from [10, 1] to [1, 10]. Then, `TENSOR##_size=1` for the original version. So the original version will handle the `counter` array every time the stride is added, while the optimized verion does not. I think the optimized version actually fixes a performance bug instead of devising a new idea. 

```
    /* Loop through the contiguous section of the Tensor */ \
    for(TENSOR##_i = 0; TENSOR##_i < TENSOR##_size; TENSOR##_i++, TENSOR##_data += TENSOR##_stride) /* 0 et pas TENSOR##_dim! */ \
    { \
      CODE \
    } \
\
\
    /* Handle corner case where the entire Tensor was contiguous */ \
    if(TENSOR##_dim == -1) \
       break; \
 \
    /* Reset pointer to beginning of loop */ \
    TENSOR##_data -= TENSOR##_i*TENSOR##_stride; \
    for(TENSOR##_i = TENSOR##_dim; TENSOR##_i >= 0; TENSOR##_i--) \
```

After fix the mentioned issues, I will rebase my branch to solve conflict and prepare for merge.

lib/TH/THTensorApply.h
```diff
@@ -34,70 +36,131 @@
...
+      if(TENSOR1->size[TENSOR1##_i] != 1) { \
+        if(TENSOR1->stride[TENSOR1##_i] == TENSOR1##_size) \
+          TENSOR1##_size *= TENSOR1->size[TENSOR1##_i]; \
+        else{ \
+          TH_TENSOR1_contiguous = 0; \
+	  break; \
+	} \
+      } \
+    } \
+    if (!TH_TENSOR1_contiguous) { \
```
apaszke: Wouldn't it be better to separate these parts into a different macro and just instantiate it 3 times in here? There seems to be a lot of redundancy.

lib/TH/THTensorApply.h
apaszke: Depends on the editor. In vim you can just run `:retab`.

lib/TH/generic/THTensor.c
apaszke: I think Adam meant `:select(3, 1)` (1 vs 0-based indexing)

lib/TH/generic/THTensorCopy.c
```diff
@@ -4,7 +4,20 @@
...
+  if (THTensor_(isContiguous)(tensor) && THTensor_(isContiguous)(src) && THTensor_(nElement)(tensor) == THTensor_(nElement)(src)) {
+    real *sp = THTensor_(data)(src);
+    real *rp = THTensor_(data)(tensor);
+    ptrdiff_t sz = THTensor_(nElement)(tensor);
+#ifndef TH_REAL_IS_HALF
+    THVector_(copy)(rp, sp, sz); 
+#else
+    ptrdiff_t i;
+    for (i=0; i<sz; ++i)
+      rp[i] = sp[i];
```
apaszke: Isn't it better to use `memcpy` here?

lib/TH/generic/THTensorMath.c
apaszke: I thought that using C99 is fine, isn't it?

lib/TH/generic/THTensorMath.c
```diff
@@ -592,13 +639,22 @@ void THTensor_(cadd)(THTensor *r_, THTensor *t, real value, THTensor *src)
...
+        #ifdef _OPENMP
+        size_t num_threads = omp_get_num_threads();
+        size_t tid = omp_get_thread_num();
+        #else
+        size_t num_threads = 1;
+        size_t tid = 0;
+        #endif
+        ptrdiff_t i = tid * (sz / num_threads);
+        ptrdiff_t i_end = tid == num_threads - 1 ? sz : i + sz / num_threads;
+	THVector_(cadd)(rp+i, tp+i, sp+i, value, i_end-i);
```
apaszke: nit: you have a tab here

lib/TH/generic/simd/simd.h
```diff
@@ -9,7 +9,7 @@
 #endif
 
 // Can be found on Intel ISA Reference for CPUID
-#define CPUID_AVX2_BIT 0x10       // Bit 5 of EBX for EAX=0x7
+#define CPUID_AVX2_BIT 0x20       // Bit 5 of EBX for EAX=0x7
```
apaszke: Are you sure about that? `0x20` is 6th bit, which is:

1. out of sync from the comment
2. incorrect according to [CPUID page on Wikipedia](https://en.wikipedia.org/wiki/CPUID)

lib/TH/vector/AVX.c
apaszke: Is it worth it? I think I've seen some benchmarks where both kinds of loads were equally fast for both aligned and unaligned memory, but aligned loads had a much higher overhead on unaligned accesses.

lib/TH/THTensorApply.h
adamlerer: Oh yes, you're right, because the type of TENSORX may not be the same as the type of the outer function :(

lib/TH/generic/simd/simd.h
adamlerer: @apaszke I think @Jokeren is right here: `1<<5 = 0x20`.

adamlerer:
@Jokeren thanks for the detailed responses. It took me a while to see what you meant about the original TH_TENSOR_APPLY... but you're right! The old version has a big performance big because `TENSORX##_size` should always be at least as big as the last dimension, but originally it was 1 for transposed tensors, yikes! Great catch.

Btw, if you haven't seen it, I was writing some related code yesterday:
https://github.com/torch/torch7/pull/934/

The code for `max` and `min` is not going to work now that you've changed `TH_TENSOR_APPLY`, so maybe we want to add some modified kernels for these reductions (`max`, `min`, `cumsum`, etc.).

lib/TH/THTensorApply.h
Jokeren: Thanks! I will try

lib/TH/generic/THTensorCopy.c
Jokeren: Sure!

lib/TH/generic/THTensorMath.c
Jokeren: @apaszke , Could I use `C99` flag to compile `THTensorMath.c`? As `C99` is not enabled for most files.

lib/TH/generic/simd/simd.h
apaszke: But you want to shift `0x1` by 4, not by 5, right? (first bit + shift 4 = 5th bit).

apaszke: Nvm, they're numbered from 0. My bad.

lib/TH/generic/THTensorMath.c
apaszke: Yes, C99 is fine. You can use it in all files that need it, not sure why we're not compiling everything with it enabled.

Jokeren:
Hi, all!

1. Following the instructions of @apaszke , I used C99 to clean up *openmp* pragmas. Beyond the previous modifications, I could also clean up other *openmp* usages in `THTensorMath.c`.

2. Add comments for https://github.com/torch/torch7/pull/934.

3. Fixed the `stride=0` case in `isTranposed` function.

4. Fixed coding styles.

soumith:
this is starting to look good. I'll do some final checks today and merge today/tomorrow.

lib/TH/THTensorApply.h
```diff
@@ -200,20 +272,25 @@
       TENSOR3##_i = 0; \
     } \
   } \
-  THFree(TENSOR1##_counter); \
-  THFree(TENSOR2##_counter); \
-  THFree(TENSOR3##_counter); \
+  if(TH_TENSOR1_contiguous) \
```
adamlerer: Shouldn't this be

```
if(!TH_TENSOR1_contiguous) \
```
?

Jokeren: Thanks, it is fixed.

soumith:
Thanks a TON to @Jokeren , @adamlerer and @rguthrie3 for making this happen. This is now merged into master!!!!!!!

PR #860, opened 2016-12-02
Prevent runover refcount of tensor

drahnr:
This is necessary since for integrations with C one never knows what or when the lua gc kicks in.

### DIFF
--- a/lib/TH/generic/THTensor.c
+++ b/lib/TH/generic/THTensor.c
@@ -598,7 +598,7 @@ void THTensor_(free)(THTensor *self)
   if(!self)
     return;
 
-  if(self->flag & TH_TENSOR_REFCOUNTED)
+  if((self->flag & TH_TENSOR_REFCOUNTED) && (THAtomicGet(&self->refcount) > 0))
   {
     if(THAtomicDecrementRef(&self->refcount))
     {


apaszke:
Why is that necessary? If you see refcount == 0 it likely means that the tensor struct has already been freed, and you're doing invalid memory accesses anyway (also, you're quite likely to see some garbage and stomp on reused memory because of the decref). If you want to use the tensor for time longer than the ffi call, you should call `THTensor_(retain)` and free it manually once you stop processing it.

drahnr:
True, one would access invalidated memory.

In my case I run a C Application which embeds luajit.

I run my forward pass on a video stream and what I saw was a non deterministic behaviour of recounts on the returned tensor from a lua function to C (no ffi).

apaszke:
Well, accessing freed memory sounds like a pretty serious thing to me - it's completely UB. If it happens to be read as a positive `long` you will also decref it (destroying its old contents), or if it happens to be 1, you'll also try to free it. I don't think that adding these checks is a good idea. If you need them, you likely have some logic error in your program.

Of course it might be "nondeterministic", because the GC might sometimes decide that it doesn't care and it's not the time to free the tensor yet. Still, my advice is the same, if you need to ensure that the tensor is alive for some time in your C code, just `retain` and `free` it manually. These operations are  cheap and you'll be sure that your program will work as expected every time.

drahnr:
The problem is not that I want to retain them, I want to free the mem since I am working in a memory constrained env. So every tensor that survives an iteration of my 60 fps use case is an actual problem.

I understand that undef mem access is as bad, but in that case there should be asserts added.

soumith:
@drahnr if the GC non-determinism is really a problem, you can free the tensor after it's use in lua-land with `:free()`. I dont think hacking around the refcount like this is really a good solution.

For example:

```lua
while true do
   frame = receive()
   -- let's say frame is a torch.*Tensor received from the C side   
   -- do something with frame
   -- now, you are done with using frame
    frame:free()
end
```

apaszke:
@soumith I don't think that's going to work. `free` only decrefs the refcount but it doesn't invalidate the lua object, so the GC will try to free it once more for sure. `frame:set()` will detach the tensor, and free its memory immediately.

apaszke:
@drahnr I would say that invalid memory accesses are one of the biggest problems you can possibly have in your program! The only way for them to be non-destructive is when you actually happen to hit them only when they're in the allocator cache (good luck). If they have been reused you'll end up modifying the bits of some random objects because of the decref. This is a terrible bug, as it's going to silently corrupt your state and crash your program only sometime later. If they were truly freed you'll get a SIGSEV. Both of them seem pretty bad to me.

drahnr:
The `:set()` trick only works for passing an tensor as argument, but not for the case when returning the argument.

My use case is not a lua for loop, but a C for loop with one call per iteration to lua.
```

while(1) {
input = <alloc tensor>
luaT_push(input, ...)
lua_pcall(...)
luaT_pop(output, ...)
...
}
```

The only thing that worked to some extent was static allocation of two tensors in C and then resizing the output tensor as needed in lua. This is not really what I was looking for.

drahnr:
Please see related https://github.com/torch/torch7/issues/862

PR #873, opened 2016-12-21
Different code path for contiguous tensors along dimension 1 in cat + fix cat bug + fix ambiguity

nkoumchatzky:
This is very useful for speed reasons, the generic code path has too many function calls.


soumith:
there's an additional optimization that would be really useful and very simple to add to this PR. If you think you will do it, let me know -- if not i'll merge your PR and do it.

The same code path (all contiguous, memcpy) can be taken if the dim to be cat is the 1st dim, i.e.:
torch.cat( {1 x 10 x 10, 1 x 10 x 10, 1 x 10 x 10}, 1)

Your 1D case is just a  small special case of that.

nkoumchatzky:
Indeed, I figured I'd leave it for a next review :) I'll do it in this one.

Happy holidays!

Nico

nkoumchatzky:
@soumith @pavanky Here is the augmented patch.

nkoumchatzky:
I actually spotted a bug in cat, let me fix it before merging.

lib/TH/generic/THTensorMath.c
```diff
@@ -2035,24 +2035,29 @@ void THTensor_(catArray)(THTensor *result, THTensor **inputs, int numInputs, int
   THLongStorage *size;
   int i, j;
   long offset;
-  int ndim = dimension + 1;
-  for (i = 0; i < numInputs; i++)
-  {
-    ndim = THMax(ndim, inputs[i]->nDimension);
+  int minDim = dimension + 1;
```
soumith: you can now remove minDim, it is no longer used.

nkoumchatzky:
Ok the bug was: when catting an empty tensor along the first dim, not exception was thrown and the size was incremented by 1. When catting along another dim, an exception was thrown.
I chose to allow empty tensors, just ignoring them when catting. Catting empty tensors only results in empty tensors, whichever the input dim.

Let me know if the behavior is ok @soumith

fmassa:
@nkoumchatzky I think allowing empty tensors in `torch.cat` is a reasonable behaviour. But we should adapt the cuda version to follow the same behaviour.

nkoumchatzky:
@fmassa Yes indeed. @pavanky will push a similar patch in cutorch soon.

pavanky:
@nkoumchatzky @fmassa 

Since we are allowing empty tensors, shouldn't the following also be allowed?

```
torch.cat({troch.Tensor(), torch.Tensor()})
```

This is currently (after applying the patch in this PR) throwing an error saying that the dimension is invalid. I think returning an empty tensor should be the right behavior.

nkoumchatzky:
@pavanky Just pushed a new patch that deals with that use case. Now a -1 or unspecified dimension will cat along the maximum of the last dimension over the input tensors, and will allow for empty tensors.

soumith:
@nkoumchatzky all unit tests are failing: https://travis-ci.org/torch/torch7/builds/186789362

nkoumchatzky:
@soumith I added the forgotten TensorMath.lua file, should work now.

pavanky:
@nkoumchatzky @soumith updated cutorch to have same behavior.

soumith:
thanks Nico!

PR #874, opened 2016-12-21
Add support for torch.HalfTensor

gchanan:
NOTE: this should not be merged until the cutorch equivalent is ready; I'm working on that based on https://github.com/torch/cutorch/pull/578.  This pull request won't break anything but because cutorch's generics overwrite the torch generics, you can lose some HalfTensor functionality via "require cutorch"

This is based on https://github.com/torch/torch7/pull/826/ (which is the first commit in this request), with the following improvements/simplifications (the second commit):

Improvements/Simplifications:
1) No longer generates math functions that are not actually defined
on torch.HalfTensor, e.g. maskedFill, map2, etc.
2) Adds tests for all available torch.HalfTensor functions.
3) Allows compiling without TH_GENERIC_USE_HALF (so if there's a
problem can just unset that in CMakeLists rather than backing out)
4) Some simplifications: removes a new copy optimization and
some TH_HALF literal definitions; I'd rather not include optimizations
to non-half types in this PR.

Limitations:
Because match functions are not defined, some "non-math" operators
on torch.HalfTensor give an error message, e.g. __index__/__newindex__
with a ByteTensor apply a mask, but masks aren't implemented.  These
limitations aren't always obvious (e.g. for documentation purposes), but
they should always give an error message.


gchanan:
This should be ready to merge once the cutorch changes are ready.  We now define a different TH_HALF type instead of defining 'half' as in cutorch; defining 'half' requires that files are included in the proper order so that cutorch does not redefine 'half'; this is probably doable, but tricky for downstream projects to get right.

gchanan:
equivalent cutorch PR here: https://github.com/torch/cutorch/pull/655

borisfom:
Thanks for completing this! Yes using separate type for Half makes better sense (I started with that option, too). It should make it simpler to adopt a tricky __fp16 gcc ARM type, too.

PR #875, opened 2016-12-22
Add a new function bhistc to calculate histogram of batch of images only once

atkayu:
[ Description ]
When a tensor contains multiple images, and one need to calculate the histogram of each image in this two dimension tensor. If using the original `torch.histc` function, should have to write a loop to calculate them. This way is not effective enough. So I write a new function named by `torch.histc2` to output the histogram tensor only once, the input is the 2D tensor, each row indicates an image, the default bins is 100 as the original `histc`, calculate the histogram along the last dimension of input tensor.

[ Code Change ]
* In `lib/TH/generic/THTensorMath.c`, add a new math function `void THTensor_(histc2)`
* In `lib/TH/generic/THTensorMath.h`, add a statement for `void THTensor_(histc2)`
* In `TensorMath.lua`, add the `histc2` for `wrap`
* In `test/test.lua`, add a test for new function `histc2`
* In `doc/maths.md`, add a new description for `torch.histc2` with some toy use instances

[ Compile ]
All these changes can be compiled pass locally.

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -1088,6 +1088,14 @@ static void THTensor_random1__(THTensor *self, THGenerator *gen, long b)
             {name="long",default=100},
             {name="double",default=0},
             {name="double",default=0}})
+      
+      wrap("histc2",
+           cname("histc2"),
+           {{name=Tensor, default=true, returned=true},
+            {name=Tensor},
+            {name="long",default=100},
+            {name="double",default=0},
+            {name="double",default=0}})
 
       wrap("norm",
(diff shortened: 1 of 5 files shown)


soumith:
there are a few functions that batch an operation, like bmm, baddmm, baddbmm.
To keep with the same convention, I suggest that you rename histc2 to bhistc

atkayu:
I have renamed histc2 to bhistc in all code files. Yeah~, this name is more formal. :)

soumith:
thanks!

PR #882, opened 2016-12-28
Add support for VSX vector instructions on PPC

amrobbins:
Added support for the fill, diff, scale, mul and add functions using
PPC CPU vector instructions. These are used in place of the versions
of these functions written for x86, when compiled on PPC.

This fixes a compile failure on PPC


soumith:
thanks!

PR #900, opened 2017-01-06
Fix contiguous assertion in View implementation

karthikncode:
Replace the assertion in View implementation with automatic cloning if tensor not contiguous.

### DIFF
--- a/Tensor.lua
+++ b/Tensor.lua
@@ -458,7 +458,7 @@ function Tensor.view(result, src, ...)
    local origNElement = tensor:nElement()
    size = specifyFully(size, origNElement)
 
-   assert(tensor:isContiguous(), "expecting a contiguous tensor")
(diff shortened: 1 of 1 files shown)


fmassa:
Hum, I wonder if this is the desired behaviour.
There is an assumption that `:view` returns a view of the same storage, so cloning it might lead to unwanted behaviour at some places.

karthikncode:
Not sure if it breaks anything, but I was getting the assertion error with nn.View otherwise.

karthikncode:
Ah, I see. So maybe not a global change then?

fmassa:
Can't you use a `nn.Contiguous` before `nn.View`?

karthikncode:
Ah - didn't realize there was one for that. Sorry about that!

PR #901, opened 2017-01-07
half<->float conversion cleanup

borisfom:
Refactored _raw methods from TH_half2float and TH_half2float - primarily for reusability in cutorch. 
Fixed Inf constant for fp16.


lib/TH/generic/THTensorMath.c
```diff
@@ -99,7 +99,7 @@ void THTensor_(nonzero)(THLongTensor *subscript, THTensor *tensor)
   long dim;
   long div = 1;
 #ifdef TH_REAL_IS_HALF
-#define IS_NONZERO(val) (TH_half2float(val)!=0)
+#define IS_NONZERO(val) (val.x!=TH_HALF_ZERO)
```
gchanan: I don't think this is equivalent -- previous code fails != check for 0.0f and -0.0f.  Current code only fails for 0.0f.  See e.g. https://github.com/numpy/numpy/blob/d9c1a1f91a3983469de7086ff589a6aeaa353b3e/numpy/core/src/npymath/halffloat.c#L50-L53

lib/TH/THHalf.h
```diff
@@ -24,17 +24,26 @@ typedef struct __thalign__(4) {
...
+TH_API float  TH_half2float(THHalf);
 
 #ifndef TH_HALF_BITS_TO_LITERAL
 # define TH_HALF_BITS_TO_LITERAL(n) { n }
 #endif
 
-#define TH_HALF_MAX TH_HALF_BITS_TO_LITERAL(0x7BFF)
+#define TH_HALF_ZERO 0x0U
+#define TH_HALF_MIN  0x0400U
+#define TH_HALF_MAX  0x7BFFU
```
gchanan: why do you want all these definitions that aren't used (excluding TH_INF, which is clearly a useful addition now)?  Is there something specific you are trying to do in the future?

borisfom: Yes, min() and max() were used in my working branch for cutorch math refactor.

lib/TH/generic/THTensorMath.c
borisfom: You're right; pushed the fix

gchanan:
FYI I think this is now conflicting because of https://github.com/torch/torch7/commit/3caabb9fabf00f2517d147b307718a0741a20f1b -- which changes the conversion functions slightly to avoid strict aliasing warnings.

lib/TH/THHalf.c
```diff
@@ -49,14 +49,29 @@
...
  */
 
+THHalf TH_float2half(float f)
+{
+  THHalf h;
+  TH_float2halfbits(&f, &h.x);
+  return h;
+}
+
+TH_API float  TH_half2float(THHalf h)
```
gchanan: extra spaces.

lib/TH/THHalf.h
```diff
@@ -24,17 +24,26 @@ typedef struct __thalign__(4) {
...
-/* numeric limits */
+typedef unsigned short THHalfBits;
 
+TH_API void TH_float2halfbits(float*, THHalfBits*);
+TH_API void TH_halfbits2float(THHalfBits*, float*);
 
-TH_API THHalf TH_float2half(float a);
-TH_API float TH_half2float(THHalf a);
+TH_API THHalf TH_float2half(float);
+TH_API float  TH_half2float(THHalf);
```
gchanan: extra spaces

lib/TH/THHalf.h
```diff
@@ -24,17 +24,26 @@ typedef struct __thalign__(4) {
 typedef __THHalf THHalf;
 typedef __THHalf2 THHalf2;
 
-/* numeric limits */
+typedef unsigned short THHalfBits;
```
gchanan: tbh I'd probably just get rid of the typedef; we don't typedef the actual definition, so it doesn't buy us much.  And I don't think we'd want to typedef the definition given ffi/general complexity.

lib/TH/THHalf.h
```diff
@@ -24,17 +24,26 @@ typedef struct __thalign__(4) {
...
+TH_API THHalf TH_float2half(float);
+TH_API float  TH_half2float(THHalf);
 
 #ifndef TH_HALF_BITS_TO_LITERAL
 # define TH_HALF_BITS_TO_LITERAL(n) { n }
 #endif
 
-#define TH_HALF_MAX TH_HALF_BITS_TO_LITERAL(0x7BFF)
+#define TH_HALF_ZERO 0x0U
+#define TH_HALF_MIN  0x0400U
```
gchanan: this doesn't match the bit definition in THCNumerics (but max does)?

lib/TH/THHalf.h
```diff
@@ -24,17 +24,26 @@ typedef struct __thalign__(4) {
...
-TH_API THHalf TH_float2half(float a);
-TH_API float TH_half2float(THHalf a);
+TH_API THHalf TH_float2half(float);
+TH_API float  TH_half2float(THHalf);
 
 #ifndef TH_HALF_BITS_TO_LITERAL
 # define TH_HALF_BITS_TO_LITERAL(n) { n }
 #endif
 
-#define TH_HALF_MAX TH_HALF_BITS_TO_LITERAL(0x7BFF)
```
gchanan: why did you decide to get rid of the LITERAL macro in the definitions?

borisfom: This is for constants to be usable in functions that operate on underlying bits, not THHalf and half types.

pavanky: But something like this would not work:

```
half minVal = TH_HALF_MAX;
```

There may be a need to implement both versions.

borisfom: I don't think we have to implement extra convenience layer. 
It can be done via half minVal = TH_HALF_BITS_TO_LITERAL(TH_HALF_MAX)

lib/TH/THHalf.h
borisfom: Welll, min() in THCNumerics corresponded to numeric_limits::lowest(), not ::min().
Means: you were right and we should not define those here.

borisfom:
@gchanan : should be all ready to go now.

borisfom:
@gchanan : can we finalize this?

lib/TH/generic/THTensorMath.c
```diff
@@ -99,7 +99,7 @@ void THTensor_(nonzero)(THLongTensor *subscript, THTensor *tensor)
   long dim;
   long div = 1;
 #ifdef TH_REAL_IS_HALF
-#define IS_NONZERO(val) (TH_half2float(val)!=0)
```
gchanan: use TH_HALF_ZERO? (it's unused)

lib/TH/THHalf.c
```diff
@@ -76,37 +91,31 @@ float TH_half2float(THHalf h)
...
 
-THHalf TH_float2half(float f)
+void TH_float2halfbits(float* src, unsigned short* dest)
 {
-    THHalf ret;
-
-    unsigned x;
-    memcpy(&x,&f,sizeof(f));
-    unsigned u = (x & 0x7fffffff), remainder, shift, lsb, lsb_s1, lsb_m1;
+    unsigned u = *(unsigned*)src;
```
gchanan: you are missing  & 0x7fffffff, see the old version.  This causes test/test_half.lua to fail.

lib/TH/generic/THTensorMath.c
borisfom: Well, I think using macro for full value   in this context is misleading : 
((val.x & 0x7fff) != 0)

lib/TH/THHalf.c
borisfom: Fixed, thanks!

lib/TH/generic/THTensorMath.c
gchanan: that's fine, I was just suggesting you either use TH_HALF_ZERO or get rid of it.

soumith:
thanks!

PR #908, opened 2017-01-11
simd.h: really fix the arm64 (i.e. Aarch64) build

CDLuminate:
`__NEON__` is an ISA indicator instead of architecture indicator.


soumith:
thanks Lumin! makes sense.

Issue #876, opened 2016-12-22
error while installing torch


alirezaniki:
Hi
currently i am trying to install TORCH in my machine ( ubuntu 15.04), but every time i tried i got the following errors when i ran the command ./install.sh 
-------------------------------
Scanning dependencies of target luaT
Scanning dependencies of target TH
[  3%] Building C object lib/luaT/CMakeFiles/luaT.dir/luaT.c.o
[  6%] Building C object lib/TH/CMakeFiles/TH.dir/THGeneral.c.o
[  9%] Building C object lib/TH/CMakeFiles/TH.dir/THAllocator.c.o
[ 12%] Linking C shared library libluaT.so
Building C object lib/TH/CMakeFiles/TH.dir/THStorage.c.o
[ 12%] Built target luaT
[ 15%] Building C object lib/TH/CMakeFiles/TH.dir/THTensor.c.o
[ 18%] Building C object lib/TH/CMakeFiles/TH.dir/THBlas.c.o
[ 21%] Building C object lib/TH/CMakeFiles/TH.dir/THLapack.c.o
[ 25%] Building C object lib/TH/CMakeFiles/TH.dir/THLogAdd.c.o
[ 28%] Building C object lib/TH/CMakeFiles/TH.dir/THRandom.c.o
[ 31%] Building C object lib/TH/CMakeFiles/TH.dir/THFile.c.o
[ 34%] Building C object lib/TH/CMakeFiles/TH.dir/THDiskFile.c.o
[ 37%] Building C object lib/TH/CMakeFiles/TH.dir/THMemoryFile.c.o
[ 40%] Building C object lib/TH/CMakeFiles/TH.dir/THAtomic.c.o
[ 43%] Building C object lib/TH/CMakeFiles/TH.dir/THVector.c.o
In file included from /home/sarhad1/torch/pkg/torch/lib/TH/THVector.c:2:0:
/home/sarhad1/torch/pkg/torch/lib/TH/generic/THVectorDispatch.c: In function ‘THByteVector_vectorDispatchInit’:
/home/sarhad1/torch/pkg/torch/lib/TH/generic/simd/simd.h:72:3: error: inconsistent operand constraints in an ‘asm’
   asm volatile ( "cpuid\n\t"
   ^
/home/sarhad1/torch/pkg/torch/lib/TH/generic/simd/simd.h:72:3: error: inconsistent operand constraints in an ‘asm’
   asm volatile ( "cpuid\n\t"
   ^
[...]

PR #924, opened 2017-02-02
Fix AVX2 detect bug

Jokeren:
It seems unnecessary to use GCC *builtins*, as the function `__get_cpuid` calls an asm statement internally. Therefore, I use the asm statement directly, avoiding the possible GCC bug. 

Another fix is the flag of AVX2.

### DIFF
--- a/lib/TH/generic/simd/simd.h
+++ b/lib/TH/generic/simd/simd.h
@@ -9,7 +9,7 @@
 #endif
 
 // Can be found on Intel ISA Reference for CPUID
-#define CPUID_AVX2_BIT 0x10       // Bit 5 of EBX for EAX=0x7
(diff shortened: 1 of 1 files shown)


soumith:
the reason I started using `__get_cpuid` was to get around this issue:
https://github.com/torch/torch7/issues/876

My patch is here: https://github.com/torch/torch7/pull/886
and was made after @ius had correctly pointed out the problem: https://github.com/torch/torch7/issues/876#issuecomment-269080419

Your PR will reintroduce this issue.
What is the solution...?
Maybe we can have a compile-time snippet that checks if the asm will compile, if yes, use that codepath, if not fallback to GCC intrinsic

soumith:
see https://github.com/torch/torch7/pull/943

PR #926, opened 2017-02-06
Add code for 'view' to TH

colesbury:


lib/TH/THStorage.c
```diff
@@ -12,3 +12,56 @@
...
+}
+
+TH_API THLongStorage *THLongStorage_newInferSize(THLongStorage *size, ptrdiff_t nElement)
+{
+  ptrdiff_t total_size = (size->size > 0 ? 1 : 0);
+  ptrdiff_t dim_infer = -1;
+  ptrdiff_t i;
+  for (i = 0; i < size->size; i++) {
+    if (size->data[i] == -1) {
+      THArgCheck(dim_infer == -1, 1, "only one dimension can be inferred");
```
soumith: improve the error message here. Only one dimension can be inferred, but found -1 in more than one place.

PR #934, opened 2017-02-10
Speed up reductions on non-contiguous dimensions

adamlerer:
I just did `sum`, `prod`, `mean`, `max`, and `min` for now, but you can do a bunch of the other pointwise reductions like this too. 

before:
```
th> t = torch.randn(64,1024,1024)
                                                                      [5.6855s]
th> t:max(1);
                                                                      [1.7969s]
th> t:max(2);
                                                                      [0.9834s]
th> t:max(3);
                                                                      [0.1487s]
th> t:sum(1);
                                                                      [1.5895s]
th> t:sum(2);
                                                                      [0.9665s]
th> t:sum(3);
                                                                      [0.1610s]
```

after:
```
th> t = torch.randn(64,1024,1024)
                                                                      [5.7820s]
th> t:max(1);
                                                                      [0.3097s]
th> t:max(2);
                                                                      [0.2392s]
th> t:max(3);
                                                                      [0.1456s]
th> t:sum(1);
                                                                      [0.1985s]
th> t:sum(2);
                                                                      [0.2219s]
th> t:sum(3);
```

I think the optimizations for `max` and `min` will conflict with the TH_TENSOR_APPLY improvements in https://github.com/torch/torch7/pull/852, but we can always add special macros for this purpose if we want to keep the max and min optimizations. That new macro could also support stuff like `cumsum`.

### DIFF
--- a/lib/TH/THTensorApply.h
+++ b/lib/TH/THTensorApply.h
@@ -1,484 +1,6 @@
 #ifndef TH_TENSOR_APPLY_INC
 #define TH_TENSOR_APPLY_INC
 
-#define TH_TENSOR_APPLY3(TYPE1, TENSOR1, TYPE2, TENSOR2, TYPE3, TENSOR3, CODE) \
-{ \
-  TYPE1 *TENSOR1##_data = NULL; \
-  long *TENSOR1##_counter = NULL, *TENSOR1##_sizes = NULL, *TENSOR1##_strides = NULL; \
-  long TENSOR1##_stride = 0, TENSOR1##_size = 0, TENSOR1##_dim = 0, TENSOR1##_i, TENSOR1##_n; \
-  TYPE2 *TENSOR2##_data = NULL; \
-  long *TENSOR2##_counter = NULL, *TENSOR2##_sizes = NULL, *TENSOR2##_strides = NULL; \
-  long TENSOR2##_stride = 0, TENSOR2##_size = 0, TENSOR2##_dim = 0, TENSOR2##_i, TENSOR2##_n; \
-  TYPE3 *TENSOR3##_data = NULL; \
-  long *TENSOR3##_counter = NULL, *TENSOR3##_sizes = NULL, *TENSOR3##_strides = NULL; \
-  long TENSOR3##_stride = 0, TENSOR3##_size = 0, TENSOR3##_dim = 0, TENSOR3##_i, TENSOR3##_n; \
-  int TH_TENSOR_APPLY_hasFinished = 0; \
-  int TH_TENSOR1_contiguous = 1, TH_TENSOR2_contiguous = 1, TH_TENSOR3_contiguous = 1; \
-  long TH_TENSOR_dim_index = 0; \
-\
-  TENSOR1##_n = (TENSOR1->nDimension ? 1 : 0); \
-  for(TENSOR1##_i = 0; TENSOR1##_i < TENSOR1->nDimension; TENSOR1##_i++) \
-    TENSOR1##_n *= TENSOR1->size[TENSOR1##_i]; \
-\
-  TENSOR2##_n = (TENSOR2->nDimension ? 1 : 0); \
-  for(TENSOR2##_i = 0; TENSOR2##_i < TENSOR2->nDimension; TENSOR2##_i++) \
-    TENSOR2##_n *= TENSOR2->size[TENSOR2##_i]; \
-\
-  TENSOR3##_n = (TENSOR3->nDimension ? 1 : 0); \
-  for(TENSOR3##_i = 0; TENSOR3##_i < TENSOR3->nDimension; TENSOR3##_i++) \
-    TENSOR3##_n *= TENSOR3->size[TENSOR3##_i]; \
-\
-  if(TENSOR1##_n != TENSOR2##_n || TENSOR1##_n != TENSOR3##_n) /* should we do the check in the function instead? i think so */ \
-    THError("inconsistent tensor size"); \
-\
-  if(TENSOR1->nDimension == 0) \
-    TH_TENSOR_APPLY_hasFinished = 1; \
-  else \
-  { \
-    TENSOR1##_data = TENSOR1->storage->data+TENSOR1->storageOffset; \
-    TENSOR1##_size = 1; \
-    TENSOR1##_stride = 1; \
-    for(TENSOR1##_i = TENSOR1->nDimension-1; TENSOR1##_i >= 0; TENSOR1##_i--) { \
-      if(TENSOR1->size[TENSOR1##_i] != 1) { \
-        if(TENSOR1->stride[TENSOR1##_i] == TENSOR1##_size) \
-          TENSOR1##_size *= TENSOR1->size[TENSOR1##_i]; \
-        else{ \
-          TH_TENSOR1_contiguous = 0; \
-          break; \
-        } \
-      } \
-    } \
-    if (!TH_TENSOR1_contiguous) { \
-      TENSOR1##_dim = 1; \
-      for(TENSOR1##_i = TENSOR1->nDimension-2; TENSOR1##_i >= 0; TENSOR1##_i--) \
-      { \
-        if(TENSOR1->stride[TENSOR1##_i] != TENSOR1->stride[TENSOR1##_i+1] * TENSOR1->size[TENSOR1##_i+1]) \
-          TENSOR1##_dim++; \
-      } \
-      TENSOR1##_counter = (long*)THAlloc(sizeof(long)*(3*TENSOR1##_dim)); \
-      TENSOR1##_sizes = TENSOR1##_counter + TENSOR1##_dim; \
-      TENSOR1##_strides = TENSOR1##_counter + 2*TENSOR1##_dim; \
-      TH_TENSOR_dim_index = TENSOR1##_dim-1; \
-      TENSOR1##_sizes[TH_TENSOR_dim_index] = TENSOR1->size[TENSOR1->nDimension-1]; \
-      TENSOR1##_strides[TH_TENSOR_dim_index] = TENSOR1->stride[TENSOR1->nDimension-1]; \
-      for(TENSOR1##_i = TENSOR1##_dim-1; TENSOR1##_i >= 0; --TENSOR1##_i) { \
-        TENSOR1##_counter[TENSOR1##_i] = 0; \
-      } \
-      for(TENSOR1##_i = TENSOR1->nDimension-2; TENSOR1##_i >= 0; --TENSOR1##_i) { \
-        if (TENSOR1->stride[TENSOR1##_i] == TENSOR1->stride[TENSOR1##_i+1] * TENSOR1->size[TENSOR1##_i+1]) { \
-          TENSOR1##_sizes[TH_TENSOR_dim_index] = TENSOR1->size[TENSOR1##_i] * TENSOR1##_sizes[TH_TENSOR_dim_index]; \
-        } else { \
-          --TH_TENSOR_dim_index; \
-          TENSOR1##_sizes[TH_TENSOR_dim_index] = TENSOR1->size[TENSOR1##_i]; \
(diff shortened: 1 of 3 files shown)


soumith:
it looks like we're lucky that TH_TENSOR_DIM_APPLY2 etc. dont have OpenMP optimizations yet. in the longer term, we need tree reductions.

lib/TH/generic/THTensorMath.c
```diff
@@ -1124,32 +1129,70 @@ void THTensor_(max)(THTensor *values_, THLongTensor *indices_, THTensor *t, int
...
+    } else {
+      THTensor_(fill)(values_, THTensor_(get1d)(t, 0));
+    }
+    THLongTensor_zero(indices_);
+
+    if(t->size[dimension] == 1) {
+      return;
+    }
+
+    THTensor *tempValues_ = THTensor_(newWithTensor)(values_);
```
fmassa: Don't you need to free this tensor? Same for the indices.

lib/TH/generic/THTensorMath.c
```diff
@@ -1124,32 +1129,70 @@ void THTensor_(max)(THTensor *values_, THLongTensor *indices_, THTensor *t, int
...
                          }
-                       }
-                       *indices__data = theIndex;
-                       *values__data = theMax;);
+                         *indices__data = theIndex;
+                         *values__data = theMax;);
+  } else {
+    if (THTensor_(nDimension)(t) > 1) {
+      THTensor *t0 = THTensor_(newSelect)(t, dimension, 0);
+      THTensor_(copy)(values_, t0);
```
apaszke: An optimization would be to fill it with the smallest possible value for each type. You're going to compare the values with the first slice anyway, and that would save some bandwidth.

lib/TH/generic/THTensorMath.c
```diff
@@ -1124,32 +1129,70 @@ void THTensor_(max)(THTensor *values_, THLongTensor *indices_, THTensor *t, int
...
+    // tempIndices_.expand_as(t)
+    tempIndices_->size[dimension] = t->size[dimension];
+    tempIndices_->stride[dimension] = 0;
+
+    // this shows the flexibility of using TH_TENSOR_APPLY in place of
+    // TH_TENSOR_DIM_APPLY. Unfortunately, I'm leveraging the fact that since
+    // tempIndices has stride 0 and size >1 in dimension, there will definitely
+    // be an instantiated counter dimension there; this might not be true after
+    // some new optimizations to TH_TENSOR_APPLY3, so we'll need a different
+    // set of macros.
```
apaszke: The problem arises from the fact that apply macros will merge the counters, right?

lib/TH/generic/THTensorMath.c
```diff
@@ -1192,12 +1277,24 @@ void THTensor_(sum)(THTensor *r_, THTensor *t, int dimension)
...
+                           sum += t_data[i*t_stride];
+                         *r__data = (real)sum;);
+  } else {
+    THTensor_(zero)(r_);
+    THTensor *temp_ = THTensor_(newWithTensor)(r_);
+    // r_.expand_as(t)
+    temp_->size[dimension] = t->size[dimension];
+    temp_->stride[dimension] = 0;
+
+    THTensor_(cadd)(temp_, temp_, 1, t);
```
apaszke: Are you sure that is correct? I'd be very careful with mutating operations on expanded tensors. Parallelizing `cadd` will lead to race conditions here.

lib/TH/generic/THTensorMath.c
adamlerer: Yep, thanks. 👍

lib/TH/generic/THTensorMath.c
adamlerer: That seems like a pretty annoying optimization just to save 1/size[dim] bandwidth. If you really wanted to, the clean way would be to initialize the result with t[0] and then do the APPLY over t[1:].

lib/TH/generic/THTensorMath.c
adamlerer: Yes, that's correct. It's unclear what the contract is with TH_TENSOR_APPLY... we may need one set of macros that exposes the counter and one that is allowed to optimize it away.

lib/TH/generic/THTensorMath.c
adamlerer: Good question. I know it's currently correct, but the question is whether it relies on the implementation. It comes down to whether `cadd` behavior is defined for destinations with stride-0 dimensions. Currently it does (at least for the CPU), because it only parallelizes the contiguous case.

Perhaps it would be cleaner to use `TH_APPLY` here directly instead of `cadd`, because we can require that `TH_APPLY` handles stride-0 destination properly. If we want to parallelize it in the future we'll need to split it into two versions.

apaszke: Even if it's correct at the moment, I'm afraid that we'll change `cadd` in the future, and have other seemingly unrelated things break. Not sure about the remaining part, I have to look into the apply macros to see what they do in detail.

adamlerer: Sure, I'll change it to use TH_APPLY directly.

If someone changes `TH_APPLY` to no longer handle stride-0 dimensions properly, the tests will fail and they'll have to fix it :P

adamlerer:
@soumith re OpenMP: as I said to @apaszke this comes down to the specification of `TH_TENSOR_APPLY` (not TH_TENSOR_DIM_APPLY) w.r.t. stride-0 dimensions. Currently it handles them correctly. If someone changes it in a way that breaks that, they'll need to maintain a version that handles stride-0 correctly. Note that there's already existing code that sort of depends on handling this case correctly, e.g.

```
t = torch.Tensor(10, 10, 10)
u = torch.Tensor(1, 10, 10).expand_as(t).zero_()
u.add_(t)
```

so I'm not sure if it is now safe to break that (considering the specification was ill-defined).

lib/TH/generic/THTensorMath.c
```diff
@@ -2398,35 +2401,11 @@ void THTensor_(lerp)(THTensor *r_, THTensor *a, THTensor *b, real weight)
...
-    THTensor *temp_ = THTensor_(newWithTensor)(r_);
-    // r_.expand_as(t)
-    temp_->size[dimension] = t->size[dimension];
-    temp_->stride[dimension] = 0;
-
-    THTensor_(cadd)(temp_, temp_, 1, t);
-    THTensor_(free)(temp_);
-    THTensor_(div)(r_, r_, t->size[dimension]);
-  }
+  THTensor_(sum)(r_, t, dimension);
```
fmassa: Just wondering, is there a performance penalty in the contiguous case with this change?

lib/TH/generic/THTensorMath.c
apaszke: If it's going to be a loud fail then that sounds good to me, but I'd rather save time on debugging race conditions in the future 😄

lib/TH/generic/THTensorMath.c
```diff
@@ -1124,32 +1129,73 @@ void THTensor_(max)(THTensor *values_, THLongTensor *indices_, THTensor *t, int
...
+                         *values__data = theMax;);
+  } else {
+    if (THTensor_(nDimension)(t) > 1) {
+      THTensor *t0 = THTensor_(newSelect)(t, dimension, 0);
+      THTensor_(copy)(values_, t0);
+      THTensor_(free)(t0);
+    } else {
+      THTensor_(fill)(values_, THTensor_(get1d)(t, 0));
+    }
+    THLongTensor_zero(indices_);
```
apaszke: You need to initialize them with `TH_INDEX_BASE` (applies to other functions too).

apaszke: Also, it would be good to extend the tests to cover this.

lib/TH/generic/THTensorMath.c
```diff
@@ -1124,32 +1129,73 @@ void THTensor_(max)(THTensor *values_, THLongTensor *indices_, THTensor *t, int
...
+    tempIndices_->size[dimension] = t->size[dimension];
+    tempIndices_->stride[dimension] = 0;
+
+    // this shows the flexibility of using TH_TENSOR_APPLY in place of
+    // TH_TENSOR_DIM_APPLY. Unfortunately, I'm leveraging the fact that since
+    // tempIndices has stride 0 and size >1 in dimension, there will definitely
+    // be an instantiated counter dimension there; this might not be true after
+    // some new optimizations to TH_TENSOR_APPLY3, so we'll need a different
+    // set of macros.
+    TH_TENSOR_APPLY3(real, t, real, tempValues_, long, tempIndices_,
```
Jokeren: Hi, @adamlerer , could you explain why `t` and `tempValues_` has the same size? I notice you have comment `tempIndices_.expand_as(t)`.

lib/TH/generic/THTensorMath.c
```diff
@@ -1124,32 +1129,73 @@ void THTensor_(max)(THTensor *values_, THLongTensor *indices_, THTensor *t, int
...
+                     if(!(*t_data <= *tempValues__data) && !th_isnan(*tempValues__data)) {
+                       *tempValues__data = *t_data;
+                       *tempIndices__data = tempIndices__counter[dimension];
```
Jokeren: The problem is that `dimension` works only in the raw counter, not the compressed one. So I am afraid that you must use another macro for it, or extra dimension calculations are needed.

adamlerer: Yes I know. So what should we do here? I can either remove the `min` and `max` optimizations and save them for a later pass with new macros, or I can copy-paste the old `TH_TENSOR_APPLYX` macros just for min and max in this PR. Thoughts? @apaszke

Jokeren: I would vote for using the old `TH_TENSOR_APPLYX`.

apaszke:
I think we could leave them as TH_TENSOR_APPLY_COUNTER or sth like that. They're all unacceptably slow, so let's keep the fix.

lib/TH/generic/THTensorMath.c
apaszke: I think we could leave them as TH_TENSOR_APPLY_COUNTER or sth like that. They're all unacceptably slow, so let's keep the fix.

lib/TH/generic/THTensorMath.c
adamlerer: This is not necessary, because cwrap adds 1 to the indexes in Lua. I'll add the test though.

lib/TH/generic/THTensorMath.c
adamlerer: Will do.

lib/TH/generic/THTensorMath.c
apaszke: Right. I forgot about that.

adamlerer:
Alright, I rebased this on top of https://github.com/torch/torch7/pull/852, refactored TH_TENSOR_APPLY to remove all the copypasta and added a variant that can handle these reduction cases.

You'll have to wait until https://github.com/torch/torch7/pull/852 is landed to review this.

lib/TH/THTensorApply.h
```diff
@@ -595,8 +131,90 @@
...
+      CODE \
+    } \
+    __TH_TENSOR_APPLYX_UPDATE_COUNTERS(TENSOR, 1) \
   } \
   THFree(TENSOR##_counter); \
 }
 
+#define TH_TENSOR_APPLY(TYPE, TENSOR, CODE) \
+  TH_TENSOR_APPLY_D(TYPE, TENSOR, -1, CODE)
+
```
adamlerer: To make reviewing easier, note that most of the changes to this file are just a refactoring to remove copy-pasted macros. The only functional changes, which was necessary for the reduction operations, can be found by searching for uses of the new `DIM` argument to `TH_TENSOR_APPLYX_D`

soumith:
thanks Adam!

PR #944, opened 2017-02-23
Adding AVX Optimizations with Runtime Dispatch

soumith:
Rebased version of #852


soumith:
this looks good now.
The main affected functions for SIMD are `cadd, add, cmul, mul, cdiv, div, copy`.

The testing story around this PR is a little weak, especially around SIMD.
I'll work on this PR today to introduce a few environment variables that control enabling/disabling  SIMD features.
I'm thinking of introducing environment variables:
`TH_NO_AVX2`, `TH_NO_AVX`, `TH_NO_SSE` to control the runtime dispatch to skip certain features.
That way, we can easily test different SIMD code paths. For example: `TH_NO_AVX2=1 th -ltorch -e"torch.test()"` will test `AVX`, and similarly setting `TH_NO_AVX2=1 TH_NO_AVX=1` will test the `SSE` codepath.
I see little value in making these flags to be enabled at a lua level, and the code changes needed to enable them at the lua level is relatively large (dispatch pointers are initialized at the beginning of TH's init, so these have to be changed too).

Let me know of any objections. cc: @Jokeren @adamlerer @pavanky

Jokeren:
@soumith , this is fine.

I am thinking of another problem. The VSX vector codes are implemented with the old interfaces. So after merging the `simd-opt` branch, they will not take effect anymore.

soumith:
cc: @amrobbins about the VSX breakage

soumith:
@Jokeren there are still bugs in this PR, i'm seeing stack corruption. 
I've isolated the bugs to not be in the SIMD optimizations themselves. 
I suspect bugs in the optimizations of APPLY kernels.
To reproduce, you can install this branch and then run tests for the `nn` package:
```
th -lnn -e "nn.test()"
```

Gives me double free and malloc failure errors:
```
luajit: malloc.c:3695: _int_malloc: Assertion `(unsigned long) (size) >= (unsigned long) (nb)' failed.
```

I'm running a bisect, i'll isolate it shortly.

soumith:
i've isolated the bug to "Fill bug fix - 326cbf1", but i'm not super confident about my bisect (it's a complex PR)

adamlerer:
The bug is that https://github.com/torch/torch7/commit/326cbf1def912eda20f6d1560f5556f79ea6b9b6 moves `r__data` forward by `size`, but `TH_APPLY` only moves it back by `r__i` (which is 0 since you break in the first iteration). Here's a patch that fixes the memory corruption bug.

```
diff --git a/lib/TH/generic/THTensorMath.c b/lib/TH/generic/THTensorMath.c
index f698e25..2b8c9e1 100644
--- a/lib/TH/generic/THTensorMath.c
+++ b/lib/TH/generic/THTensorMath.c
@@ -97,8 +97,9 @@ void THTensor_(fill)(THTensor *r_, real value)
     TH_TENSOR_APPLY(real, r_,
       if (r__stride == 1) {
         THVector_(fill)(r__data, value, r__size);
-       r__data += r__stride * r__size;
-       break;
+        r__i = r__size;
+        r__data += r__stride * r__size;
+        break;
       } else {
         *r__data = value;
       }
```

Unfortunately, once you do that, there are other failures in nn. These are caused by the fact tht @Jokeren has changed the semantics of `THVector_(add)` (while leaving the function signature the same), which is used by `nn` in a couple places. Here's an nn patch that fixes them:

```
diff --git a/lib/THNN/generic/VolumetricConvolutionMM.c b/lib/THNN/generic/VolumetricConvolutionMM.c
index 4085e2b..caaaa55 100644
--- a/lib/THNN/generic/VolumetricConvolutionMM.c
+++ b/lib/THNN/generic/VolumetricConvolutionMM.c
@@ -152,7 +152,8 @@ static void THNN_(unfolded_acc_vol)(
                   }
                   else
                   {
-                    THVector_(add)(dst+it*inputHeight*inputWidth+iy*inputWidth+ix, src+t*outputHeight*outputWidth+y*outputWidth+x, 1, 1);
+                    real *dst_slice = dst+it*inputHeight*inputWidth+iy*inputWidth+ix;
+                    THVector_(cadd)(dst_slice, dst_slice, src+t*outputHeight*outputWidth+y*outputWidth+x, 1, 1);
                   }
                 }
               }
@@ -169,7 +170,8 @@ static void THNN_(unfolded_acc_vol)(
                 for(x = 0; x < outputWidth; x++)
                 {
                   ix = x*dW + kw;
-                  THVector_(add)(dst+it*inputHeight*inputWidth+iy*inputWidth+ix, src+t*outputHeight*outputWidth+y*outputWidth+x, 1, 1);
+                  real *dst_slice = dst+it*inputHeight*inputWidth+iy*inputWidth+ix;
+                  THVector_(cadd)(dst_slice, dst_slice, src+t*outputHeight*outputWidth+y*outputWidth+x, 1, 1);
                 }
               }
             }
diff --git a/lib/THNN/generic/unfold.c b/lib/THNN/generic/unfold.c
index 25146c0..e718320 100644
--- a/lib/THNN/generic/unfold.c
+++ b/lib/THNN/generic/unfold.c
@@ -52,14 +52,17 @@ void THNN_(unfolded_acc)(
                  ix = (long long)(0 - padW + kw);
                  lpad = fmaxf(0,(int)(padW-kw));
                  rpad = fmaxf(0,(int)(padW-(kW-kw-1)));
-                 THVector_(add)(dst+(size_t)(iy*inputWidth+ix+lpad), src+(size_t)(y*outputWidth+lpad), 1, outputWidth - lpad - rpad); /* note: THVector_add could handle 1 value better */
+                 real *dst_slice = dst+(size_t)(iy*inputWidth+ix+lpad);
+                 THVector_(cadd)(dst_slice, dst_slice, src+(size_t)(y*outputWidth+lpad), 1, outputWidth - lpad - rpad); /* note: THVector_add could handle 1 value better */
               }
               else{
                 for (x=0; x<outputWidth; x++){
                    ix = (long long)(x*dW - padW + kw);
                    if (ix < 0 || ix >= inputWidth){
-                   }else
-                     THVector_(add)(dst+(size_t)(iy*inputWidth+ix), src+(size_t)(y*outputWidth+x), 1, 1);
+                   }else{
+                     real *dst_slice = dst+(size_t)(iy*inputWidth+ix);
+                     THVector_(cadd)(dst_slice, dst_slice, src+(size_t)(y*outputWidth+x), 1, 1);
+                   }
                 }
               }
             }
@@ -68,11 +71,14 @@ void THNN_(unfolded_acc)(
           for(y = 0; y < outputHeight; y++) {
             iy = (long long)(y*dH + kh);
             ix = (long long)(0 + kw);
-            if (dW == 1 )
-               THVector_(add)(dst+(size_t)(iy*inputWidth+ix), src+(size_t)(y*outputWidth), 1, outputWidth); /* note: THVector_add could handle 1 value better */
-            else{
-              for(x = 0; x < outputWidth; x++)
-                THVector_(add)(dst+(size_t)(iy*inputWidth+ix+x*dW), src+(size_t)(y*outputWidth+x), 1, 1);
+            if (dW == 1 ) {
+               real *dst_slice = dst+(size_t)(iy*inputWidth+ix);
+               THVector_(cadd)(dst_slice, dst_slice, src+(size_t)(y*outputWidth), 1, outputWidth); /* note: THVector_add could handle 1 value better */
+            }else{
+              for(x = 0; x < outputWidth; x++) {
+                real *dst_slice = dst+(size_t)(iy*inputWidth+ix+x*dW);
+                THVector_(cadd)(dst_slice, dst_slice, src+(size_t)(y*outputWidth+x), 1, 1);
+              }
             }
           }
         }
```

This patch isn't great though because it requires `torch` and `nn` to be updated in parallel, and will produce incorrect results silently if `torch` and `nn` are out of sync. Instead, @Jokeren needs to rename `THVector_(add)` to something like `THVector_(addc)` (for addConstant).

Jokeren:
@adamlerer , @soumith , Thank you for comments!

I am curious about the bisect result because in the final commit I do not use that logic (https://github.com/torch/torch7/commit/326cbf1def912eda20f6d1560f5556f79ea6b9b6) for `fill` operation. I will take examine my code carefully. So sorry that I didn't consider testing `THNN`.

Jokeren:
@adamlerer , it's very kind of you to do the debugging! But I support using the patch you proposed instead of renaming `add` API.

I think `add` stands for adding a vector with a constant just as `mul` and `div`. And `cadd`, `cmul` and `cdiv` are functions that handle two vectors (we have a constant in `cadd` to support fma). So in this way, I think the meaning of old `add` API is not correct. Why will `torch` and `nn` be out of sync? If we change both APIs now and reference the new version, it will be correct.

adamlerer:
@Jokeren you're right that your new naming is better, but the problem is that you are changing the semantics of an existing function that downstream packages may be using, which will cause them to silently produce incorrect results.

Even if we update `torch7` and `nn` at the same time, they are in separate repos someone may sync only one of the repos and build it, which will cause their code to produce incorrect results. Similarly, some other downstream code (not `nn`) may be using `THVector_(add)` and will start silently producing incorrect results.

soumith:
i'll rename `THVector_(add)` to `THVector_(addscalar)`and wrap up this work on monday.
Thanks a lot @adamlerer for debugging and giving the patches to fix the bugs.

PR #948, opened 2017-03-01
New helper macro + fixes for old gcc 

pavanky:
- `ADD_TORCH_LIBRARY` makes it easier to link `libTH` from various other packages.
- Static version additionally checks for environment variable.


cmake/TorchPackage.cmake
```diff
@@ -1,5 +1,26 @@
...
+MACRO(ADD_TORCH_LIBRARY package type src)
+IF ("${type}" STREQUAL "STATIC")
+  if ("${src}" MATCHES "cu$" OR "${src}" MATCHES "cu;")
+    CUDA_ADD_LIBRARY(${package} STATIC ${src})
+  else()
+    ADD_LIBRARY(${package} STATIC ${src})
+  endif()
+ELSE()
+  IF (BUILD_STATIC OR "$ENV{TH_ONLY_STATIC}" STREQUAL "YES")
+    ADD_LIBRARY(${package} ${type} "")
```
soumith: isn't this missing a STATIC

pavanky: Hmm it looks like I messed something up during rebasing. Let me double check this.

pavanky: Ok, it isn't missing `STATIC`, it is missing source files. It was a hacky way to disable dynamic builds when only static versions were required. 

I am going to remove this logic and force push.

pavanky:
@soumith Removed that code segment. The current behavior is similar to the current version in master where both static and dynamic libs are built when static builds are requested. If not only dynamic libs are built.

soumith:
if you rebase, i'll merge this.

pavanky:
@soumith sorry about not getting back to this, I'll clean this up tomorrow.

PR #949, opened 2017-03-01
custom convolution kernels for 3x3

pavanky:


soumith:
just fyi it's prob most efficient at this point on CPU to either use NNPack or MKLDNN:
https://github.com/Maratyszcza/NNPACK
https://github.com/szagoruyko/nnpack.torch

pavanky:
@soumith Thanks for the links. This PR is not super important from our end :) Sent it in because it was the last commit on our fork that diverged from upstream torch.

Jokeren:
@soumith , @pavanky , Just want to point out that MKL kernels depend on special data layout, so that we need to allocate a large buffer before convolution and also transform the output format after convolution. I am not saying it will affect the whole performance. But this is not convenient to combine MKL into torchnn, as MKL only define some nn functions that utilize the special data layout without transformation. Besides, what about other processors that do not support MKL? I think torch should maintain its own convolution implementations.

apaszke:
Wouldn't it be better to just integrate nnpack then?

lib/TH/cmake/FindMKL.cmake
```diff
@@ -161,8 +165,8 @@ FOREACH(mklrtl ${mklrtls} "")
       FOREACH(mklthread ${mklthreads})
         IF (NOT MKL_LIBRARIES AND NOT INTEL_MKL_SEQUENTIAL)
           CHECK_ALL_LIBRARIES(MKL_LIBRARIES cblas_sgemm
-            "mkl_${mkliface}${mkl64};${mklthread};mkl_core;${mklrtl};pthread;${mkl_m}" "")
-        ENDIF (NOT MKL_LIBRARIES AND NOT INTEL_MKL_SEQUENTIAL)          
+            "mkl_${mkliface}${mkl64};${mklthread};mkl_core;${mklrtl};pthread;${mkl_m};dl" "-Wl,--start-group")
```
soumith: dl is not a thing on OSX.

pavanky: Hmm, I wonder why this wasn't an issue internally. I'll look into it thanks.

soumith: maybe on OSX it's being linked against Accelerate, instead of MKL

pavanky:
Closing this as this does not seem relevant anymore.

PR #952, opened 2017-03-02
C99 cleanup broke MSVC

shivak:
The recent refactoring of THTensorMath.c uses C99's _Pragma. This isn't supported by MSVC, which uses __pragma. I confirm this builds on MSVC.

### DIFF
--- a/lib/TH/generic/THTensorMath.c
+++ b/lib/TH/generic/THTensorMath.c
@@ -1,7 +1,7 @@
 #ifndef TH_GENERIC_FILE
 #define TH_GENERIC_FILE "generic/THTensorMath.c"
 #else
-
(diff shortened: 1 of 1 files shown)


lib/TH/generic/THTensorMath.c
```diff
@@ -9,10 +9,17 @@
 #define TH_OMP_OVERHEAD_THRESHOLD 100000
 
 #ifdef _OPENMP
+
+#ifndef _WIN32
+#define PRAGMA(P) _Pragma("P")
```
apaszke: Isn't this going to wrap P in quotes twice?

soumith:
thanks!

PR #959, opened 2017-03-03
Fix dimension check for cat

pavanky:
- Fixes  default dimension check from -2 to use `TH_INDEX_BASE`. I think master probably fails on PyTorch.(haven't checked). 
- Also added tests for cat when no dimension is specified (Similar to what is happening in https://github.com/torch/cutorch/pull/719)
- As done with cutorch, changed the variable name to make life easier when debugging this function.


soumith:
thanks!
pytorch does not use the no-dimension case (it computes it's own dims and passes them as it supports negative dims), and hence does not fail.

PR #981, opened 2017-03-22
make it compile on Windows + use ilp64 MKL

elikosan:
- fix vector/avx Windows compile (missing _dllexport + __AVX__ not defined on Windows)
- fix BLAS functions to operate on 64 bits __int64 (equiv. to long on linux) when compiling 64bit platform
- fix FindSSE.cmake so it does not crash on Windows (unitialized variable)


lib/TH/vector/AVX.h
```diff
@@ -2,22 +2,23 @@
 #define TH_AVX_H
 
 #include <stddef.h>
+#include "THGeneral.h"
```
BTNC: AVX.h is included after THGeneral.h in file ./lib/TH/THVector.c. So there is no need to include THGeneral.h here.

lib/TH/vector/AVX.h
```diff
@@ -2,22 +2,23 @@
...
-void THDoubleVector_adds_AVX(double *y, const double *x, const double c, const ptrdiff_t n);
-void THFloatVector_copy_AVX(float *y, const float *x, const ptrdiff_t n);
-void THFloatVector_fill_AVX(float *x, const float c, const ptrdiff_t n);
-void THFloatVector_cdiv_AVX(float *z, const float *x, const float *y, const ptrdiff_t n);
-void THFloatVector_divs_AVX(float *y, const float *x, const float c, const ptrdiff_t n);
-void THFloatVector_cmul_AVX(float *z, const float *x, const float *y, const ptrdiff_t n);
-void THFloatVector_muls_AVX(float *y, const float *x, const float c, const ptrdiff_t n);
-void THFloatVector_cadd_AVX(float *z, const float *x, const float *y, const float c, const ptrdiff_t n);
-void THFloatVector_adds_AVX(float *y, const float *x, const float c, const ptrdiff_t n);
+TH_API void THDoubleVector_copy_AVX(double *y, const double *x, const ptrdiff_t n);
```
BTNC: These apis are not supposed to be exported, but are used for specific implementations for different instructions internally. Please refer to lib/TH/generic/THVectorDispatch.c

lib/TH/vector/AVX.c
```diff
@@ -1,4 +1,3 @@
-#if defined(__AVX__)
```
BTNC: Though it is harmless to remove the macro check for msvc, I think it can be kept because I am not sure the behavior of other platforms. Besides, this macro is the  source I found that caused the compilation broken. #991

lib/TH/vector/AVX2.c
```diff
@@ -1,4 +1,3 @@
-#if defined(__AVX2__)
```
BTNC: same comment as AVX.c

lib/TH/vector/AVX2.h
```diff
@@ -2,8 +2,9 @@
 #define TH_AVX2_H
 
 #include <stddef.h>
+#include "THGeneral.h"
```
BTNC: same comment as AVX.h

elikosan:
I agree with BTNC reviews. Pull request #991 fixes the issues that i was having, thus i reverted the changes to avx.h,avx.c,avx2.h,avx2.c.

soumith:
thanks Eric!

lib/TH/cmake/FindBLAS.cmake
```diff
@@ -242,32 +242,44 @@ endif()
...
 #include <stdlib.h>
 #include <stdio.h>
 float x[4] = { 1, 2, 3, 4 };
 float y[4] = { .1, .01, .001, .0001 };
-int four = 4;
-int one = 1;
+#ifdef WIN32
+  typedef __int64 BLINT;
+#else
+  typedef long BLINT;
```
apaszke: Are you sure this typedef is correct? It used to give `sdot_` an `int*` on non WIN32 platforms, now it uses `long*`, but `sizeof(long) != sizeof(int)` on most x86_64 systems. The same problem seems to be present in other parts of this PR too.

apaszke: @elikosan

PR #991, opened 2017-04-02
add /arch:AVX /arch:AVX2 explicitly for msvc so it compiles on windows

BTNC:
Hi,

The AVX/AVX2 test for msvc can pass without /arch:AVX or /arch:AVX2 compiler options. And FindSSE.cmake will test with no flags at first, so the empty flag is returned. However the `__AVX__` and `__AVX2__` macro will only be defined automatically when /arch:AVX and /arch:AVX2 options are specified.

I am not sure about gcc or clang's behavior, so I did not change FindSSE.cmake directly. Instead I added those must have options directly in SET_SOURCE_FILES_PROPERTIES.

Thanks,

### DIFF
--- a/lib/TH/CMakeLists.txt
+++ b/lib/TH/CMakeLists.txt
@@ -25,8 +25,8 @@ ENDIF()
 ######################################################################
 
 IF(MSVC)
-  # MSVC now supports C99 since VS2013/VS2015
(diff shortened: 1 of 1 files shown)


soumith:
thanks!

PR #1000, opened 2017-04-10
Update CmdLine.lua

willfrey:
Very small cosmetic change to make positional arguments look better.

Previously, positional arguments would look like:
```
$ th my_script.lua --help
Usage: [options] <arg1><arg2>
-opt1 blah blah [default]
```
and now it should look like
```
$ th my_script.lua --help
Usage: [options] <arg1> <arg2>
-opt1 blah blah [default]
```
Very minorly cosmetic.


soumith:
thanks will!

PR #1011, opened 2017-04-21
TH implementation of Tensor expansion

killeent:
@soumith not sure if you wanted me to migrate the Lua-level expand to this too, or if this should just be wrapped from PyTorch.

### DIFF
--- a/lib/TH/THStorage.c
+++ b/lib/TH/THStorage.c
@@ -43,7 +43,8 @@ TH_API THLongStorage *THLongStorage_newInferSize(THLongStorage *size, ptrdiff_t 
   ptrdiff_t i;
   for (i = 0; i < size->size; i++) {
     if (size->data[i] == -1) {
-      THArgCheck(dim_infer == -1, 1, "only one dimension can be inferred");
+      THArgCheck(dim_infer == -1, 1, "asking to infer dimension %ld, but already inferring dimension %ld. \
+          You can only infer one dimension.", i, dim_infer);
       dim_infer = i;
     } else {
       total_size *= size->data[i];
--- a/lib/TH/generic/THTensor.c
+++ b/lib/TH/generic/THTensor.c
@@ -285,6 +285,58 @@ void THTensor_(resize5d)(THTensor *self, long size0, long size1, long size2, lon
   THTensor_(resizeNd)(self, 5, size, NULL);
 }
 
+void THTensor_(expandNd)(THTensor *tensor, long *sizes, ptrdiff_t ndim) {
+  THArgCheck(tensor->nDimension > 0, 0, "can't expand an empty tensor");
+  long numUnsqueezed = ndim - tensor->nDimension;
+
+  long *expandedSizes = THAlloc(sizeof(long)*ndim);
+  long *expandedStrides = THAlloc(sizeof(long)*ndim);
+
+  for (long i = 0; i < numUnsqueezed; ++i) {
+    expandedSizes[i] = 1;
+    expandedStrides[i] = 0;
+  }
+
+  for (long i = numUnsqueezed; i < ndim; ++i) {
+    expandedSizes[i] = THTensor_(size)(tensor, i);
+    expandedStrides[i] = THTensor_(stride)(tensor, i);
+  }
+
+  for (long i = numUnsqueezed - 1; i > -1; --i) {
+    expandedStrides[i] = expandedSizes[i+1] * expandedStrides[i+1];
+  }
+
+  // create a new geometry for the tensor
+  for (long i = 0; i < ndim; ++i) {
+    long size = expandedSizes[i];
+    long targetSize = sizes[i];
+    if (size == 1) {
+      if (targetSize != 1) {
+        expandedSizes[i] = targetSize;
+        expandedStrides[i] = 0;
+      }
+    } else if (size != targetSize) {
+      THError("incorrect size: only supporting singleton expansion (size=1)");
+    }
+  }
+
+  THTensor_(setStorageNd)(tensor, tensor->storage, tensor->storageOffset, ndim, expandedSizes, expandedStrides);
+  THFree(expandedSizes);
+  THFree(expandedStrides);
+}
+
+void THTensor_(expand)(THTensor *tensor, THLongStorage *sizes) {
+  THArgCheck(sizes->size >= tensor->nDimension, 1, "the number of sizes provided \
+      must be greater or equal to the number of dimensions in the tensor");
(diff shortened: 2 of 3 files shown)


lib/TH/generic/THTensor.c
```diff
@@ -285,6 +285,58 @@ void THTensor_(resize5d)(THTensor *self, long size0, long size1, long size2, lon
...
+  for (long i = 0; i < ndim; ++i) {
+    long size = expandedSizes[i];
+    long targetSize = sizes[i];
+    if (size == 1) {
+      if (targetSize != 1) {
+        expandedSizes[i] = targetSize;
+        expandedStrides[i] = 0;
+      }
+    } else if (size != targetSize) {
+      THError("incorrect size: only supporting singleton expansion (size=1)");
```
gchanan: free expandedSizes/Strides here?

killeent: yes.

Do we have examples of where Torch does not just exit out on THError? The default behavior calls `exit(-1)`.

gchanan: Good question; I was just basing this on e.g.: https://github.com/torch/torch7/blob/5f9d30769789bc36ec628e3eec2f2fb85a1e6b50/generic/Tensor.c#L121-L124

soumith: the error handler is re-registed in both lua and python at runtime. This will give a standard runtime error.

lib/TH/generic/THTensor.c
```diff
@@ -285,6 +285,60 @@ void THTensor_(resize5d)(THTensor *self, long size0, long size1, long size2, lon
...
+    long targetSize = sizes[i];
+    if (size == 1) {
+      if (targetSize != 1) {
+        expandedSizes[i] = targetSize;
+        expandedStrides[i] = 0;
+      }
+    } else if (size != targetSize) {
+      THFree(expandedSizes);
+      THFree(expandedStrides);
+      THError("incorrect size: only supporting singleton expansion (size=1)");
```
soumith: can you make the error messages more informative. The view one doesn't do a great job either.

lib/TH/generic/THTensor.c
```diff
@@ -285,6 +285,61 @@ void THTensor_(resize5d)(THTensor *self, long size0, long size1, long size2, lon
...
+    long targetSize = sizes[i];
+    if (size == 1) {
+      if (targetSize != 1) {
+        expandedSizes[i] = targetSize;
+        expandedStrides[i] = 0;
+      }
+    } else if (size != targetSize) {
+      THFree(expandedSizes);
+      THFree(expandedStrides);
+      THError("The specified size (%d) at non-singleton dimension %ld does not match \
```
gchanan: This is a difficult error message to write, because it's called from both expand (where sizes are specified) and expandAs (where a tensor is specified), so using "specified" is tricky.  How about something like:

"The expanded size of the tensor (%d) must match the existing size (%d) at non-singleton dimension %ld."

That might take awhile to parse though.  Maybe something simpler like:
"Cannot expand dimension %ld.  The existing size (%d) of that dimension must either be 1 or equal to the expanded size (%d)."

killeent:
@gchanan I went with the first one, in any event they will probably look to the docs for further clarification.

gchanan:
thanks!  BTW -- is there any reason expandNd is not declared in generic/THTensor.h?

killeent:
Its helper code, while expand/expandAs are top-level API calls we will wrap. I'm not sure (however) with how the code is generated, whether or not the header will be needed for it to work.

lib/TH/THStorage.c
```diff
@@ -43,7 +43,8 @@ TH_API THLongStorage *THLongStorage_newInferSize(THLongStorage *size, ptrdiff_t
   ptrdiff_t i;
   for (i = 0; i < size->size; i++) {
     if (size->data[i] == -1) {
-      THArgCheck(dim_infer == -1, 1, "only one dimension can be inferred");
+      THArgCheck(dim_infer == -1, 1, "asking to infer dimension %ld, but already inferring dimension %ld. \
```
apaszke: How about `asking to infer two dimensions (%ld and %ld), but can only infer one`?

lib/TH/generic/THTensor.c
```diff
@@ -285,6 +285,61 @@ void THTensor_(resize5d)(THTensor *self, long size0, long size1, long size2, lon
...
+void THTensor_(expandNd)(THTensor *tensor, long *sizes, ptrdiff_t ndim) {
+  THArgCheck(tensor->nDimension > 0, 0, "can't expand an empty tensor");
+  long numUnsqueezed = ndim - tensor->nDimension;
+
+  long *expandedSizes = THAlloc(sizeof(long)*ndim);
+  long *expandedStrides = THAlloc(sizeof(long)*ndim);
+
+  for (long i = 0; i < numUnsqueezed; ++i) {
+    expandedSizes[i] = 1;
+    expandedStrides[i] = 0;
```
apaszke: This seems unnecessary - it's overwritten in the loop below. Maybe just fill the sizes in that one too.

lib/TH/THStorage.c
killeent: They could in theory be asking to infer more than 2 dimensions, so i think its simpler to have this error message.

lib/TH/generic/THTensor.c
```diff
@@ -285,6 +285,57 @@ void THTensor_(resize5d)(THTensor *self, long size0, long size1, long size2, lon
...
 }
 
+void THTensor_(expandNd)(THTensor *tensor, long *sizes, ptrdiff_t ndim) {
+  THArgCheck(tensor->nDimension > 0, 0, "can't expand an empty tensor");
+  long numUnsqueezed = ndim - tensor->nDimension;
+
+  long *expandedSizes = THAlloc(sizeof(long)*ndim);
+  long *expandedStrides = THAlloc(sizeof(long)*ndim);
+
+  for (long i = numUnsqueezed; i < ndim; ++i) {
```
gchanan: This doesn't look correct to me.  Consider the case where dims(tensor) == 1 and ndim = 2.  Then the first time through the loop i = 1, and we try to get THTensor_(size)(tensor, 1), but there is only a dim 0.

killeent:
I'll just do this in pytorch

lib/TH/generic/THTensor.c
apaszke: Yeah, it should be i - numUnsqueezed on the right

PR #1016, opened 2017-04-28
fix cdiv/div unit tests; fix Mac OS X require ffi bug

nicholas-leonard:
* fixes cdiv and div unit tests.
 * fixes issue on Mac OS X where `cd torch7 && th -e "torch.[anything]"` would have FFI.lua require itself (because the filesystem isn't case sensitive...)
so basically, within torch directory, require 'ffi' in FFI.lua requires itself because to Mac OS X, the FFI.lua and ffi.lua files are equivalent. And require looks in the local dir first.


soumith:
thanks!

Issue #1026, opened 2017-05-10
Slow THFloatTensor_sigmoid()


nicholas-leonard:
Hi @soumith @andresy @pavanky ,

Currently, the core of the LSTM (in Lua code) looks like this:

```
         local c = torch.Timer()
         cur_gates:addmm(bias_expand, cur_x, Wx)
         cur_gates:addmm(prev_h, Wh)
         times.addmm = (times.addmm or 0) + c:time().real; c:reset()

         cur_gates[{{}, {1, 3 * hiddensize}}]:sigmoid()
         times.sigmoid = (times.sigmoid or 0) + c:time().real; c:reset()

         cur_gates[{{}, {3 * hiddensize + 1, 4 * hiddensize}}]:tanh()
         times.tanh = (times.tanh or 0) + c:time().real; c:reset()

         local i = cur_gates[{{}, {1, hiddensize}}] -- input gate
         local f = cur_gates[{{}, {hiddensize + 1, 2 * hiddensize}}] -- forget gate
         local o = cur_gates[{{}, {2 * hiddensize + 1, 3 * hiddensize}}] -- output gate
         local g = cur_gates[{{}, {3 * hiddensize + 1, 4 * hiddensize}}] -- input transform
         times.narrow = (times.narrow or 0) + c:time().real; c:reset()

         next_h:cmul(i, g)
         next_c:cmul(f, prev_c):add(next_h)
         next_h:tanh(next_c):cmul(o)
         times.cmul = (times.cmul or 0) + c:time().real; c:reset()
```

The timers above benchmark the CPU FloatTensor.
Here is the benchmarked time and percentage of time of each part:
```
addmm   0.029978036880493       0.31822806466962
narrow  0.00019574165344238     0.0020778708024985
sigmoid 0.034398794174194       0.36515605543688
tanh    0.013311147689819       0.14130280727685
cmul    0.015677928924561       0.16642707458063
```

So as you can see the sigmoid() is the bottleneck on CPU.
Torch has its own custorm sigmoid CPU implementation: https://github.com/torch/torch7/blob/master/lib/TH/THMath.h#L4-L6

TensorFlow seems to be using an accelerated version of sigmoid:
http://www.literacystudies.com/src2/tensorflow/core/cwise__ops_8h_source.html

[...]

PR #1027, opened 2017-05-11
fast sigmoid

nicholas-leonard:
This PR provides a faster implement of `torch.Tensor.sigmoid`. As outlined in  #1026, the rationale for this PR is that TensorFlow has a 1.77x speedup over Torch LSTMs on CPU using Float tensors. This PR brings us to par with TF LSTMs. 

The downside is that `torch.Tensor.sigmoid` will lose some precision. The rough approximation introduces about 0.2% error. The speedup seems to warrant the approximation because the throughput for a typical LSTM goes from 455 samples/sec to 910 samples/sec, which compares well to the 809 samples/sec for TF. For a smaller LSTM those numbers are respectively, 3087 to 3900 for Torch vs  4139 for TF.

An alternative could be to implement a separate THTensor_(fastsigmoid) for this rough approximation.

### DIFF
--- a/lib/TH/THMath.h
+++ b/lib/TH/THMath.h
@@ -2,7 +2,11 @@
 #define _THMATH_H
 
 static inline double TH_sigmoid(double value) {
-  return 1.0 / (1.0 + exp(-value));
+  // faster version of 1.0 / (1.0 + exp(-value));
+  double x = fabs(value);
+  double x2 = x*x;
+  double e = 1.0f + x + x2*0.555f + x2*x2*0.143f;
+  return 1.0f / (1.0f + (value > 0 ? 1.0f / e : e));
 }
 
 static inline double TH_frac(double x) {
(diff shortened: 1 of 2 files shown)


soumith:
we used to have fast exponential approximations in TH for various things, and we phased them all out because of very subtle training issues for various use-cases. So I am going to reject this PR for the same reason.
However, I think @pavanky the wizard can easily speed this up without loss of precision.

pavanky:
@soumith The easiest thing to try is to try and incorporate the float versions of the math functions instead of defaulting to double versions for everything. This does still lose a bit of accuracy to existing versions but it should have the expected precision for floating point numbers.

soumith:
that does seem reasonable

nicholas-leonard:
@soumith Wizard @pavanky is sending a PR with TH_sigmoidf that gives about the same speedup. Wait out.

PR #1040, opened 2017-05-26
allow torch to link with Intel MKL _ilp64 model (large integers)

elikosan:
Sorry for the large pull request.
I improved upon the reverted pull request #981. Thanks for the suggestions.

The goal is to allow compilation and link with Intel MKL _ILP64 (64b integers) model. This is necessary when Torch is linked with another application that itself is linked to _ILP64, otherwise the linker on linux links to the incorrect version resulting in a MKL runtime error.
The default build will not change will not be impacted. To trigger linking to _ILP64, the environment variable MKL_ILP64 has to be defined prior to building.
The INTEL_MKL_DIR environment variable now can be set so that Intel MKL can be found by the build.
I tested this patch on both linux and Windows, both with and without MKL_ILP64 defined.
Again, there should be no impact when MKL_ILP64 is not defined.


elikosan:
Also, apologies for the un-squashed commits. I yet have to learn the intricacies of Git (I'm a SVN user).

PR #1046, opened 2017-06-14
Implementation of Alias Multinomial for faster Multinomial sampling

amartya18x:
This allows for faster sampling from a multinomial distribution using the __Alias__ method which is described in this [paper](http://www.tandfonline.com/doi/abs/10.1080/00031305.1979.10482697).

It allows almost `5x` speed in sampling than `torch.multinomial`. To see the difference on your machine run ` th test/test_aliasMultinomial.lua`

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -189,21 +189,17 @@ torch.Tensor.isTensor = torch.isTensor
 torch.setheaptracking(true)
 
 function torch.multinomialAliasSetup(probs, state)
-    local prob_tbl, alias_tbl
-    if state then 
-      assert(torch.type(state) == 'table')
+   if torch.type(state) == 'table' then 
       state[1], state[2] = torch.multinomialAliasSetup_(probs, state[1], state[2])
-    else 
-      prob_tbl, alias_tbl = torch.multinomialAliasSetup_(probs)
-      state = {prob_tbl, alias_tbl}
-    end
-    return state
+   else
+      state = {}
+      state[1], state[2] = torch.multinomialAliasSetup_(probs)
+   end
+   return state
 end
 
 function torch.multinomialAlias(output, state)
-   local prob_tbl, alias_tbl
-   prob_tbl, alias_tbl = state[1], state[2]
-   torch.DoubleTensor.multinomialAlias_(output, prob_tbl, alias_tbl)
+   torch.DoubleTensor.multinomialAlias_(output, state[1], state[2])
    return output
 end
 
--- a/init.lua
+++ b/init.lua
@@ -159,7 +159,6 @@ require('torch.FFInterface')
 require('torch.Tester')
 require('torch.TestSuite')
 require('torch.test')
-
 function torch.totable(obj)
    if torch.isTensor(obj) or torch.isStorage(obj) then
       return obj:totable()
@@ -188,5 +187,24 @@ torch.Tensor.isTensor = torch.isTensor
 
 -- remove this line to disable automatic heap-tracking for garbage collection
 torch.setheaptracking(true)
+
+function torch.multinomialAliasSetup(probs, state)
+    local prob_tbl, alias_tbl
+    if state then 
+      assert(torch.type(state) == 'table')
+      state[1], state[2] = torch.multinomialAliasSetup_(probs, state[1], state[2])
+    else 
+      prob_tbl, alias_tbl = torch.multinomialAliasSetup_(probs)
+      state = {prob_tbl, alias_tbl}
+    end
+    return state
+end
+
+function torch.multinomialAlias(output, state)
+   local prob_tbl, alias_tbl
+   prob_tbl, alias_tbl = state[1], state[2]
+   torch.DoubleTensor.multinomialAlias_(output, prob_tbl, alias_tbl)
+   return output
+end
 
 return torch
--- a/lib/TH/generic/THTensorRandom.c
+++ b/lib/TH/generic/THTensorRandom.c
@@ -70,6 +70,118 @@ void THTensor_(logNormal)(THTensor *self, THGenerator *_generator, double mean, 
   TH_TENSOR_APPLY(real, self, *self_data = (real)THRandom_logNormal(_generator, mean, stdv););
 }
 
+
+void THTensor_(multinomialAliasSetup)(THTensor *probs, THLongTensor *J, THTensor *q)
+{
+  long inputsize = THTensor_(nElement)(probs);
+  long i = 0;
+  THLongTensor *smaller = THLongTensor_newWithSize1d(inputsize);
+  THLongTensor *larger = THLongTensor_newWithSize1d(inputsize);
+  long small_c = 0;
+  long large_c = 0;
+  THLongTensor_resize1d(J, inputsize);
+  THTensor_(resize1d)(q, inputsize);
+  
+  for(i = 0; i < inputsize; i++)
+    {
+      THTensor_fastSet1d(J, i, 0L);
+      
+      real val = THStorage_(get)(probs->storage,
+                                 probs->storageOffset+i*probs->stride[0]);
+      THTensor_fastSet1d(q, i, inputsize*val);
+      
+      if (inputsize * val < 1.0)
+        {
+          THTensor_fastSet1d(smaller, small_c, i);
+          small_c += 1;
+        }
+      else
+        {
+          THTensor_fastSet1d(larger, large_c, i);
+          large_c += 1;
+        }
+    }
+
+  // Loop through and create little binary mixtures that
+  // appropriately allocate the larger outcomes over the
+  // overall uniform mixture.
+  long large, small;
+  while(small_c > 0 && large_c > 0)
+    {
+      large = THTensor_fastGet1d(larger, large_c-1);
+      small = THTensor_fastGet1d(smaller, small_c-1);
+      
+      THTensor_fastSet1d(J, small, large);
+      
+       THTensor_(data)(q)[large*q->stride[0]] -= 1.0 - THTensor_fastGet1d(q, small);
+      
+      if(q->storage->data[large] < 1.0)
+        {
+          THTensor_fastSet1d(smaller, small_c-1, large);
+          large_c -= 1;
+        }
+      else
+        {
+          THTensor_fastSet1d(larger, large_c-1, large);
+          small_c -= 1;
+        }
+    }
+
+  real q_min = THTensor_fastGet1d(q, inputsize-1);
+  real q_max = q_min;
+  real q_temp;
+  for(i=0; i < inputsize; i++)
+    {
+      q_temp = THTensor_fastGet1d(q, i);
+      if(q_temp < q_min)
+        q_min = q_temp;
+      else if(q_temp > q_max)
+        q_max = q_temp;
+    }
+  THArgCheckWithCleanup((q_min > 0),
+                        THCleanup(THLongTensor_free(smaller); THLongTensor_free(larger);), 2,
+                        "q_min is less than 0");
+  
+  if(q_max > 1)
+    {
+      for(i=0; i < inputsize; i++)
+        {
+          THTensor_(data)(q)[i*q->stride[0]] /= q_max;
+        }
+    }
+  for(i=0; i<inputsize; i++)
+    {
+      // sometimes an large index isn't added to J. 
+      // fix it by making the probability 1 so that J isn't indexed.
+      if(J->storage->data[i] <= 0)
+        q->storage->data[i] = 1.0;
+    }
+  THLongTensor_free(smaller);
+  THLongTensor_free(larger);
+}
+void THTensor_(multinomialAliasDraw)(THLongTensor *self, THGenerator *_generator, THLongTensor *J, THTensor *q)
+{
+  long K = THLongTensor_nElement(J);
+  long output_nelem = THLongTensor_nElement(self);
+  THGenerator* gen = THGenerator_new();
+  
+  int i = 0, _mask=0;
+  real _q;
+  long rand_ind, sample_idx, J_sample, kk_sample;
+  for(i=0; i< output_nelem; i++)
+    {
+      rand_ind = (long)THRandom_uniform(gen, 0, K) ;
+      _q = THTensor_fastGet1d(q, rand_ind);
+
+      _mask = THRandom_bernoulli(gen, _q);
+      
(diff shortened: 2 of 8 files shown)


aliasMultinomial.lua
```diff
@@ -0,0 +1,13 @@
+local AM = torch.class('torch.AliasMultinomial')
```
soumith: let's not expose this as a class like this.
Almost all math/RNG stuff in torch.* are exposed functionally.

If one wants a repeated draw, they should be able to call:
```
gen = torch.aliasmultinomial_generator(probs)
out = torch.aliasmultinomial(gen) -- or, out = torch.aliasmultinomial(out, gen)
```

lib/TH/generic/THTensorRandom.c
```diff
@@ -70,6 +70,141 @@ void THTensor_(logNormal)(THTensor *self, THGenerator *_generator, double mean,
   TH_TENSOR_APPLY(real, self, *self_data = (real)THRandom_logNormal(_generator, mean, stdv););
 }
 
+
+void THTensor_(alias_multinomial_setup)(THTensor *probs, THLongTensor *J, THTensor *q)
```
apaszke: Can you make the name conforming to everything else in the library? Use e.g. `multinomialAliasSetup` and `multinomialAliasDraw`

lib/TH/generic/THTensorRandom.c
```diff
@@ -70,6 +70,141 @@ void THTensor_(logNormal)(THTensor *self, THGenerator *_generator, double mean,
...
+  long output_nelem = THLongTensor_nElement(self);
+  THGenerator* gen = THGenerator_new();
+  
+  int i = 0, _mask=0;
+  real _q;
+  long rand_ind, sample_idx, J_sample, kk_sample;
+  for(i=0; i< output_nelem; i++)
+    {
+      rand_ind = (long)THRandom_uniform(gen, 0, K) ;
+      _q = THStorage_(get)(q->storage,
```
apaszke: `THStorage_(get)` and `THStorage_(set)` are inefficient, because they do bound checks, that are unnecessary here. Use `THStorage_(fastGet)`

lib/TH/generic/THTensorRandom.c
```diff
@@ -70,6 +70,141 @@ void THTensor_(logNormal)(THTensor *self, THGenerator *_generator, double mean,
...
+  while(small_c > 0 && large_c > 0)
+    {
+      large = THLongStorage_get(larger->storage,
+                                larger->storageOffset+(large_c-1)*larger->stride[0]);
+      small = THLongStorage_get(smaller->storage,
+                                smaller->storageOffset+(small_c-1)*smaller->stride[0]);
+      
+      THLongStorage_set(J->storage,
+                        J->storageOffset+small*J->stride[0],
+                        large+1L);
```
apaszke: This can be replaced with `THLongTensor_fastSet(J, small, large+1L)`. Same applies to most other places. There's no need to unpack the storage from the tensor

aliasMultinomial.lua
amartya18x: @soumith Does this seem okay ?```gen = torch.aliasmultinomial_generator(probs)
out = torch.aliasmultinomial(gen, n_sample ) ``` 
As @pavanky  suggested, instead of passing the output tensor, we should pass the size instead

lib/TH/generic/THTensorRandom.c
amartya18x: Okay, will do that.

lib/TH/generic/THTensorRandom.c
amartya18x: Okay. Sorry, didn't know about this.

aliasMultinomial.lua
soumith: this sounds okay. in the Torch world, passing output Tensor is the standard way of doing things. Passing the size is almost never done (as the habit of pre-allocating and re-using buffers is prevalent)

lib/TH/generic/THTensorRandom.c
amartya18x: Okay, makes sense. I wasn't aware of the `fastGet` and the `fastSet` methods.

aliasMultinomial.lua
pavanky: I should have clarified that I wanted this function to support the formats `torch.multinomial` supported. This included `torch.multinomial(p, n)`. 

@amartya18x If possible we should be doing both. If not possible, stick with the current API for the function.

amartya18x: This has to preprocess though. So, it will still not be a single call like `torch.multinomial` but a setup and then `torch.aliasMultinomialDraw(n)`

amartya18x:
@soumith @pavanky This doesn't include any class but includes two function calls.
```lua
th> probs = torch.DoubleTensor({0.2, 0.2, 0.3, 0.3})
                                                                      [0.0001s]
                                                                      
th> prob_tbl, alias_tbl = torch.multinomialAliasSetup(probs)
                                                                      [0.0002s]
                                                                      
th> output = torch.LongTensor(2,3)
                                                                      [0.0001s]
                                                                      

th> alias_tbl.multinomialAliasDraw(output:view(-1), prob_tbl, alias_tbl)
 3
 4
 3
 1
 2
 3
[torch.LongTensor of size 6]

                                                                      [0.0003s]
                                                                      
th> output
 3  4  3
 1  2  3
[torch.LongTensor of size 2x3]

```

amartya18x:
@apaszke Thanks for suggesting those. It made the functions pretty fast. :)

soumith:
why would you need the table return?
can you just return one Tensor that contains all state?

soumith:
and two functions 
state = torch.aliasMultinomialSetup(probs)
torch.aliasMultinomial(out, state)

amartya18x:
Its two tensors. They are of different types. One is Long and the other is double. 

And `torch.aliasMultinomial` isn't working because the `output` needs to be a `longTensor` and `torch.aliasMultinomial` says that the function isn't defined for `torch.LongTensor`. This is the only workaround I got.

Could you suggest someway to make that work ?

soumith:
in cwrap, define aliasMultinomial_
in lua, maybe in random.lua or something, define torch.aliasMultinomial as a lua function that calls torch.LongTensor.multinomial_(...)

amartya18x:
@soumith Is it okay to do that `init.lua` . `torch` isn't available in `random.lua`.

soumith:
sure, I'll move it if i need to, but do it  in init.lua

amartya18x:
Great. That helps everything.

amartya18x:
@soumith 
```lua 
th> probs = torch.DoubleTensor({0.2, 0.2, 0.3, 0.3})
                                                                      [0.0001s]
th> state = torch.multinomialAliasSetup(probs)
                                                                      [0.0001s]
th> state
{
  1 : LongTensor - size: 4
  2 : DoubleTensor - size: 4
}
                                                                      [0.0002s]
th> output = torch.LongTensor(2,3)
 
                                                                      [0.0003s]
th> torch.multinomialAlias(output:view(-1), state)
 4
 1
 2
 3
 2
 2
[torch.LongTensor of size 6]

                                                                      [0.0002s]
th> output
 4  1  2
 3  2  2
[torch.LongTensor of size 2x3]

                                                                      [0.0001s]

**Comment by nicholas-leonard** [comment]
This seems ready to merge

**Comment by amartya18x** [comment]
@nicholas-leonard There is a failing test but that is due the log-normal and it is on clang.

**Comment by pavanky** [comment]
@amartya18x can you rebase tis PR into a single commit?

**Comment by amartya18x** [comment]
@pavanky Yes ofcourse

File: init.lua [hunk 5]
```diff
@@ -189,4 +188,23 @@ torch.Tensor.isTensor = torch.isTensor
...
+    return state
+end
+
+function torch.multinomialAlias(output, state)
+   local prob_tbl, alias_tbl
+   prob_tbl, alias_tbl = state[1], state[2]
+   torch.DoubleTensor.multinomialAlias_(output, prob_tbl, alias_tbl)
+   return output
+end
+
```
pavanky: Do these functions need to be moved into a different file ? Or is it ok to leave them here?

lib/TH/generic/THTensorRandom.c
```diff
@@ -70,6 +70,118 @@ void THTensor_(logNormal)(THTensor *self, THGenerator *_generator, double mean,
...
+        q->storage->data[i] = 1.0;
+    }
+  THLongTensor_free(smaller);
+  THLongTensor_free(larger);
+}
+void THTensor_(multinomialAliasDraw)(THLongTensor *self, THGenerator *_generator, THLongTensor *J, THTensor *q)
+{
+  long K = THLongTensor_nElement(J);
+  long output_nelem = THLongTensor_nElement(self);
+  THGenerator* gen = THGenerator_new();
```
soumith: this seems like a bug. why are you creating `gen` and using it as the generator to create bernoulli and uniform below? I think you should use `_generator` right?

init.lua
```diff
@@ -189,4 +188,23 @@ torch.Tensor.isTensor = torch.isTensor
 -- remove this line to disable automatic heap-tracking for garbage collection
 torch.setheaptracking(true)
 
+function torch.multinomialAliasSetup(probs, state)
+    local prob_tbl, alias_tbl
```
soumith: this whole thing can be simplified as:

```
state = state or {}
assert(torch.type(state) == 'table')
state[1], state[2] = torch.multinomialAliasSetup_(probs, state[1], state[2])
return state
```

init.lua
```diff
@@ -189,4 +188,23 @@ torch.Tensor.isTensor = torch.isTensor
...
+      state[1], state[2] = torch.multinomialAliasSetup_(probs, state[1], state[2])
+    else 
+      prob_tbl, alias_tbl = torch.multinomialAliasSetup_(probs)
+      state = {prob_tbl, alias_tbl}
+    end
+    return state
+end
+
+function torch.multinomialAlias(output, state)
+   local prob_tbl, alias_tbl
```
soumith: the verbosity here seems unnecessary. 
Also, you should call `output.multinomialAlias`, instead of `torch.DoubleTensor.multinomialAlias` right? because if it's a FloatTensor, you'd want to call the float version directly.
```
function torch.multinomialAlias(output, state)
    output.multinomialAlias_(output, state[1], state[2])
    return output
end
```

init.lua
soumith: it's fine.

lib/TH/generic/THTensorRandom.c
amartya18x: I am creating the uniform random numbers outside the kernel and passing it to the kernels instead of calling the random number sampler internally from the kernel

lib/TH/generic/THTensorRandom.c
```diff
@@ -70,6 +70,118 @@ void THTensor_(logNormal)(THTensor *self, THGenerator *_generator, double mean,
...
+  long small_c = 0;
+  long large_c = 0;
+  THLongTensor_resize1d(J, inputsize);
+  THTensor_(resize1d)(q, inputsize);
+  
+  for(i = 0; i < inputsize; i++)
+    {
+      THTensor_fastSet1d(J, i, 0L);
+      
+      real val = THStorage_(get)(probs->storage,
```
soumith: Simplify using fastGet1d:
```
real val = THTensor_fastGet1d(probs, i);
```

init.lua
amartya18x: But `output` is a `torch.LongTensor` and not a `FloatTensor`.

lib/TH/generic/THTensorRandom.c
```diff
@@ -70,6 +70,118 @@ void THTensor_(logNormal)(THTensor *self, THGenerator *_generator, double mean,
...
+  // overall uniform mixture.
+  long large, small;
+  while(small_c > 0 && large_c > 0)
+    {
+      large = THTensor_fastGet1d(larger, large_c-1);
+      small = THTensor_fastGet1d(smaller, small_c-1);
+      
+      THTensor_fastSet1d(J, small, large);
+      
+       THTensor_(data)(q)[large*q->stride[0]] -= 1.0 - THTensor_fastGet1d(q, small);
```
soumith: between lines 116 and 118, you used two different ways to get q's data pointer.
Keep this consistent:

```
real *q_data = THTensor_(data)(q);
q_data[large * q->stride[0]] -= 1.0 - THTensor_fastGet1d(q, small);

if (q_data[large] < 1.0)
```

lib/TH/generic/THTensorRandom.c
```diff
@@ -70,6 +70,118 @@ void THTensor_(logNormal)(THTensor *self, THGenerator *_generator, double mean,
...
+      for(i=0; i < inputsize; i++)
+        {
+          THTensor_(data)(q)[i*q->stride[0]] /= q_max;
+        }
+    }
+  for(i=0; i<inputsize; i++)
+    {
+      // sometimes an large index isn't added to J. 
+      // fix it by making the probability 1 so that J isn't indexed.
+      if(J->storage->data[i] <= 0)
```
soumith: again, inconsistent way of getting data pointer, lines 156/157 vs line 149. Maybe you can get the data pointers of these variables at the beginning as q_data, J_data and reuse them over the whole function.

lib/TH/generic/THTensorRandom.c
soumith: i still dont get it. if you create another generator `gen`, the random seed for it'll be different and it wont be deterministic. You need to use `_generator` for all of your RNG creation inside this function so that you respect the user's `torch.manualseed`

amartya18x: Got it! Thanks! My fault.

init.lua
amartya18x: What if state[1] and state[2] are nil ? In that case torch.multinomialAliasSetup_() should only take `probs` as input.

i.e. `state[1], state[2] = torch.multinomialAliasSetup_(probs)`

soumith: oh i see, okay.

init.lua
amartya18x: This one too, right ? I couldn't figure a war around it. Unlike `torch.multinomial` this one doesn't take a `FloatTensor` as the first argument but `output` which  is a `DoubleTensor`.

amartya18x:
Does this look good ? I will rebase the commits then.

soumith:
Thanks Amartya!

PR #1161, opened 2018-07-18
Minor change to TH/CMakeLists.txt

Quibilia:
With this change, I've managed to get Torch to compile on Windows via LuaRocks and MSVC. By configuring LuaRocks for Lua 5.3 and maintaining a directory with the corresponding Lua source and binaries, in addition to the headers and libraries for GNU Readline as a dependency for TREPL, the installation proceeds with a few equation warnings, but no compiler errors. I plan to publish a Batch file for ease of installation via command line using the MSVC tools, but in the meantime this pull is more pressing.


Quibilia:
Should be noted that a custom CMake command needed to be used, in order to specify generation of NMake Makefiles. This .cmd file can be provided if necessary.

